PR #36, opened 2017-06-19
Disable maintaining image maps in the IPE

### LINKED ISSUE #35: [Image] Image map not supported
Although the inplace editor of the image component does allow to maintain an image map (https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/image/v1/image/_cq_editConfig.xml#L69) this is not considered during rendering. I.e. the HTL script does not include any logic to include an image map (i.e. `usemap` attribute or `map` elements).

kwin:
Image maps are currently not supported in the HTL markup and would
require some javascript to be usable in a responsive website, therefore
we should just disable them for now.

This fixes #35


bpauli:
@kwin Thank you for the contribution.

Maybe we will add map support again in a future release for the image component.

Issue #39, opened 2017-06-21
Image: NPE in case an asset was referenced not having a dot in its name


kwin:
In case an asset is referenced in an image component whose name does not contain a dot, this leads to an NPE
```
...
Caused by: org.apache.sling.models.factory.PostConstructException: Post-construct method has thrown an exception for model class com.adobe.cq.wcm.core.components.models.impl.v1.ImageImpl
	at org.apache.sling.models.impl.ModelAdapterFactory.createObject(ModelAdapterFactory.java:609)
	at org.apache.sling.models.impl.ModelAdapterFactory.internalCreateModel(ModelAdapterFactory.java:335)
	at org.apache.sling.models.impl.ModelAdapterFactory.createModel(ModelAdapterFactory.java:223)
	at org.apache.sling.scripting.sightly.models.impl.SlingModelsUseProvider.provide(SlingModelsUseProvider.java:126)
	at org.apache.sling.scripting.sightly.impl.engine.extension.use.UseRuntimeExtension.call(UseRuntimeExtension.java:72)
	... 115 common frames omitted
Caused by: java.lang.NullPointerException: null
	at com.adobe.cq.wcm.core.components.models.impl.v1.ImageImpl.initModel(ImageImpl.java:148)
	at sun.reflect.GeneratedMethodAccessor129.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.sling.models.impl.ModelAdapterFactory.invokePostConstruct(ModelAdapterFactory.java:792)
	at org.apache.sling.models.impl.ModelAdapterFactory.createObject(ModelAdapterFactory.java:607)
	... 119 common frames omitted
```

The reason is the extension always being extracted from the asset's node name (https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/blob/master/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/impl/v1/ImageImpl.java#L128). See also the related ticket #38. The recommended fix here is similar: Always derive the extension from the mime type!

raducotescu:
Fixed in 14e8ad3.

PR #41, opened 2017-06-21
Always generate extension in image link from the underlying mime type.

### LINKED ISSUE #38: Image: Upper case extensions in referenced assets lead to a 404
The `AdaptiveImageServlet` is only registered for the lower-case extensions `jpg`, `jpeg`, `gif` and `png`. Unfortunately the extension for images which refer to a DAM asset is extracted from the asset node name (https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/blob/master/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/impl/v1/ImageImpl.java#L128). So in case the assets is named `myimage.PNG` the used image url is not correctly resolved by the `AdaptiveImageServlet`. The extension should rather always be generated depending on the asset's metadata (i.e. the mime type) via the MimeTypeService like it is done for locally uploaded binaries (https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/blob/master/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/impl/v1/ImageImpl.java#L134).

kwin:
This fixes #38 and #39

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/impl/v1/ImageImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/impl/v1/ImageImpl.java
@@ -74,7 +74,6 @@ public class ImageImpl implements Image {
     private static final Logger LOGGER = LoggerFactory.getLogger(ImageImpl.class);
     private static final String DOT = ".";
     private static final String MIME_TYPE_IMAGE_JPEG = "image/jpeg";
-    private static final String PN_DAM_METADATA_MIME_TYPE = "dam:MIMEtype";
     private static final String MIME_TYPE_IMAGE_PREFIX = "image/";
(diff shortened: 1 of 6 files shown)


kwin:
I temporarily had to ignore the test `ImageImplTest.testTIFFImage()` because with this new approach the referenced asset (i.e. the TIFF) would need to be provided as test content even if you just test the image markup.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/impl/v1/ImageImpl.java
```diff
@@ -71,6 +74,9 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(ImageImpl.class);
     private static final String DOT = ".";
     private static final String MIME_TYPE_IMAGE_JPEG = "image/jpeg";
+    private static final String PN_DAM_METADATA_MIME_TYPE = "dam:MIMEtype";
```
kwin: Maybe one should use dc:format instead?

PR #43, opened 2017-06-26
Image component: wcmmode should be optional

stefanseifert:
wcmmode should be made optional because it's not available when calling Sling Models exporter view with ".model.json". Otherwise ".mode.json" view fails with an error in the error.log.


bpauli:
Thanks for the contribution!

PR #44, opened 2017-06-29
Mobile Emulator Provider registration for page component

stefanseifert:
pages with page components derived from `core/wcm/components/page/v1/page` do not show the "Emulator" button in the page editor, some other features do not work as well then.
registering the resource type as mobile emulator provider fixes the problem.

### DIFF
--- a/config/src/content/jcr_root/apps/core/wcm/config.author/com.day.cq.wcm.mobile.core.impl.MobileEmulatorProvider-core-components.config
+++ b/config/src/content/jcr_root/apps/core/wcm/config.author/com.day.cq.wcm.mobile.core.impl.MobileEmulatorProvider-core-components.config
@@ -1,0 +1,15 @@
+# Copyright 2017 Adobe Systems Incorporated
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


justinedelson:
Wouldn't the expectation be that the site's config does this against their proxy components as is in the case in we-retail (https://github.com/Adobe-Marketing-Cloud/aem-sample-we-retail/blob/master/config/src/content/jcr_root/apps/weretail/config.author/com.day.cq.wcm.mobile.core.impl.MobileEmulatorProvider-we-retail.xml) ?

justinedelson:
The downside of doing this is that it forces the use of the emulator on sites which may not want to use it. I'm not sure this is a *huge* downside, but it is one.

stefanseifert:
AEM 6.3 OOTB already contains such a registration for the included page component {{wcm/foundation/components/page}}.

why not for the page component from the core components? this simplifies the setup of a new project. if its forgotten the layout mode can still be opened, but is subtly broken (e.g. device selector is shown, but no device can be selected)

what is the use case of using the core component page component without this registration?

stefanseifert:
bq. The downside of doing this is that it forces the use of the emulator on sites which may not want to use
i agree for older AEM versions. but for AEM 6.3 for example the emulator is even shown e.g. when switching to layout mode (and stays visible when switching back) without this registration - but it's broken.
perhaps this is a bug in AEM.

bpauli:
@stefanseifert Thank you for the contribution!

Indeed in 6.3 it make no sense at the moment to provide no config. The broken emulator will fixed in one of the next Cumulative Fix Packs for 6.3.

PR #99, opened 2018-01-18
 Fixes #98: NullPointerException when ordering List component with nul…

### LINKED ISSUE #98: List component ordering throws NullPointerException when modification date or title is null
When using the ordering by lastModified date or by title in the List component, a NullPointerException will be thrown when getLastModified() or getTitle() return null respectively. 

According to the documentation com.day.cq.wcm.api.Page#getLastModified and getTitle may return null, so this needs to be expected and checked (https://helpx.adobe.com/experience-manager/6-3/sites/developing/using/reference-materials/javadoc/com/day/cq/wcm/api/Page.html#getLastModified()).

phornig:
…l lastModified or title

I have added a check for null for getLastModified and getTitle. For the comparison I have defined that null is after nonnull values for ascending ordering.


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ListImpl.java
```diff
@@ -391,9 +391,19 @@ public static OrderBy fromString(String value) {
         public int compare(Page item1, Page item2) {
             int i = 0;
             if (orderBy == OrderBy.MODIFIED) {
-                i = item1.getLastModified().compareTo(item2.getLastModified());
+                if (item1.getLastModified() != null && item2.getLastModified() != null) {
+                    i = item1.getLastModified().compareTo(item2.getLastModified());
+                } else {
+                    // define null date to be after nonnull, two null dates are equal
+                    i = (item1.getLastModified() == null ? 0 : -1) + (item2.getLastModified() == null ? 0 : 1);
```
bpauli: @phornig First let me thank you for this catch and your effort to fix this!

Given that `item1.getLastModified() != null` and `item2.getLastModified() == null` the result of i would be -1. If a null date should be after an nonnull date, the result of i have to be a positive integer.
Can you please update your PR to deal with this case? 

Thanks again!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ListImpl.java
```diff
@@ -391,9 +391,19 @@ public static OrderBy fromString(String value) {
...
+                }
             } else if (orderBy == OrderBy.TITLE) {
-                i = item1.getTitle().compareTo(item2.getTitle());
+                if (item1.getTitle() != null && item2.getTitle() != null) {
+                    i = item1.getTitle().compareTo(item2.getTitle());
+                } else {
+                    // define null title is after nonnull, two null titles are equal
+                    i = (item1.getTitle() == null ? 0 : -1) + (item2.getTitle() == null ? 0 : 1);
```
bpauli: Same as my comment in L398.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ListImpl.java
phornig: Thanks for reviewing this part, it is a bit twisted. I think my logic is correct, since item1 not null and item2 null means item1 < item2 which should return -1 to be aligned with Calendar.compareTo.

Nevertheless I replaced the logic with ObjectUtils.compare as @henrykuijpers suggested.

Also I have added unit test, which should catch errors in the comparison. Please have a look at the unit test to make sure my assertions are correct there. Thanks!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ListImpl.java
phornig: Also see at L398 :)

bpauli:
Thank you for the contribution!

PR #131, opened 2018-03-14
added schema.org microdata according to http://schema.org/BreadcrumbList

### LINKED ISSUE #129: Breadcrumb component misses schema.org microdata
The Breadcrumb component misses schema org microdata that is relevant for SEO.

I added microdata to Breadcrumb  according to http://schema.org/BreadcrumbList - see PR https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/pull/128 here

janrembold:
fixes #129


raducotescu:
@janrembold, this looks like a cool idea! I guess we should figure out if we can add microdata to all components and only to the Breadcrumb.

PR #135, opened 2018-03-15
[Page] Expose components from Template in Page Model

### LINKED ISSUE #134: [Page] Expose components from Template in Page Model
Components of the template are currently also not exposed when requesting the page with .model.json selector.

vladbailescu:
* get the page model for the structured resources returned by the template manager
* unit tests

(fixes #134)

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageImpl.java
@@ -50,9 +50,12 @@ import com.adobe.cq.wcm.core.components.models.Page;
 import com.day.cq.tagging.Tag;
 import com.day.cq.wcm.api.NameConstants;
 import com.day.cq.wcm.api.Template;
+import com.day.cq.wcm.api.TemplateManager;
+import com.day.cq.wcm.api.components.ComponentContext;
 import com.day.cq.wcm.api.designer.Design;
 import com.day.cq.wcm.api.designer.Designer;
 import com.day.cq.wcm.api.designer.Style;
+import com.day.cq.wcm.commons.WCMUtils;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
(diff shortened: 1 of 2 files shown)


pfauchere:
@vladbailescu could you please review the current PR?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageImpl.java
```diff
@@ -241,7 +244,32 @@ public String getExportedType() {
...
+                TemplateManager templateManager = resolver.adaptTo(TemplateManager.class);
+
+                // Resources backed by the template structure
+                if (templateManager != null) {
+                    children = templateManager.getStructureResources(componentContext);
+                }
+            }
+        }
+
+        if (children == null || !children.iterator().hasNext()) {
```
raducotescu: @pfauchere, what's the intention here? It seems like if the template provides structure resources then the page will not expose any more its own components.

pfauchere: They will be returned by the `TemplateManager#getStructureResources` function extracting the content of the page using the `componentContext`. And extend that content with the one from the structure of the template.

If no structured resources are found, we fallback to the default

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageImplTest.java
```diff
@@ -194,53 +213,135 @@ public void testDefaultDesign() {
...
+        testGetChildModel(false);
+    }
+
+    @Test
+    public void testGetChildModelEditableTemplate() {
+        testGetChildModel(true);
+    }
+
+    @Test
+    public void testGetChildModelStructured() {
```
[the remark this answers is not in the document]
pfauchere: Leftover after the refactoring. Thx for the catch

PR #137, opened 2018-03-15
[CF] Provide Content Fragment configuration for resource types from Core Components

### LINKED ISSUE #136: [CF] Provide Content Fragment configuration for resource types from Core Components

vladbailescu:
* Added OSGi configuration

(fixes #136)


raducotescu:
@gknobloch, for these projects we cannot use the provisioning file, since they're not delivered with the Quickstart. Core Components also provide configurations like this.

PR #139, opened 2018-03-15
Update references to /etc/mobile

### LINKED ISSUE #138: Update references to /etc/mobile
Update outdated references

vladbailescu:
(fixes #138)

### DIFF
--- a/bundles/core/src/test/resources/page/test-conf.json
+++ b/bundles/core/src/test/resources/page/test-conf.json
@@ -170,7 +170,7 @@
                 "structure": {
                     "jcr:content": {
                         "cq:deviceGroups": [
-                            "/etc/mobile/groups/responsive"
+                            "mobile/groups/responsive"
                         ],
                         "cq:responsive": {
                             "breakpoints": {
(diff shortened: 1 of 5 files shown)


keepthebyte:
does that work on AEM 6.3? w

vladbailescu:
@keepthebyte, I verified with 6.3 SP1 and the emulators are being picked correctly from `/etc/mobile/groups/responsive` with this change.

vladbailescu:
@keepthebyte, you are actually right: relative paths are not supported in 6.3, they will be supported starting with 6.4.
However, given this is test content and our tests do not check responsive functionality (out of scope for the components) and pass with these changes, we would prefer to update the paths to the cleaner version, just in case new projects targeting 6.4 get inspiration from our content.

kwin:
@vladbailescu If this really works only with AEM 6.4 at least a short comment should be placed in that test content.

vladbailescu:
The content installs in 6.3 and produces no errors. However, using a relative path for for the responsive mobile groups will cause the mobile emulators to not load. Testing the emulators is not in scope of the core components so arguably the change _works_ for this project in 6.3 too.

Issue #140, opened 2018-03-15
[Image] Enable the Image Flip support and processing


raducotescu:
The Image Component and related image processing code (e.g. the `AdaptiveImageServlet`) should be enhanced for providing support for flipping images.

PR #141, opened 2018-03-16
[Image] Enable the Image Flip support and processing (#140)

raducotescu:
* added backend support for rendering flipped images; the flipping transformation is performed last, with the horizontal flip having priority over the vertical one

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -395,33 +395,20 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
             boolean flipVertically = componentProperties.get(Image.PN_FLIP_VERTICAL, Boolean.FALSE);
             if (is != null) {
                 if (rotationAngle != 0 || rectangle != null || resizeWidth > 0 || flipHorizontally || flipVertically) {
-                    Layer layer = null;
+                    Layer layer = new Layer(is);
                     if (rectangle != null) {
-                        layer = new Layer(is);
                         layer.crop(rectangle);
                         LOGGER.debug("Applied cropping transformation.");
                     }
                     if (rotationAngle != 0) {
-                        if (layer == null) {
-                            layer = new Layer(is);
-                        }
                         layer.rotate(rotationAngle);
                         LOGGER.debug("Applied rotation transformation ({} degrees).", rotationAngle);
                     }
                     if (flipHorizontally) {
-                        if (layer == null) {
-                            layer = new Layer(is);
-                        }
                         layer.flipHorizontally();
                     }
                     if (flipVertically) {
-                        if (layer == null) {
-                            layer = new Layer(is);
-                        }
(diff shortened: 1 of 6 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -287,12 +281,14 @@ private void resizeAndStreamAsset(SlingHttpServletResponse response, ValueMap co
         }
         int rotationAngle = getRotation(componentProperties);
         Rectangle rectangle = getCropRect(componentProperties);
-        if (rotationAngle != 0 || rectangle != null || resizeWidth > 0) {
+        boolean flipHorizontally = componentProperties.get(Image.PN_FLIP_HORIZONTAL, Boolean.FALSE);
```
richardhand: @raducotescu - for the other transform properties, we rely on the `ImageResource` class, shouldn't we use the same approach for flip, rather than defining them on the core image model.

raducotescu: `ImageResource` doesn't yet have properties for indicating that these transformations need to be performed. The enhancement in the TouchUI editor is recent and will probably be made available through a FP, but a FP cannot provide new API. Hence why I added these constants in our API.

bundles/core/src/test/resources/image/test-content.json
```diff
@@ -284,6 +284,36 @@
                             "jcr:uuid"          : "22a0a463-9614-48c5-8215-a35ffc7d1089"
                         }
                     }
+                },
+                "image22": {
+                    "jcr:primaryType"    : "nt:unstructured",
+                    "jcr:createdBy"      : "admin",
+                    "cq:lastModified"    : "2017-05-15T17:56:17.756+01:00",
+                    "sling:resourceType" : "core/wcm/components/image/v1/image",
```
jckautzmann: Are we also testing image/v2?

raducotescu: We're not actually testing the image component's implementation, but the AIS' behaviour.

richardhand:
We should document the strict transform order (crop -> rotate -> flipH -> flipV) somewhere suitable.

raducotescu:
True. I think that should be in the Image Component's README files, since that's where we also documented the URL patterns, which also map to the `AdaptiveImageServlet`.

> We should document the strict transform order (crop -> rotate -> flipH -> flipV) somewhere suitable.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -394,6 +408,18 @@ private void resizeAndStreamFile(SlingHttpServletResponse response, ValueMap com
                         layer.rotate(rotationAngle);
                         LOGGER.debug("Applied rotation transformation ({} degrees).", rotationAngle);
                     }
+                    if (flipHorizontally) {
```
vladbailescu: I'd put these two new `if`s after the null-test for cleaner code.

PR #145, opened 2018-03-26
Resource `cq/cloudconfig/components/scripttags/header` missing

### LINKED ISSUE #102: Resource `cq/cloudconfig/components/scripttags/header` missing
Hi Team,
Release version : 2.0.0

Under Page component (v2) 
`/apps/core/wcm/components/page/v2/page/headlibs.html` 
has a reference to resource 
`cq/cloudconfig/components/scripttags/header`

> `<sly data-sly-resource="${'cloudconfig-header' @ resourceType='cq/cloudconfig/components/scripttags/header'}"></sly>`

However, I couldn't find it anywhere in crxde. Not sure if i'm missing anything here.

vladbailescu:
* Conditionally including the cloudconfig script tags

(fixes #102)

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
@@ -55,6 +55,8 @@ public class PageImpl extends com.adobe.cq.wcm.core.components.internal.models.v
     protected static final String RESOURCE_TYPE = "core/wcm/components/page/v2/page";
     public static final String PN_REDIRECT_TARGET = "cq:redirectTarget";
 
+    private Boolean hasCloudconfigSupport;
+
     @OSGiService
     private HtmlLibraryManager htmlLibraryManager;
 
@@ -154,9 +156,14 @@ public class PageImpl extends com.adobe.cq.wcm.core.components.internal.models.v
 
     @Override
     public boolean hasCloudconfigSupport() {
-        if (productInfoProvider == null) {
-            return false;
+        if (hasCloudconfigSupport == null) {
+            if (productInfoProvider == null || productInfoProvider.getProductInfo() == null ||
+                    productInfoProvider.getProductInfo().getVersion() == null) {
+                hasCloudconfigSupport = false;
+            } else {
+                hasCloudconfigSupport = productInfoProvider.getProductInfo().getVersion().compareTo(new Version("6.4.0")) >= 0;
+            }
         }
-        return productInfoProvider.getProductInfo().getVersion().compareTo(new Version("6.4.0")) >= 0;
+        return hasCloudconfigSupport;
     }
 }
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
@@ -157,6 +157,6 @@ public class PageImpl extends com.adobe.cq.wcm.core.components.internal.models.v
         if (productInfoProvider == null) {
             return false;
         }
-        return productInfoProvider.getProductInfo().getVersion().compareTo(new Version("6.3.0")) > 0;
+        return productInfoProvider.getProductInfo().getVersion().compareTo(new Version("6.4.0")) >= 0;
     }
 }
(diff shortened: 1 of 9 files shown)


content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/footer.html
```diff
@@ -19,5 +19,7 @@
          data-sly-call="${clientlib.js @ categories=clientlibCategories}"></sly>
     <sly data-sly-include="customfooterlibs.html"></sly>
     <sly data-sly-resource="${'cloudservices' @ resourceType='cq/cloudserviceconfigs/components/servicecomponents'}"></sly>
-    <sly data-sly-resource="${'cloudconfig-footer' @ resourceType='cq/cloudconfig/components/scripttags/footer'}"></sly>
+    <sly data-sly-use.pip="com.adobe.granite.license.ProductInfoProvider" data-sly-test.aemVersion="${pip.productInfo.version}"
+         data-sly-test="${aemVersion.major > 6 || (aemVersion.major == 6 && aemVersion.minor > 3)}"
```
jckautzmann: where is the code that defines aemVersion?

content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/footer.html
```diff
@@ -19,5 +19,7 @@
          data-sly-call="${clientlib.js @ categories=clientlibCategories}"></sly>
     <sly data-sly-include="customfooterlibs.html"></sly>
     <sly data-sly-resource="${'cloudservices' @ resourceType='cq/cloudserviceconfigs/components/servicecomponents'}"></sly>
-    <sly data-sly-resource="${'cloudconfig-footer' @ resourceType='cq/cloudconfig/components/scripttags/footer'}"></sly>
+    <sly data-sly-use.pip="com.adobe.granite.license.ProductInfoProvider" data-sly-test.aemVersion="${pip.productInfo.version}"
```
joerghoh: Do not have this logic in the HTL script, but rather move it to a dedicated model class.

vladbailescu: I considered doing that but adding a `Page#hasCloudconfigSupport` method seemed clunky and misleading (as it's a product feature, not a page property). @joerghoh, would that make better sense in your opinion?

jckautzmann: I would favor the approach of having the logic in the Page model.

raducotescu: Like @vladbailescu said, I find the approach brittle given how often the cloud configs are changed. Having something like that in the `Page` model leads to API changes, whereas doing this in the `headlibs.html` + `footer.html` is clean enough.

joerghoh: Do not change the Page API (it's already too large), but rather  move this logic into a sling model (ProductFeatureModel).

Please consider that the way the core components are built should be following best practices; and doing a version check in HTL scripts is definitly not.

vladbailescu: Again, providing an API that describes the product features from the Core Components seems out-of-order, we already make use of the `com.adobe.granite.license.ProductInfoProvider`. 

Unfortunately the version check is a bit verbose in HTL but apart from this I don't really see how checking the WCM mode or a product feature is preferable to a version check.

joerghoh: My point is that this check is done at the wrong place (no matter how it is implemented, that's a different question). If this check would be in a Java Model, it would be much easier to come up with a different way to determining this information.

jckautzmann: How about extending the Page model (not the Page API) with a method `hasCloudconfigSupport`, as suggested by @vladbailescu ?

joerghoh: I am fine with it

vladbailescu: @joerghoh @jckautzmann Please check 017eccc. For me the only upside is that we can unit-test the version comparison.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -146,4 +151,12 @@ public String getCssClassNames() {
     public NavigationItem getRedirectTarget() {
         return redirectTarget;
     }
+
+    @Override
+    public boolean hasCloudconfigSupport() {
+        if (productInfoProvider == null) {
+            return false;
+        }
+        return productInfoProvider.getProductInfo().getVersion().compareTo(new Version("6.4.0")) >= 0;
```
jckautzmann: I would also check for non `null` for `productInfoProvider.getProductInfo()` and `productInfoProvider.getProductInfo().getVersion()`.

joerghoh: Not sure if these will ever return null; at least in the API docs it's not mentioned.

jckautzmann: Exactly, better safe than sorry :)

Issue #150, opened 2018-04-03
Add code Lint for the clientlibs of the Core Components


richardhand:
We should verify that the JavaScript and CSS code is following best practice code conventions and style guidelines.

PR #154, opened 2018-04-04
Add code Lint for the clientlibs of the Core Components (#150)

richardhand:
Adds linting to the `core.wcm.components.content` build in the validate goal via the maven frontend plugin [0] and fixes all found issues.

The configured plugin:
• Installs node and npm
• Runs a defined npm "lint" script that in turn runs both "eslint" and "stylelint" scripts.

### JavaScript
• Uses eslint [1] to do the work.
• Extends from the base `eslint:recommended` configuration, which handles a number of pre-defined errors. The configurations covered are marked by a checkmark at [2].
• Can be run outside of the maven build by doing an `npm install`, followed by `npm run eslint`.
• Errors that can be auto-fixed (not all can) are tackled with `npm run eslint:fix`.
• Targets `ecmaVersion: 5` for now.
• I've defined what I see as a decent first set of rules, including style opinions and limiting language features, but they don't cover all possible options. Opinions welcome.
• Applied on both clientlib JS and ui-js End-to-End tests.

### Style
• Uses stylelint [3] to do the work.
• We add as dev dependency and extends `stylelint-config-recommended` which enables all "possible errors" [4].
• Additionally adds a decent first set of rules, including style opinions and limiting language features. I went through each rule.
• Can be run outside of the maven build by doing an `npm install`, followed by `npm run stylelint`.
• Errors that can be auto-fixed (not all can) are tackled with `npm run stylelint:fix`.
• Includes a plugin that validates the style selectors are valid BEM. Can be enabled on a per-component basis with a `/** @define componentName */` comment. Enabled for image and breadcrumb, but the search rules fail (the `.cmp-search__item--is-focused` selector is invalid BEM).

### Editor Config
• Adds an editor config.
• Conforms on spaces and an indent size of 4, plus the trimming of trailing whitespace and insistance on new lines at end of file.

### Thoughts
• **Site vs Editor linting** This is a blanket config for editor and site clientlibs, but we may have different needs for each case, for example we allow jQuery, Coral and Granite globals, but they aren't required / shouldn't be allowed for sites. We should look into a structure that means we don't have to repeat the `.eslintrc` config for each site clientlib.
• **Extend existing configurations?** we take the approach of defining our own (full) set of rules on top of recommended error handling, but we could also extend from some recognised, openly available configurations rather than maintaining this list, but it would have to align closely to our opinions / needs.

[0] - https://github.com/eirslett/frontend-maven-plugin
[1] - https://eslint.org/
[2] - https://eslint.org/docs/rules/
[3] - http://stylelint.io/
[4] - https://stylelint.io/user-guide/rules/#possible-errors


content/package.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "com.adobe.cq.core.wcm.components",
+    "description": "Adobe Experience Manager Core WCM Components Content Package",
+    "version": "1.0.0",
```
bpauli: Shouldn't we align the version of the npm module with the maven version?

richardhand: @bpauli - we should, but we have to sync with the pom version somehow

wimsymons: @bpauli you could use https://www.npmjs.com/package/sync-pom-version-to-package

jckautzmann:
@richardhand It would be nice to keep all the info from the description in a github docs page so that everybody knows how the plugin works and what it checks.

richardhand:
@jckautzmann - we could add a dedicated README.md in both the content and ui-js packages to document the npm commands.

PR #161, opened 2018-04-09
Adding filter rule for /apps/core/wcm/install in the all content pack…

### LINKED ISSUE #159: [Maven] Overwrite full install directory with installation of core.wcm.components.all
Currently `core.wcm.components.all` gets its full filter.xml automatically generated from the Maven Plugin. That leads to the fact that old bundle versions never get removed from the repository as the filter rule is only covering one specific bundle version, e.g. `/apps/core/wcm/install/core.wcm.components.core-1.0.6.jar`. Since the old bundle versions are not deployed anyways it would make sense to completely remove them from the repository by adding a filter rule for root `/apps/core/wcm/install`.

bpauli:
…age to remove old core bundles

fixes #159


bpauli:
Will create a new PR against development branch.

PR #171, opened 2018-04-26
[Image] Add support for SVG images

### LINKED ISSUE #170: [Image] Add support for SVG images
The Core Image Component should be enhanced to allow SVG images to be used with it.

jckautzmann:
- support the following: dragg-n-dropp an SVG asset from DAM, SVG file upload from a local file system
- if an SVG image is detected, the AdaptiveImageServlet streams it as is, it doesn’t do any resizing, cropping or mapping.
- note that AEM does not support editing of an SVG image
- the original SVG is never called directly, it is called through ``. As such the browser prevents embedded scripts to be executed.

fixes #170


jckautzmann:
@bpauli @vladbailescu Could you please have a look at this PR? Thx.

vladbailescu:
Looks ok to me. One thing I would add: disabling lazy loading and multiple image sizes for SVG images.

vladbailescu:
@bpauli Disabling the in-place editor is indeed something we should do. If we change the implementation to disable lazy loading and multiple sizes, then we don't need a different markup, as the JS logic will be bypassed completely.

jckautzmann:
@vladbailescu @bpauli Thx for your feedback. I created CQ-4244030 to disable inline editing for SVG images and as Vlad suggested I will change the implementation to disable lazy loading and multiple sizes in this PR.

jckautzmann:
@vladbailescu @bpauli I reconsidered your feedback and proceeded as follows:
- I didn’t disable lazy loading for SVG images
- "smart sizes” is not supported for SVG images: the model returns an empty array
See commit: https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/pull/171/commits/c71b93abac823e3d98e3b65a3a56ab99f00fe1f6

Issue #155, opened 2018-04-05
[Sandbox][Teaser] Add support for multiple call to action (CTA)


vladbailescu:
Our teaser component should allow adding multiple call-to-action (CTA). Behavior will be controlled via a `Call to action(s)` checkbox:
- When unchecked
  - A link field will be provided
  - The link will be applied to the image and title of the teaser
  - If the link is an AEM page path the title and description of the teaser will fallback to the page title and description.
- When checked
  - The link field is hidden and instead a multifield control will allow authors to select the CTA link and text
  - No fallback values for title and description in this case.

vladbailescu:
Changes we added to `development` branch via:

* https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/pull/156 - removing link text and cleanup
* https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/pull/172 - updating dialog to support multiple CTAs
* https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/pull/179 - implementation and test for multiple CTAs

PR #172, opened 2018-04-27
[Teaser] Add support for multiple call to action (CTA) (#155)

vladbailescu:
Initial proposal for multiple CTAs implementation

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
@@ -15,11 +15,14 @@
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.sandbox.models;
 
+import java.util.List;
+
 import javax.annotation.Nonnull;
 
 import org.apache.sling.api.resource.Resource;
 
 import com.adobe.cq.export.json.ComponentExporter;
+import com.adobe.cq.wcm.core.components.models.ListItem;
 
 /**
  * Defines the {@code Teaser} Sling Model for the {@code /apps/core/wcm/components/teaser} component.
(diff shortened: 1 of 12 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
```diff
@@ -28,6 +31,20 @@
  */
 public interface Teaser extends ComponentExporter {
 
+    /**
+     * Name of the property that will store if the teaser has Call-to-Action elements
```
jckautzmann: I'd write it in present form: 
Name of the property that stores if the teaser has Call-to-Action elements

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
```diff
@@ -28,6 +31,20 @@
...
+     *
+     * @since com.adobe.cq.wcm.core.components.sandbox.models 1.0.0
+     */
+    String PN_WITH_CTA = "withCTA";
+
+    /**
+     * Name of the node where the Call-to-Action elements will be stored
```
jckautzmann: I'd write it in present form: 
Name of the parent node where the Call-to-Action elements are stored

content/src/content/jcr_root/apps/core/wcm/sandbox/components/teaser/v1/teaser/_cq_dialog/.content.xml
```diff
@@ -22,14 +22,52 @@
                             sling:resourceType="granite/ui/components/coral/foundation/container"
                             margin="{Boolean}true">
                         <items jcr:primaryType="nt:unstructured">
+                            <withCTA
+                                    jcr:primaryType="nt:unstructured"
+                                    sling:resourceType="granite/ui/components/coral/foundation/form/checkbox"
+                                    fieldDescription="When checked, allow adding Call To Action (CTA) elements."
```
jckautzmann: I'd reformulate as follows: 
Check to allow adding Call To Action (CTA) elements.

vladbailescu:
Merging this as @gabrielwalt is ok with the looks and behaviour of the dialogs. @andreeadracea will submit a new PR with the final changes in the model and implementation.

PR #179, opened 2018-05-10
[Teaser] Add support for multiple call to action (CTA)

vladbailescu:
* Added the implementation for Title Types
 * Removed the "Don't add links to description"
 * Created unit tests for Title Types
 * Updated some unit tests
 * Implemented the exporter for CTAs

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TitleImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TitleImpl.java
@@ -19,7 +19,6 @@ package com.adobe.cq.wcm.core.components.internal.models.v1;
 import javax.annotation.Nonnull;
 import javax.annotation.PostConstruct;
 
-import com.adobe.cq.wcm.core.components.internal.Utils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.resource.Resource;
@@ -30,6 +29,7 @@ import org.apache.sling.models.annotations.injectorspecific.ValueMapValue;
 
 import com.adobe.cq.export.json.ComponentExporter;
 import com.adobe.cq.export.json.ExporterConstants;
+import com.adobe.cq.wcm.core.components.internal.Utils;
 import com.adobe.cq.wcm.core.components.models.Title;
 import com.day.cq.commons.jcr.JcrConstants;
 import com.day.cq.wcm.api.Page;
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TitleImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TitleImpl.java
@@ -19,6 +19,7 @@ package com.adobe.cq.wcm.core.components.internal.models.v1;
 import javax.annotation.Nonnull;
 import javax.annotation.PostConstruct;
 
+import com.adobe.cq.wcm.core.components.internal.Utils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.resource.Resource;
@@ -59,9 +60,9 @@ public class TitleImpl implements Title {
     private String type;
 
     /**
-     * The {@link Heading} object for the type of this title.
+     * The {@link com.adobe.cq.wcm.core.components.internal.Utils.Heading} object for the type of this title.
      */
-    private Heading heading;
+    private Utils.Heading heading;
 
     @PostConstruct
     private void initModel() {
@@ -70,9 +71,9 @@ public class TitleImpl implements Title {
         }
 
         if (heading == null) {
-            heading = Heading.getHeading(type);
+            heading = Utils.Heading.getHeading(type);
             if (heading == null) {
-                heading = Heading.getHeading(currentStyle.get(PN_DESIGN_DEFAULT_TYPE, String.class));
+                heading = Utils.Heading.getHeading(currentStyle.get(PN_DESIGN_DEFAULT_TYPE, String.class));
             }
         }
     }
@@ -94,35 +95,6 @@ public class TitleImpl implements Title {
     @Override
     public String getExportedType() {
         return resource.getResourceType();
-    }
-
-    private enum Heading {
-
-        H1("h1"),
-        H2("h2"),
-        H3("h3"),
-        H4("h4"),
-        H5("h5"),
-        H6("h6");
-
-        private String element;
-
-        Heading(String element) {
-            this.element = element;
-        }
-
-        private static Heading getHeading(String value) {
-            for (Heading heading : values()) {
-                if (StringUtils.equalsIgnoreCase(heading.element, value)) {
-                    return heading;
-                }
-            }
-            return null;
-        }
-
-        public String getElement() {
-            return element;
-        }
     }
 
 }
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/internal/models/v1/TeaserImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/internal/models/v1/TeaserImpl.java
@@ -63,14 +63,14 @@ public class TeaserImpl extends AbstractImageDelegatingModel implements Teaser {
     private String description;
     private String linkURL;
     private String titleType;
-    private boolean withCTA = false;
-    private boolean hideTitle = false;
-    private boolean hideDescription = false;
-    private boolean hideImageLink = false;
-    private boolean hideTitleLink = false;
-    private boolean titleValueFromPage = false;
-    private boolean descriptionValueFromPage = false;
-    private List<ListItem> ctas = new ArrayList<>();
+    private boolean actionsEnabled = false;
+    private boolean titleHidden = false;
+    private boolean descriptionHidden = false;
+    private boolean imageLinkHidden = false;
+    private boolean titleLinkHidden = false;
+    private boolean titleFromPage = false;
+    private boolean descriptionFromPage = false;
+    private List<ListItem> actions = new ArrayList<>();
     private final List<String> hiddenImageResourceProperties = new ArrayList<String>() {{
         add(JcrConstants.JCR_TITLE);
         add(JcrConstants.JCR_DESCRIPTION);
@@ -99,33 +99,30 @@ public class TeaserImpl extends AbstractImageDelegatingModel implements Teaser {
 
     @PostConstruct
     private void initModel() {
-        withCTA = properties.get(Teaser.PN_WITH_CTA, withCTA);
+        actionsEnabled = properties.get(Teaser.PN_ACTIONS_ENABLED, actionsEnabled);
 
         populateStyleProperties();
 
-        titleValueFromPage = properties.get(Teaser.PN_TITLE_VALUE_FROM_PAGE, titleValueFromPage);
-        descriptionValueFromPage = properties.get(Teaser.PN_DESCRIPTION_VALUE_FROM_PAGE, descriptionValueFromPage);
+        titleFromPage = properties.get(Teaser.PN_TITLE_FROM_PAGE, titleFromPage);
+        descriptionFromPage = properties.get(Teaser.PN_DESCRIPTION_FROM_PAGE, descriptionFromPage);
         linkURL = properties.get(ImageResource.PN_LINK_URL, String.class);
 
-        if (withCTA) {
+        if (actionsEnabled) {
             hiddenImageResourceProperties.add(ImageResource.PN_LINK_URL);
-            populateCTAs();
-            if (ctas.size() > 0) {
-                ListItem firstCTA = ctas.get(0);
-                linkURL = firstCTA.getPath();
-                if (linkURL.startsWith("/")) {
-                    linkURL = firstCTA.getPath().substring(0, linkURL.lastIndexOf('.'));
-                }
+            populateActions();
+            if (actions.size() > 0) {
+                ListItem firstAction = actions.get(0);
+                linkURL = firstAction.getPath();
             }
         }
 
         targetPage = pageManager.getPage(linkURL);
 
-        if (hideTitle) {
+        if (titleHidden) {
             title = null;
         } else {
             title = properties.get(JcrConstants.JCR_TITLE, String.class);
-            if (titleValueFromPage) {
+            if (titleFromPage) {
                 if (targetPage != null) {
                     title = StringUtils.defaultIfEmpty(targetPage.getPageTitle(), targetPage.getTitle());
                 } else {
@@ -133,11 +130,11 @@ public class TeaserImpl extends AbstractImageDelegatingModel implements Teaser {
                 }
             }
         }
-        if (hideDescription) {
+        if (descriptionHidden) {
             description = null;
         } else {
             description = properties.get(JcrConstants.JCR_DESCRIPTION, String.class);
-            if (descriptionValueFromPage) {
+            if (descriptionFromPage) {
                 if (targetPage != null) {
                     description = targetPage.getDescription();
                 } else {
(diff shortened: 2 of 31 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TitleImpl.java
```diff
@@ -19,6 +19,7 @@
 import javax.annotation.Nonnull;
 import javax.annotation.PostConstruct;
 
+import com.adobe.cq.wcm.core.components.internal.Utils;
```
richardhand: can we move this import statement to the correct position in the list?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
```diff
@@ -45,6 +45,27 @@
      */
     String NN_CTAS = "ctas";
```
richardhand: I find "ctas" quite unreadable if you're not familiar with the concept. Can't we simply refer to these as "actions".

`NN_ACTIONS="actions"`
`PN_ACTION_LINK="link"`
`PN_ACTION_TEXT="text"`
`PN_ACTION_DISABLED="disabled"`

Also, `disableCTA` sounds more like a directive than a state, `disabled` seems more suitable.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
```diff
@@ -81,18 +102,19 @@
     String PN_HIDE_DESCRIPTION = "hideDescription";
 
     /**
-     * Name of the property that will store if the description link should be hidden.
+     * Name of the property that will store if the description value should be taken from the linked page.
      *
      * @since com.adobe.cq.wcm.core.components.sandbox.models 1.0.0
      */
-    String PN_HIDE_DESCRIPTION_LINK = "hideDescriptionLink";
+    String PN_DESCRIPTION_VALUE_FROM_PAGE = "descriptionValueFromPage";
```
richardhand: `VALUE` seems redundant here, can we go with `PN_DESCRIPTION_FROM_PAGE`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/internal/models/v1/TeaserImpl.java
```diff
@@ -99,21 +100,27 @@
     @PostConstruct
     private void initModel() {
         withCTA = properties.get(Teaser.PN_WITH_CTA, withCTA);
```
richardhand: I would also call this something like `actionsEnabled` or we're sticking with CTA `ctasEnabled`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
richardhand: The description also seems wrong here, shouldn't it be "Name of the child node where the Call-to-Actions are stored" (child rather than parent)

bundles/core/src/test/resources/teaser/exporter-teaser9.json
```diff
@@ -0,0 +1,20 @@
...
+  "title": "Teaser",
+  "description": "Description",
+  "linkURL": "http://www.adobe.com",
+  "withCTA": true,
+  "hideImageLink": false,
+  "hideTitleLink": false,
+  "ctas": [
+    {
+      "title": "Adobe",
+      "path": "http://www.adobe.com"
```
richardhand: seeing a URL at a path key seems strange here.

bundles/core/src/test/resources/teaser/exporter-teaser9.json
```diff
@@ -0,0 +1,20 @@
+{
+  "title": "Teaser",
+  "description": "Description",
+  "linkURL": "http://www.adobe.com",
```
richardhand: would go with either `link` or `url` for this property rather than a combination and make it consistent with the output of the actions.

content/src/content/jcr_root/apps/core/wcm/sandbox/components/teaser/v1/teaser/teaser.html
```diff
@@ -17,17 +17,13 @@
...
-    <h2 class="cmp-teaser__title" data-sly-test.title="${teaser.title}">
-        <a class="cmp-teaser__title-link" href="${teaser.linkURL}" data-sly-unwrap="${!teaser.linkURL || withCTA || teaser.hideTitleLink}">${title}</a>
+    <h2 class="cmp-teaser__title" data-sly-test.title="${teaser.title}" data-sly-element="${teaser.titleType}">
+        <a class="cmp-teaser__title-link" href="${teaser.linkURL}" data-sly-unwrap="${!teaser.linkURL || teaser.hideTitleLink}">${title}</a>
     </h2>
-    <p class="cmp-teaser__description" data-sly-test.description="${teaser.description}">
-        <a class="cmp-teaser__description-link" href="${teaser.linkURL}" data-sly-unwrap="${!teaser.linkURL || withCTA || teaser.hideDescriptionLink}">${description}</a>
-    </p>
+    <div class="cmp-teaser__description" data-sly-test.description="${teaser.description}">${teaser.description @ context="html"}</div>
+    <div class="cmp-teaser__call-to-action-container" data-sly-test.withCTA="${teaser.withCTA}" data-sly-list="${teaser.CTAs}">
```
richardhand: I find this class name a little verbose. Proposal:
`cmp-teaser__actions`
`cmp-teaser__action`

content/src/content/jcr_root/apps/core/wcm/sandbox/components/teaser/v1/teaser/teaser.html
```diff
@@ -17,17 +17,13 @@
...
+        <a class="cmp-teaser__call-to-action-link" href="${item.path}">${item.title}</a>
```
richardhand: I wonder if we should split this template up a little to enable easier override of the sub elements. It might be common for example to want to provide a different action markup to include e.g. an icon.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
vladbailescu: `actions` is indeed easier to understand, I think it makes sense to replace `CTA(s)` with `action(s)`.

@jckautzmann also suggested we switch from directives to states for properties, we can rename them all to be consistent. We should also describe in the javadoc where the property is located (resource, style/content-policy).

The `actions` node is a parent of each action, yet a child of the teaser ;)

bundles/core/src/test/resources/teaser/exporter-teaser9.json
vladbailescu: We used the same property name as the `Image` component, it's needed if we want to delegate rendering.

bundles/core/src/test/resources/teaser/exporter-teaser9.json
vladbailescu: I made the ctas/actions a list of `ListItem`s for POC of the mockups. Perhaps we should revisit this and use a different interface.

content/src/content/jcr_root/apps/core/wcm/sandbox/components/teaser/v1/teaser/teaser.html
vladbailescu: I agree. How granular do you think we should be here?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
richardhand: +1 - we should review these all at the same time and ensure we're taking the same approach cross-component.

content/src/content/jcr_root/apps/core/wcm/sandbox/components/teaser/v1/teaser/teaser.html
richardhand: I would be tempted to go fully granular:
```
teaser.html
image.html
title.html
description.html
actions.html
action.html
```

So, an override of `teaser.html` would then simply be about reordering the constituent parts (with may be common in order to achieve a particular layouts which may not be easily achieved with CSS alone) or to introduce some additional grouping (e.g. title and description under a "copy" element).

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/internal/models/v1/TeaserImpl.java
```diff
@@ -98,39 +99,42 @@
...
+        titleFromPage = properties.get(Teaser.PN_TITLE_FROM_PAGE, titleFromPage);
+        descriptionFromPage = properties.get(Teaser.PN_DESCRIPTION_FROM_PAGE, descriptionFromPage);
+        linkURL = properties.get(ImageResource.PN_LINK_URL, String.class);
+
+        if (actionsEnabled) {
+            hiddenImageResourceProperties.add(ImageResource.PN_LINK_URL);
+            populateActions();
+            if (actions.size() > 0) {
+                ListItem firstAction = actions.get(0);
+                linkURL = firstAction.getPath();
```
jckautzmann: I'd add a null check.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/internal/models/v1/TeaserImpl.java
```diff
@@ -166,26 +170,35 @@ private void initModel() {
...
         }
     }
 
-    private void populateCTAs() {
-        Resource ctasNode = resource.getChild(Teaser.NN_CTAS);
-        if (ctasNode != null) {
-            for(Resource cta : ctasNode.getChildren()) {
-                ctas.add(new ListItem() {
+    private void populateActions() {
+        Resource actionsNode = resource.getChild(Teaser.NN_ACTIONS);
```
jckautzmann: I'd call it `actionsRes` or `actionsResource` as it is a `Resource`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
```diff
@@ -36,71 +36,93 @@
...
+     * Name of the child node where the Call-to-Action elements are stored
      *
      * @since com.adobe.cq.wcm.core.components.sandbox.models 1.0.0
      */
-    String NN_CTAS = "ctas";
+    String NN_ACTIONS = "actions";
 
     /**
-     * Name of the property that will store if the image link should be hidden.
+     * Name of the resource property that stores the Call-to-Action link
```
jckautzmann: I'd remove `resource`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
```diff
@@ -36,71 +36,93 @@
...
      *
      * @since com.adobe.cq.wcm.core.components.sandbox.models 1.0.0
      */
-    String PN_HIDE_IMAGE_LINK = "hideImageLink";
+    String PN_ACTION_LINK = "link";
 
     /**
-     * Name of the property that will store if the title should be hidden.
+     * Name of the resource property that stores the Call-to-Action text
```
jckautzmann: I'd remove `resource`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
```diff
@@ -36,71 +36,93 @@
...
+     * Name of the policy property that stores if the Call-to-Action is disabled
      *
      * @since com.adobe.cq.wcm.core.components.sandbox.models 1.0.0
      */
-    String PN_HIDE_TITLE_LINK = "hideTitleLink";
+    String PN_ACTIONS_DISABLED = "disabled";
 
     /**
-     * Name of the property that will store if the title value should be taken from the linked page.
+     * Name of the policy property that will store if the image link should be hidden.
```
jckautzmann: I'd use present form: that will store -> that stores

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
```diff
@@ -36,71 +36,93 @@
...
-     * Name of the property that will store if the description value should be taken from the linked page.
+     * Name of the resource property that will store if the title value should be taken from the linked page.
      *
      * @since com.adobe.cq.wcm.core.components.sandbox.models 1.0.0
      */
-    String PN_DESCRIPTION_VALUE_FROM_PAGE = "descriptionValueFromPage";
+    String PN_TITLE_FROM_PAGE = "titleFromPage";
+
+    /**
+     * Name of the policy property that will store if the description should be hidden.
```
jckautzmann: I'd rename:
Name of the policy property that stores if the description is hidden.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
```diff
@@ -36,71 +36,93 @@
...
+     *
+     * @since com.adobe.cq.wcm.core.components.sandbox.models 1.0.0
+     */
+    String PN_DESCRIPTION_HIDDEN = "descriptionHidden";
+
+    /**
+     * Name of the resource property that will store if the description value should be taken from the linked page.
```
jckautzmann: see above

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
```diff
@@ -36,71 +36,93 @@
...
+     *
+     * @since com.adobe.cq.wcm.core.components.sandbox.models 1.0.0
+     */
+    String PN_DESCRIPTION_FROM_PAGE = "descriptionFromPage";
+
+    /**
+     * Name of the policy property that will store the value for this title's HTML element type.
```
jckautzmann: see above

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/internal/models/v1/TeaserImpl.java
```diff
@@ -98,39 +99,42 @@
...
+                linkURL = firstAction.getPath();
             }
-        } else {
-            targetPage = pageManager.getPage(linkURL);
         }
-        if (hideTitle) {
+
+        targetPage = pageManager.getPage(linkURL);
+
+        if (titleHidden) {
```
richardhand: @vladbailescu, @andreeadracea - so we render the elements if they're marked hidden, but simply don't provide any text content? Shouldn't we simply not render the element at all?

vladbailescu: I don't think it's necessary to expose both if the title/description are hidden and their value from the model. It's enough to remove the value as rendering will take care of not showing anything, see https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/blob/910a3060ffaa728cc98be7d1ba1c482fe5cba562/content/src/content/jcr_root/apps/core/wcm/sandbox/components/teaser/v1/teaser/title.html#L18: `data-sly-test.title="${teaser.title}"`

richardhand: I missed the `data-sly-test`, looks good

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
jckautzmann: I was a bit confused as we are sometimes mixing Node and Resource terminology: e.g. `node name` and here `resource property`. Maybe we should start using `page property` as opposed to `policy property`?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/sandbox/models/Teaser.java
vladbailescu: @jckautzmann & @richardhand, please check https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/pull/179/commits/172afc21185965b6cd070f4cb991cc59ceb2fd4b

content/src/content/jcr_root/apps/core/wcm/sandbox/components/teaser/v1/teaser/_cq_dialog/.content.xml
```diff
@@ -22,26 +22,26 @@
...
                                     value="{Boolean}true"
                                     uncheckedValue="{Boolean}false"
-                                    granite:hide="${cqDesign.disableCTA}">
-                            </withCTA>
-                            <ctas
+                                    granite:hide="${cqDesign.disabled}">
+                            </actionsEnabled>
+                            <actions
                                     jcr:primaryType="nt:unstructured"
                                     sling:resourceType="granite/ui/components/coral/foundation/form/multifield"
```
richardhand: @vladbailescu, @andreeadracea - I see a few UX issues with the teaser dialog.

* Actions multifield
  * It's not easy to distinguish each action - there's just a vertical list of text inputs with equal weight. _proposal:_ Group the action inputs better, perhaps putting the link and text inputs on one line (see how we did this with the style system policy dialog).
  * It takes up a lot of vertical space and can expand downwards, pushing the other key options out of view. _proposal:_ Move the multifield to its own "Actions" tab which gets shown / hidden when selecting "with Call-To-Action(s)".
* Link
  * hiding this on enabling "with Call-To-Action(s)" isn't in line with what we do for fetching title/description from the page. _proposal_ disable it instead of hiding and move the "with Call-To-Action(s)" field below the link field.
* Description
  * The input background is always gray, even if the field isn't disabled (as a result of fetching it from the page). _expected:_ has a white background when it's enabled.

screen shot 2018-05-23 at 14 25 57

content/src/content/jcr_root/apps/core/wcm/sandbox/components/teaser/v1/teaser/_cq_dialog/.content.xml
```diff
@@ -22,46 +22,47 @@
...
-                                    name="./withCTA"
+                                    name="./actionsEnabled"
                                     value="{Boolean}true"
                                     uncheckedValue="{Boolean}false"
-                                    granite:hide="${cqDesign.disableCTA}">
-                            </withCTA>
-                            <ctas
+                                    granite:hide="${cqDesign.disabled}">
+                            </actionsEnabled>
+                            <actions
```
richardhand: There's some strange behavior with the validation of the link / multifield. 

If I:
* Allow actions via the "with Call-To-Actions" checkbox (I'm informed that the first link of the first action is required).
* Disable actions again.
-> Result: It's not possible to confirm the dialog any longer
-> Expectation: Can always confirm the dialog if there is no link and actions are disabled

richardhand:
@vladbailescu, @andreeadracea - I made some improvements to the dialog labels / descriptions in https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/pull/179/commits/29a4d61805cc734b324bd03f3833c6316b83d5ec

vladbailescu:
@richardhand That looks neat but it's a far stretch from what we can build at the moment. Also, no more feature creep please! ;)

PR #181, opened 2018-05-15
[Image] Image map support for image component (#35)

### LINKED ISSUE #35: [Image] Image map not supported

richardhand:
- adds ImageArea API for outputting image map areas
- adds getAreas API to Image Component
- documents ./imageMap property format
- parses ./imageMap property format
- adds support for responsive resize of image maps in the JavaScript component
- enables image map editing feature via the edit config
- updates Java tests

(fixes #35)


bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImplTest.java
```diff
@@ -52,6 +55,15 @@ public void testExportedType() {
         assertEquals(ImageImpl.RESOURCE_TYPE, image.getExportedType());
     }
 
+    @Test
```
vladbailescu: Could you please move the `testImageWithNoSmartSize` and `testImageWithTwoOrMoreSmartSizes` methods back to their original position? They have no changes AFAICT

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -117,6 +116,12 @@
      */
     String PN_FLIP_VERTICAL = "imageFlipVertical";
 
+    /**
+     * Name of the resource property that defines areas of an image map.
+     *
+     * @since com.adobe.cq.wcm.core.components.models 12.4.0
+     */
+    String PN_MAP = "imageMap";
```
vladbailescu: It would make more sense to have this as a plural, since it supports multiple maps. We should also document the storage format for this property and this seems like the best place for it.

richardhand: @vladbailescu - this was deliberate to match the HTML markup definition as well as the property name itself. A single `map` has multiple `area`.

vladbailescu: Got it, thx!

vladbailescu: Also, I can see the format described at https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/pull/181/files#diff-561deebea47b85bc8ffddf78fe11602dR146 but for me it would make more sense to have it in the interface, as it's public.

PR #194, opened 2018-05-28
Added config option to enable toolbar items of the Image IPE only for…

### LINKED ISSUE #193: [Image] Disable IPE toolbar items for certain mime types
After the added support for SVG images in the core components image component (#171) we have to find a way to disable toolbar items of the IPE (Inplace Editor) which are currently not supported for certain mime types. For example cropping of an SVG image is currently not possible and should therefore not be provided in the IPE.

Note: This config option will only be taken into consideration after a new feature pack is available and is installed on top of AEM 6.4.

bpauli:
… specific mime types

fixes #193

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/_cq_editConfig.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/_cq_editConfig.xml
@@ -23,7 +23,7 @@
             <plugins jcr:primaryType="nt:unstructured">
                 <crop
                     jcr:primaryType="nt:unstructured"
-                    enabled="[image/gif,image/jpeg,image/png,image/webp,image/tiff]"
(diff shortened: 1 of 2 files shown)


content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/_cq_editConfig.xml
```diff
@@ -45,12 +46,14 @@
                 </crop>
                 <flip
                     jcr:primaryType="nt:unstructured"
+                    enabled="[image/gif,image/jpeg,image/png,image/webp,image/tiff]"
                     features="-"/>
                 <map
                     jcr:primaryType="nt:unstructured"
```
richardhand: Do image maps function okay on top of an SVG image? I suppose they should.

PR #196, opened 2018-05-29
[Image] Add support for SVG images

jckautzmann:
[Image] Add support for SVG images

- fix the broken thumbnail in the file upload widget when the image is an SVG


jckautzmann:
After further discussions, we decided that it was better to fix the issue in AEM.

Issue #216, opened 2018-06-21
[Page] Support loading configurable JS ClientLib in page head element


kwin:
The clientlib category being maintained in the template (via a content policy) leads to loading the according JS in the body area (compare with https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/blob/903a54005aae0d3cd071bf7715a6ce1852f5f317/content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/footer.html#L19 and https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/README.md#loading-of-cssjs). Although this is a good default and makes sense for most of the javascript, there might be some scripts, which necessarily need to be loaded in the head section. Currently that requires overwriting HTL scripts. 

It would be nice if there would be an additional category to maintain on the template level, which is getting loaded in the head section (JS only).

gabrielwalt:
[Page] Support loading configurable JS ClientLib in page head element (CQ-4247319)
As a template author, I want to configure JS that can optionally be loaded in the page head element, instead of the body element, so that I can accommodate scripts that must be loaded upfront, before the page loads.

Acceptance criteria:
In the page policy (design) dialog, allow to define client libraries of which the JS should also be loaded in the page head element (next to the CSS inclusion).

PR #244, opened 2018-07-10

Samyak1080:
| Fixed Issues?            | #216 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
@@ -14,16 +14,16 @@
  ~ limitations under the License.
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.internal.models.v2;
+import com.day.cq.wcm.api.Template;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.PostConstruct;
 
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.models.annotations.Exporter;
@@ -56,6 +56,8 @@ public class PageImpl extends com.adobe.cq.wcm.core.components.internal.models.v
     public static final String PN_REDIRECT_TARGET = "cq:redirectTarget";
 
     private Boolean hasCloudconfigSupport;
+    protected String[] headClientLibCategories = new String[0];
+    protected static final String PN_HEADCLIENTLIBS = "headclientlibs";
 
     @OSGiService
     private HtmlLibraryManager htmlLibraryManager;
@@ -89,6 +91,7 @@ public class PageImpl extends com.adobe.cq.wcm.core.components.internal.models.v
             }
         }
         setRedirect();
+        populateHeadClientlibCategories();
     }
 
     private void setRedirect() {
@@ -116,9 +119,24 @@ public class PageImpl extends com.adobe.cq.wcm.core.components.internal.models.v
         return path;
     }
 
+
+    protected void populateHeadClientlibCategories() {
+        if (currentStyle != null) {
+                headClientLibCategories = currentStyle.get(PN_HEADCLIENTLIBS, ArrayUtils.EMPTY_STRING_ARRAY);
+        }
+    }
+
+
     @Override
     protected void loadFavicons(String designPath) {
     }
+
+    @Override
+    @JsonIgnore
+    public String[] getHeadClientLibCategories() {
+        return Arrays.copyOf(headClientLibCategories, headClientLibCategories.length);
+    }
+
 
     @Override
     public Map<String, String> getFavicons() {
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
@@ -14,10 +14,13 @@
  ~ limitations under the License.
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.internal.models.v2;
-import com.day.cq.wcm.api.Template;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -57,7 +60,7 @@ public class PageImpl extends com.adobe.cq.wcm.core.components.internal.models.v
 
     private Boolean hasCloudconfigSupport;
     protected String[] headClientLibCategories = new String[0];
-    protected static final String PN_HEADCLIENTLIBS = "headclientlibs";
+    protected static final String PN_HEAD_CLIENTLIBS = "headClientlibs";
 
     @OSGiService
     private HtmlLibraryManager htmlLibraryManager;
@@ -90,8 +93,8 @@ public class PageImpl extends com.adobe.cq.wcm.core.components.internal.models.v
                 appResourcesPath = getProxyPath(clientLibraryList.get(0));
             }
         }
-        setRedirect();
         populateHeadClientlibCategories();
+        setRedirect();
     }
 
     private void setRedirect() {
@@ -122,10 +125,9 @@ public class PageImpl extends com.adobe.cq.wcm.core.components.internal.models.v
 
     protected void populateHeadClientlibCategories() {
         if (currentStyle != null) {
-                headClientLibCategories = currentStyle.get(PN_HEADCLIENTLIBS, ArrayUtils.EMPTY_STRING_ARRAY);
+            headClientLibCategories = currentStyle.get(PN_HEAD_CLIENTLIBS, ArrayUtils.EMPTY_STRING_ARRAY);
         }
     }
-
 
     @Override
     protected void loadFavicons(String designPath) {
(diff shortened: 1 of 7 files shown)


content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/_cq_design_dialog/.content.xml
```diff
@@ -40,15 +40,16 @@
                                 jcr:primaryType="nt:unstructured"
                                 sling:resourceType="granite/ui/components/coral/foundation/heading"
                                 level="{Long}4"
-                                text="Client-Side Libraries"/>
+                                text="Client-Side Libraries for Body element"/>
```
kwin: This text is not 100% correct. CSS having that category is still loaded in the head (https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/blob/8912754d5dd0a90c99b0191268dfe87878a4145c/content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/headlibs.html#L28).

content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/_cq_design_dialog/.content.xml
```diff
@@ -57,6 +58,28 @@
                                     </clientlibs>
                                 </items>
                             </well>
+                            <heading1
+                                jcr:primaryType="nt:unstructured"
+                                sling:resourceType="granite/ui/components/coral/foundation/heading"
+                                level="{Long}4"
+                                text="Client-Side Libraries for Head element"/>
```
kwin: What about CSS clientlibs? Currently this is not loaded. This should be clarified at least in the description.

kwin:
Can you please also adjust the readme for the page component which lists in a table which categories are loaded where?

znikolovski:
Maybe I'm misunderstanding the purpose of this PR but isn't this already enabled via the Page Policy Client-Side Libraries property?

richardhand:
@znikolovski - this PR is about allowing additional configuration in the Page Policy for defining JS to be add to the head. Currently everything is added to body end. See https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/issues/216

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -14,16 +14,16 @@
  ~ limitations under the License.
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.internal.models.v2;
+import com.day.cq.wcm.api.Template;
```
richardhand: @Samyak1080 - please add these imports to the correct position in the list

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -14,16 +14,16 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
richardhand: Would rather not use a wildcard, can we list the classes we use

znikolovski:
@richardhand thanks for the clarification.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -56,6 +56,8 @@
     public static final String PN_REDIRECT_TARGET = "cq:redirectTarget";
 
     private Boolean hasCloudconfigSupport;
+    protected String[] headClientLibCategories = new String[0];
+    protected static final String PN_HEADCLIENTLIBS = "headclientlibs";
```
richardhand: • Perhaps this property could go to the Page model
• can we camelCase it "headClientlibs"

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -89,6 +91,7 @@ protected void initModel() {
             }
         }
         setRedirect();
+        populateHeadClientlibCategories();
```
richardhand: this one is picky, but could we move this before `setRedirect` so we have the clientlib related logic together.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -116,10 +119,25 @@ private String getProxyPath(ClientLibrary lib) {
...
     }
 
+
+    protected void populateHeadClientlibCategories() {
+        if (currentStyle != null) {
+                headClientLibCategories = currentStyle.get(PN_HEADCLIENTLIBS, ArrayUtils.EMPTY_STRING_ARRAY);
+        }
+    }
+
+
```
richardhand: let's not introduce extra lines between functions, single line is enough

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -227,7 +227,7 @@ default String getTitle() {
     }
 
     /**
-     * If this page is associated with a Template, then this method will return the Template's client libraries categories.
+     * If this page is associated with a Template, then this method will return the Template's client libraries categories for the body tag.
```
richardhand: * client libraries categories -> client library categories
* would be more specific "client library categories to be included before the body end tag"

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -238,6 +238,18 @@ default String getTitle() {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * If this page is associated with a Template, then this method will return the Template's client libraries categories for the head tag.
```
richardhand: • client libraries categories -> client library categories
• "client library categories included in the page head"

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -238,6 +238,18 @@ default String getTitle() {
...
+     *
+     * @return an array of client libraries categories; the array can be empty if the page doesn't have an associated template or if the
+     * template has no client libraries
```
richardhand: Would communicate the fact that these are head client libraries in the return doc

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -227,7 +227,7 @@ default String getTitle() {
     }
 
     /**
-     * If this page is associated with a Template, then this method will return the Template's client libraries categories.
+     * If this page is associated with a Template, then this method will return the Template's client libraries categories for the body tag.
      *
      * @return an array of client libraries categories; the array can be empty if the page doesn't have an associated template or if the
```
richardhand: Would communicate the fact that these are body client libraries in the return doc

bundles/core/src/test/resources/page/v2/test-conf.json
```diff
@@ -36,6 +36,9 @@
                                 "clientlibs": [
                                     "coretest.product-page"
                                 ],
+                                "headclientlibs": [
```
richardhand: headClientlibs

content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/_cq_design_dialog/.content.xml
richardhand: +1 we should make it clear that this is JS related, also in the java docs

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -116,10 +119,25 @@ private String getProxyPath(ClientLibrary lib) {
         return path;
     }
 
+
+    protected void populateHeadClientlibCategories() {
```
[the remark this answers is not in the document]
richardhand: @znikolovski - agree, what would your expectation be here for a clash? I think I would expect the JS for a clientlib defined in both fields to be added to the head but not the body.

content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/_cq_design_dialog/.content.xml
```diff
@@ -49,7 +49,7 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/form/multifield"
                                         class="foundation-layout-util-maximized-alt"
-                                        fieldDescription="The client-side libraries that are to be loaded in the body element.">
+                                        fieldDescription="The client-side javascript libraries that are to be loaded in the body element.">
```
kwin: This is still  not correct, as these may also be CSS libraries (loaded in the head)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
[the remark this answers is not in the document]
znikolovski: @richardhand that's what I was thinking - I almost consider this an "explicit" request to put the clientlib in the head so it should win out.

Wouldn't this also mean that the populateClientlibCategories() method should also be overriden to implement the check?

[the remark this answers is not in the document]
Samyak1080: @richardhand I feel what you're saying is right and makes a lot of sense as loading the js in the head is the better option. In that case we won't be needing the field for loading the js in the body. Also, we'll need to make adjustments to the HTL call since currently it's calling the clientlibs in the body element as well.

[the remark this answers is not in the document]
richardhand: It's not that loading in the head is a better option, rather we want to provide the option to "explicitly" do so. We still need a general clientlib definition (both CSS and JS) with body end being the default.

content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/_cq_design_dialog/.content.xml
vladbailescu: I too think it's best to allow CSS as well in the head (with checks to avoid loading the same clientlib twice)

kwin: Also for backwards-compatibility reasons this clientlib must be loaded for type CSS in the head (even in the future).

richardhand:
New PR at #289, closing here.

PR #249, opened 2018-07-24
[Navigation] Exclude root levels options #151

### LINKED ISSUE #151: [Navigation] Exclude root levels options
**Requested feature**

It would be nice to have a level option in the Navigation component's dialog/design dialog that uses [page.getAbsoluteParent(int)](https://helpx.adobe.com/experience-manager/6-3/sites/developing/using/reference-materials/javadoc/com/day/cq/wcm/api/Page.html#getAbsoluteParent(int)) for rendering the navigation items.

**Example**

An author specifies level 3 in the navigation component's dialog and we have the following pages:

/content/my-site/en/foo/foo1
/content/my-site/en/foo/foo2
/content/my-site/en/anotherfoo/anotherfoo1
/content/my-site/en/anotherfoo/anotherfoo2

In this case, foo and anotherfoo are level 3 pages.

On the /content/my-site/en page, the navigation should render nothing.

On the /content/my-site/en/foo page, the navigation component should render:
foo1
foo2

On the /content/my-site/en/anotherfoo page, the navigation component should render:
anotherfoo1
anotherfoo2

On the /content/my-site/en/anotherfoo/anotherfoo1 page (or deeper), the navigation component should still render:
anotherfoo1
anotherfoo2

Thanks in advance

Shivanshmathur:
| Fixed Issues?            | Fixes #151
| Patch: Bug Fix?          |
| Major: Breaking Change?  |


vladbailescu:
@Shivanshmathur, do we really need a new version here? The HTML markup will be the same, the model interface does not change in a breaking way, and there are no new tabs added to the dialogs.

We should keep the current version and make sure to implement the new `rootLevel` behaviour in such a way that older instances of the component render don't get broken.

vladbailescu:
A `skipNavigationRoot==true` is equivalent to `rootLevel==1` while `skipNavigationRoot==false` is equivalent to `rootLevel==0`, right? I'm sure we can implement the change to take `skipNavigationRoot` into consideration when `rootLevel` is not defined.

The implementation is not public/API. We can use whatever variables we need. Like I mentioned, I would use the `rootLevel` concept in the implementation but also fallback to setting the value based on `skipNavigationRoot` for compatibility.

Shivanshmathur:
@vladbailescu Done.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/NavigationImpl.java
```diff
@@ -85,6 +86,7 @@
     private String navigationRootPage;
     private List<NavigationItem> items;
     private boolean skipNavigationRoot;
```
vladbailescu: Can we reduce the scope of this variable and only define it locally in `initModel()`?

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/README.md
```diff
@@ -31,15 +31,15 @@ The Navigation component uses the `com.adobe.cq.wcm.core.components.models.Navig
 The following configuration properties are used:
 
 1. `./navigationRoot` - the root page from which to build the navigation. Can be a blueprint master, language master or regular page.
-2. `./skipNavigationRoot` - if `true`, excludes the navigation root in the resulting tree, including its descendants only.
```
vladbailescu: I would rather we mark this as deprecated and and replaced by `rootLevel` instead of removing it.

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_design_dialog/.content.xml
```diff
@@ -47,15 +47,16 @@
...
-                                checked="{Boolean}true"
-                                uncheckedValue="{Boolean}false"/>
+                                sling:resourceType="granite/ui/components/coral/foundation/form/numberfield"
+                                fieldDescription="The levels below navigation root that are to be excluded. To include navigation root enter 0."
+                                fieldLabel="Exclude Root Levels"
+                                max="100"
+                                min="0"
+                                name="./rootLevel"
+                                step="1"
+                                value="1"/>
```
vladbailescu: Can we set the value here based on `skipNavigationRoot` (if that was defined previously)?

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
```diff
@@ -57,15 +57,16 @@
...
-                                                checked="${not empty cqDesign.skipNavigationRoot ? cqDesign.skipNavigationRoot : true}"
-                                                uncheckedValue="{Boolean}false"/>
+                                                sling:resourceType="granite/ui/components/coral/foundation/form/numberfield"
+                                                fieldDescription="The levels below navigation root that are to be excluded. To include navigation root enter 0."
+                                                fieldLabel="Exclude Root Levels"
+                                                max="100"
+                                                min="0"
+                                                name="./rootLevel"
+                                                step="1"
+                                                value="${not empty cqDesign.rootLevel ? cqDesign.rootLevel : 1}"/>
```
vladbailescu: Same comment as above: would be great to be able to take `skipNavigationRoot` into consideration, if already defined.

vladbailescu:
The build failure seems unrelated to the changes. I cleaned up the maven caches and retriggered the build.

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_design_dialog/.content.xml
vladbailescu: I wasn't referring to that, just that if you previously had selected `skipNavigationRoot==false` then you should see `rootLevel==0` in the dialog.

Shivanshmathur: Got it.

PR #252, opened 2018-07-30
[Image] Enable asset tracking on image component

### LINKED ISSUE #238: [Image] Enable asset tracking on image component
For asset insights we need to know the asset-id (uuid) used by the image component, so that Analytics can measure the number of views that the individual assets received.

We should add an attribute like the following to the markup generated by the image component:
`data-cq-asset-id="0ad999e0-4886-4e7a-b264-4a751a3bbaae"`

(CQ-4205936)

Samyak1080:
| Fixed Issues?            | Fixes #238 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

For the time being, I have commented out testGetUuid() method in ImageImplTest.java because it uses getUuid() method which receives the UUID from getID() method(when the UUID field is enabled in the design dialog). The getID() method was previously throwing "UnsupportedOperationException" which is why the testGetUuid() method produced errors and therefore, I had to set the 'uuidDisabled' field as true for the images in their exporter files. As recommended by @vladbailescu, I have raised a PR in [wcm-io-testing](https://github.com/wcm-io/wcm-io-testing) repository after providing a concrete implementation for the getID() method in MockAsset.java file. Below are the links to the issue I created on wcm-io-testing platform and the required PR on github:-
Issue: https://wcm-io.atlassian.net/projects/WTES/issues/WTES-32?filter=allissues&orderby=priority%20DESC 
PR: https://github.com/wcm-io/wcm-io-testing/pull/8

As soon as the PR for testing gets merged, the testGetUuid() method in ImageImplTest.java can be used as a working function for testing if the UUID generated is correct.


content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/image.html
```diff
@@ -16,6 +16,7 @@
 <div data-sly-use.image="com.adobe.cq.wcm.core.components.models.Image"
      data-sly-use.templates="core/wcm/components/commons/v1/templates.html"
      data-sly-test="${image.src}"
+     data-cq-asset-id="${image.uuid}"
```
kwin: can't we finally get rid of the term "cq"?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
```diff
@@ -52,23 +57,33 @@
...
         boolean titleValueFromDAM = properties.get(PN_TITLE_VALUE_FROM_DAM, currentStyle.get(PN_TITLE_VALUE_FROM_DAM, true));
         displayPopupTitle = properties.get(PN_DISPLAY_POPUP_TITLE, currentStyle.get(PN_DISPLAY_POPUP_TITLE, true));
+        uuidDisabled = currentStyle.get(PN_DISABLE_UUID, false);
         if (StringUtils.isNotEmpty(fileReference)) {
             // the image is coming from DAM
             final Resource assetResource = request.getResourceResolver().getResource(fileReference);
             if (assetResource != null) {
                 Asset asset = assetResource.adaptTo(Asset.class);
                 if (asset != null) {
+                    if(!uuidDisabled) {
```
vladbailescu: Please add a space after `if`: `if (`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -52,6 +52,13 @@
      */
     String PN_IS_DECORATIVE = "isDecorative";
 
+    /**
+     * Name of the policy property that defines whether or not the UUID is disabled.
+     *
+     * @since com.adobe.cq.wcm.core.components.models 12.4.0
+     */
+    String PN_DISABLE_UUID = "uuidDisabled";
```
richardhand: can we call this PN_UUID_DISABLED

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -162,6 +169,26 @@ default String getTitle() {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Returns the value for the image's {@code uuid} attribute, if one was set.
```
richardhand: I wouldn't mention the HTML attribute here, it's markup implementation specific.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -162,6 +169,26 @@ default String getTitle() {
...
+     *
+     * @return the value for the image's {@code uuid} attribute, if one was set, or {@code null}
+     * @since com.adobe.cq.wcm.core.components.models 11.0.0; marked <code>default</code> in 12.4.0
```
richardhand: These since versions look wrong to me

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -162,6 +169,26 @@ default String getTitle() {
...
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Checks if UUID is disabled on the asset.
+     *
+     * @return {@code true} if UUID is disabled on the asset, {@code false} otherwise
+     * @since com.adobe.cq.wcm.core.components.models 12.4.0
+     */
+    default boolean isUuidDisabled() {
```
richardhand: What do we need this API for? Can't we just determine this from a null UUID?

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/image.html
richardhand: Thanks @kwin - I agree, we already have `data-asset`, `data-title` so I would continue with `data-asset-id`, I don't see a real win with a namespace. 

I would also move this attribute below `data-asset`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
Samyak1080: This is used to give user the option of disabling the uuid field from the design dialog.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImplTest.java
```diff
@@ -56,6 +56,21 @@ public void testExportedType() {
         assertEquals(ImageImpl.RESOURCE_TYPE, image.getExportedType());
     }
 
+    /*
+    @Test
```
richardhand: @Samyak1080 - do you plan to reenable this test case?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
richardhand: @Samyak1080 - yes we have that option in the design dialog, I'm just wondering if we really need to provide it as API in the Image model, it's not used and is an implementation detail.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImplTest.java
richardhand: What code do we need in wcm-io-testing?

Samyak1080: https://github.com/wcm-io/wcm-io-testing/blob/develop/aem-mock/src/main/java/io/wcm/testing/mock/aem/MockAsset.java#L215
As explained in PR description, this implementation of getID() method did not exist before. Now that it has been implemented, it will be included in the upcoming released version 2.2.16 of AEM Mocks. Once this version is released, we can update the dependency in our POM and re-enable this test case.

richardhand: Thanks for clarifying @Samyak1080!

richardhand:
@Samyak1080 - will you have any time to update the PR now that [aem-mock-2.2.16](https://github.com/wcm-io/wcm-io-testing/releases/tag/io.wcm.testing.aem-mock-2.2.16) is released? If not, I can do it.

Samyak1080:
@richardhand - Unfortunately, I'll be able to reference the update only after a couple of days. I think it would be in the best of our interest if it is done as early as possible. I apologise for not being able to do it within due time.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImplTest.java
bpauli: Unfortunately the update version of AEM Mocks (2.2.16) doesn't return a proper `jcr:uuid`. I created https://wcm-io.atlassian.net/browse/WTES-33 for this issue

bpauli:
I created a new fork of this PR at [1] to enable the JUnit test requested by @richardhand 

[1] https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/pull/282

PR #287, opened 2018-09-11
Disable Lazy load of Image if the image has lazyLoad attribute

shivendra096:
- Export a site containing image component using .export.zip method

 - Extract the downloaded zip and open the .html file in your browser

 - Observe that the images are not displayed in the site even though they are present in the exported zip folder

 - Site Image links are written in a temporary path '/var/designs/' and written in the site before exporting the site and have a lazy Load disable attribute set in them

 - Page Exporter exports the images of the site in /var/designs/ directory in the zip

 - Image js again rewrites the links and thus breaking the img element upon page load

 - Don't rewrite the links upon page load if the image doesn't allow lazy load

 - CQ-4241179

| Fixed Issues?            | CQ-4241179
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Tests Added + Pass?      |
| Any Dependency Changes?  |

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
@@ -200,7 +200,7 @@
             var imageElement = temporaryDocument.querySelector(selectors.image);
             // Remove src if the src image referred is not a temp image file
             var srcAttribute = imageElement.getAttribute("src");
-            if (srcAttribute && srcAttribute.indexOf(TEMP_DESIGN_PATH) == -1) {
+            if (srcAttribute && srcAttribute.indexOf(TEMP_DESIGN_PATH) === -1) {
                 imageElement.removeAttribute("src");
             }
             that._elements.container.insertBefore(imageElement, that._elements.noscript);
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
@@ -20,7 +20,7 @@
     var IS = "image";
 
     var EMPTY_PIXEL = "data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7";
-    var TEMP_DESIGN_PATH = "/var/designs/";
+    var LAZY_LOAD_DISABLED_ATTR = "disablelazyload";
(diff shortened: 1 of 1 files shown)


shivendra096:
@vladbailescu could you please review

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
```diff
@@ -196,6 +198,11 @@
             // temporary document avoids requesting the image before removing its src
             var temporaryDocument = parser.parseFromString(markup, "text/html");
             var imageElement = temporaryDocument.querySelector(selectors.image);
+            // Remove src if the src image referred is not a temp image file
+            var srcAttribute = imageElement.getAttribute("src");
+            if (srcAttribute && srcAttribute.indexOf(TEMP_DESIGN_PATH) == -1) {
```
richardhand: @shivendra096 - this clause is redundant, the attribute is still removed on the next line

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
```diff
@@ -20,6 +20,7 @@
     var IS = "image";
 
     var EMPTY_PIXEL = "data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7";
+    var TEMP_DESIGN_PATH = "/var/designs/";
```
richardhand: This is implementation specific, would rather find a general solution.

vladbailescu:
I think a better solution would be to update the exporter to rewrite the `data-cmp-src` as well or disable the smart image size loading completely.

shivendra096:
@vladbailescu i have updated the implementation to check for the attribute rather than a path. 
Could you please review it again ?

vladbailescu:
@shivendra096 I'm not sure what you're trying to do. It seems that you are trying to disable smart loading of images. That happens automatically when lazy loading is disabled and there's only one allowed image width.

Lazy loading means deferring image load to when it enters the viewport, it's only one aspect of smart loading.

Like I said before, ideally you would also rewrite the `data-cmp-src` attribute from the exporter. If that is not possible I think you could force a disablement of smart loading by removing the `data-cmp-lazy` and reducing image width to only the exported image (via `data-cmp-widths`).

I think there's no need to introduce a new attribute that is only used by the exporter.

vladbailescu:
@shivendra096 , @vikramsr: closing this as this is not the proper fix and there was no answer to previous comment describing recommended alternatives.

PR #289, opened 2018-09-18

### LINKED ISSUE #216: [Page] Support loading configurable JS ClientLib in page head element

richardhand:
This builds on the original pull request at #244 with the following details:
* maintain a `clientlibsJsHead` property which allows definition of clientlibs for which JavaScript is loaded in the page head.
* the existing `clientlibs` property still contains all clientlib categories defined in the original input.
* new `getClientLibCategoriesJsBody` and `getClientLibCategoriesJsHead` API allow output of the JavaScript client libraries to the correct page location.
* adds a client library category datasource and autocomplete input to make selection of client libraries more user friendly and applies it to all clientlib fields ("Client Libraries", "Page head Client Libraries" and "Web Resource Client Library").

screen shot 2018-09-24 at 22 04 27

| Fixed Issues?            | Fixes #216 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation Provided   | 👍 (code comments and or markdown)
| Any Dependency Changes?  |


richardhand:
@kwin - if you have time would also be great to get your opinion on this updated PR :)

kwin:
The description is still a bit confusing to me. What is with css clientlibs defined via „clientlibs“. They are still loaded in head, right?

richardhand:
@kwin, yes for both fields CSS is loaded to the head. This is the "clientlibs" field description

screen shot 2018-09-18 at 17 32 58

kwin:
Actually after having a 2nd look I think this is clear enough now. Thanks for taking it up.

content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/headlibs.html
```diff
@@ -13,7 +13,7 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
-<template data-sly-template.headlibs="${@ designPath, staticDesignPath, clientlibCategories, hasCloudconfigSupport}"
```
vladbailescu: I think this has the potential of breaking other people's existing customizations, since we are changing the order of parameters. It's probably safer to add the new option at the end.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -166,4 +187,18 @@ public boolean hasCloudconfigSupport() {
         }
         return hasCloudconfigSupport;
     }
+
+    protected void filterClientlibCategories() {
```
vladbailescu: The logic here seems a bit over-complicated. What I would do instead is:
1. add all elements of `clientLibCategories` to a `LinkedHashSet`
2. iterate over the `headClientLibCategories` (no need to convert it to `List`, an array can be iterated as well)
3. remove from the set all matching head client libraries
4. replace the `clientLibCategories` with the contents of the set

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/ClientLibraryCategoriesDataSourceServlet.java
```diff
@@ -0,0 +1,126 @@
...
+        service = { Servlet.class },
+        property = {
+                "sling.servlet.resourceTypes="+ ClientLibraryCategoriesDataSourceServlet.RESOURCE_TYPE,
+                "sling.servlet.methods=GET",
+                "sling.servlet.extensions=html"
+        }
+)
+public class ClientLibraryCategoriesDataSourceServlet extends SlingSafeMethodsServlet {
+
+    public final static String RESOURCE_TYPE = "core/wcm/components/page/v2/datasource/clientlibrarycategories";
```
richardhand: @vladbailescu - do you think this resource type makes sense? perhaps we should be a bit more general with the naming of our datasource servlets rather than tying them to a particular component + version as we currently do.

content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/headlibs.html
vladbailescu: Just ignore, the expression options must be passed by name anyway.

content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/_cq_design_dialog/.content.xml
```diff
@@ -35,36 +35,64 @@
                         jcr:title="Properties"
                         sling:resourceType="granite/ui/components/coral/foundation/container"
                         margin="{Boolean}true">
-                        <items jcr:primaryType="nt:unstructured">
-                            <heading
+                        <items
+                            jcr:primaryType="nt:unstructured"
+                            sling:hideChildren="[heading,well]">
```
richardhand: @vladbailescu - note, I had to hide the heading and well nodes that contained the multifield and that we inherit from the basic page component.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/ClientLibraryCategoriesDataSourceServlet.java
richardhand: It could even be a control that makes it into the product as a commons data source / autocomplete.

vladbailescu: It does make sense to have versioning IMO. I don't think there's anything wrong with a future version referencing an older one either.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
richardhand: Fixed in 456d33fb04a9d21

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/ClientLibraryCategoriesDataSourceServlet.java
```diff
@@ -0,0 +1,126 @@
...
+                }
+            }
+        }
+        SimpleDataSource clientLibraryCategoriesDataSource = new SimpleDataSource(getCategoryResourceList(request, libraryType).iterator());
+        request.setAttribute(DataSource.class.getName(), clientLibraryCategoriesDataSource);
+    }
+
+    private List<Resource> getCategoryResourceList(@Nonnull SlingHttpServletRequest request, LibraryType libraryType) {
+        List<Resource> categoryResourceList = new ArrayList<>();
+        HashSet<String> clientLibraryCategoryHashSet = new HashSet<String>();
```
jckautzmann: I would call it `clientLibraryCategory` instead.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/ClientLibraryCategoriesDataSourceServlet.java
richardhand: Yes, I like that approach. Would prefer `datasources` to match the naming already in the product:
`core/wcm/components/commons/datasources/clientlibrarycategories/v1`

richardhand:
@jckautzmann, @vladbailescu - I made some adjustments here since your last review to add both `getClientLibCategoriesJsBody` and `getClientLibCategoriesJsHead` API allow output of the JavaScript client libraries to the correct page location and to maintain the integrity of the original `getClientlibs` which we still use to output CSS categories.

I also created a second variant at #294 for the reasons listed there.

richardhand:
Closing here in favour of #294 which continues to allow easy reordering of client lib categories

PR #294, opened 2018-09-24

### LINKED ISSUE #216: [Page] Support loading configurable JS ClientLib in page head element

richardhand:
Multifield variant of #289, the advantage being that it's easier to determine the order of the clientlib includes than with the multi-select autocomplete + tags:

screen shot 2018-09-24 at 21 51 16

| Fixed Issues?            | Fixes #216  
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation Provided   | 👍 (code comments and or markdown)
| Any Dependency Changes?  |


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -116,6 +125,15 @@ private String getProxyPath(ClientLibrary lib) {
         return path;
     }
 
+    protected void populateClientLibCategoriesJs() {
+        if (currentStyle != null) {
+            clientLibCategoriesJsHead = currentStyle.get(PN_CLIENTLIBS_JS_HEAD, ArrayUtils.EMPTY_STRING_ARRAY);
+            LinkedHashSet<String> categories = new LinkedHashSet<>(Arrays.asList(clientLibCategories));
+            categories.removeAll(Arrays.asList(clientLibCategoriesJsHead));
+            clientLibCategoriesJsBody = categories.toArray(new String[0]);
```
jckautzmann: `clientLibCategoriesJsBody` might contain CSS clientlibs, right? So I would call it `clientLibCategoriesBody`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -76,6 +81,9 @@
     private String appResourcesPath;
     private NavigationItem redirectTarget;
 
+    protected String[] clientLibCategoriesJsBody = new String[0];
```
jckautzmann: I'd write `clientlib` in lower case -> `clientlibCategoriesJsBody`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -76,6 +81,9 @@
...
+    protected String[] clientLibCategoriesJsHead = new String[0];
```
jckautzmann: See above: `clientLibCategoriesJsHead` -> `clientlibCategoriesJsHead`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -125,6 +143,18 @@ protected void loadFavicons(String designPath) {
         throw new UnsupportedOperationException();
     }
 
+    @Override
+    @JsonIgnore
+    public String[] getClientLibCategoriesJsBody() {
```
jckautzmann: See above: `getClientLibCategoriesJsBody` -> `getClientlibCategoriesBody` (lower case `L` and no `js`)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -227,17 +227,44 @@ default String getTitle() {
...
+    /**
+     * If this page is associated with a Template, then this method will return the JavaScript client library categories
+     * which are expected to be included at the page body end, as defined by the user in the policy.
+     *
+     * @return an array of JavaScript client library categories which are expected to be included at the page body end;
+     * the array can be empty if the page doesn't have an associated template or if the template has no body client libraries defined.
+     * @since com.adobe.cq.wcm.core.components.models 12.5.0;
+     */
+    @JsonIgnore
+    default String[] getClientLibCategoriesJsBody() {
```
jckautzmann: `getClientLibCategoriesJsBody` -> `getClientlibCategoriesBody`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -227,17 +227,44 @@ default String getTitle() {
...
+    /**
+     * If this page is associated with a Template, then this method will return the JavaScript client library categories
+     * which are expected to be included up front in the page head, as defined by the user in the policy.
+     *
+     * @return an array of JavaScript client library categories which are expected to be included in the page head; the array can be empty
+     * if the page doesn't have an associated template or if the template has no head client libraries defined.
+     * @since com.adobe.cq.wcm.core.components.models 12.5.0;
+     */
+    @JsonIgnore
+    default String[] getClientLibCategoriesJsHead() {
```
jckautzmann: `getClientLibCategoriesJsHead` -> `getClientlibCategoriesJsHead`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
richardhand: @jckautzmann - it could contain CSS clientlib categories, but they wouldn't be loaded. This function is supposed to return JS clientlibs for the page body. We need the `Js`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
richardhand: @jckautzmann - I kept the camelCase for consistency with the existing `getClientLibCategories` API function as well as the `clientLibCategories` property. I think you could argue for camelCase in `clientLibrary` -> `clientLib`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
richardhand: * Still require `js` here
* For the `L` I'm could be convinced that we shouldn't be consistent with the existing API / property naming.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
jckautzmann: ok, sure, let's be consistent with the existing API naming patterns.

PR #295, opened 2018-09-25
Issue #296 [UI Tests] Tests fail for Form Options and Teaser

msagolj:
Fixed Issues? #296 

- The granite UI multifield component is going to change behavior in next release which will be backported to 6.3. and 6.4 as well.
- The change involves the value of the name attribute of all elements inside a multifield item. It its going to be fully expanded upon dialog rendering. E.g. instead of:

```name="text"```

it will be rendered as:

```name ="./items/item0/text"```

- A number of FormOptions UI tests will fail as they use the```name``` attribute and value to target elements.
- The teaser will no longer work properly since the teaser's editor clientlib also use the ```name``` attribute and value as selectors in the code. This also causes tests to fail.
- The fix for FormOption tests is to change the selector to search for ending substring.
- The fix for the Teaser clientlib is to add ```data-``` attributes to be able to target the element independently of the underlying UI element.

As a follow up task you wo should decide on a common pattern to identify UI elements for testing and clientlibs that is independent of the underlying implementation by using either ```data-``` attributes or classes.

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/clientlibs/editor/js/teaser.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/clientlibs/editor/js/teaser.js
@@ -88,7 +88,7 @@
                     var newMultifieldItem = new Coral.Multifield.Item();
                     actionsMultifield.items.add(newMultifieldItem);
                     Coral.commons.ready(newMultifieldItem, function(element) {
-                        var linkField = $(element).find('foundation-autocomplete[name="link"]');
+                        var linkField = $(element).find("[data-cmp-teaser-v1-dialog-edit-hook='actionLink']");
                         if (linkField) {
                             linkField.val(linkURL);
                             linkField.trigger("change");
@@ -107,8 +107,8 @@
 
     function toggleActionItems(actionsMultifield, disabled) {
         actionsMultifield.find("coral-multifield-item").each(function(ix, item) {
-            var linkField = $(item).find("foundation-autocomplete[name='link']").adaptTo("foundation-field");
-            var textField = $(item).find("input[name='text']").adaptTo("foundation-field");
+            var linkField = $(item).find("[data-cmp-teaser-v1-dialog-edit-hook='actionLink']").adaptTo("foundation-field");
+            var textField = $(item).find("[data-cmp-teaser-v1-dialog-edit-hook='actionTitle']").adaptTo("foundation-field");
             if (disabled && linkField.getValue() === "" && textField.getValue() === "") {
                 actionsMultifield[0].items.remove(item);
             }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: testing/it/ui-js/src/content/jcr_root/apps/core/wcm/tests/test-cases/FormOptions/v1/FormOptions.js [hunk 1]
```diff
@@ -99,9 +99,9 @@ window.CQ.CoreComponentsIT.FormOptions.v1 = window.CQ.CoreComponentsIT.FormOptio
         // press the Add button
         .click("button :contains('Add')")
         // set the value
-        .fillInput("input[name='./value']", value)
+        .fillInput("input[name$='value']", value)
```
richardhand: I still find these selectors flimsy. If we came along with a new `<input type="text" value="./optionValue"/>` somewhere in the dialog, it would pass here.

content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/clientlibs/editor/js/teaser.js
```diff
@@ -107,8 +107,8 @@
 
     function toggleActionItems(actionsMultifield, disabled) {
         actionsMultifield.find("coral-multifield-item").each(function(ix, item) {
-            var linkField = $(item).find("foundation-autocomplete[name='link']").adaptTo("foundation-field");
-            var textField = $(item).find("input[name='text']").adaptTo("foundation-field");
+            var linkField = $(item).find("[data-cmp-teaser-v1-dialog-edit-hook='actionLink']").adaptTo("foundation-field");
```
richardhand: I understand it's out of the scope of the issue, but would be great if we could apply the mechanism to the whole dialog, or at least have a follow up issue to track the effort for all components.

testing/it/ui-js/src/content/jcr_root/apps/core/wcm/tests/test-cases/FormOptions/v1/FormOptions.js
msagolj: In the case of the FormOptions tests I went for the quick fix, to make the tests pass again. But I agree that we should define how in the future we are going to make the UI elements of dialogs 
selectable with either classes or data attributes.

content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/clientlibs/editor/js/teaser.js
msagolj: see comment above.

testing/it/ui-js/src/content/jcr_root/apps/core/wcm/tests/test-cases/FormOptions/v1/FormOptions.js
richardhand: ok, could you make sure we have this effort tracked?

msagolj:
> @msagolj - could you set a title for this PR and a bit of a description of the fix?

added title and description.

testing/it/ui-js/src/content/jcr_root/apps/core/wcm/tests/test-cases/FormOptions/v1/FormOptions.js
msagolj: created #297 as follow up issue

PR #301, opened 2018-09-26

richardhand:
- further clarity in the description of the head clientlib JS dialog field


msagolj:
+1

PR #306, opened 2018-10-03
[Carousel][Image] Images don't get the proper size/width when they are selected into the view

### LINKED ISSUE #305: [Carousel][Image] Images don't get the proper size/width when they are selected into the view
Steps to reproduce:

Add a carousel to the page
Add two image components and configure the images
Switch the active slide
Expected result: the image that appears has the proper resolution/width

Actual results: the image that appears has the lowest resolution/width

vladbailescu:
* Getting the width from the first visible parent in case the image is hidden
* Fixes #305


msagolj:
Verified Edit/Preview mode on chrome, firefox, edge

PR #331, opened 2018-10-12
[Tabs][Carousel] Update documentation 

richardhand:
- adds AEM 6.3, AEM 6.4 columns to system requirements
- updates Tabs, Carousel README.md
- fixes Tabs dialog label for new “item” naming

New system requirements table is as follows:
screen shot 2018-10-12 at 20 14 54


joerghoh:
What about version 2.1.0 and AEM 6.4.2? Is this a supported case?

richardhand:
@joerghoh yes that is supported.

But your question highlight that there's still some problems with interpreting this table. The AEM requirements should be read as minimum requirements, which is confusing as the Java ones aren't.

joerghoh:
Well, it's not clear on first sight (for people not familiar with the  AEM release policy), that AEM 6.4.1 + FP 24267 < AEM 6.4.2.

richardhand:
I agree, we could list both items as we do for the `1.1.0` release: 6.4.1.0 + FP24267 or 6.4.2.0

kwin:
@richardhand What is the reason the system requirements have been adjusted even for already released versions?
E.g. for 2.1.0 (released in July) it used to be `AEM 6.4.1.0 + FP24267`, after this merge it is `AEM 6.4.2.0`. I think you should really reference the minimum and not adjust the requirements for already released versions afterwards. Why has this been increased to 6.4.2.0?

gabrielwalt:
Hi @kwin, sorry for the late reply.
The thing is that the FP are kind of interim solutions until a SP consolidates them. Adobe's support really recommends to move to the SP when they are available instead of staying with the FP.

kwin:
Thanks @gabrielwalt. Although I do understand the recommendation from Adobe to always upgrade to the last SP this is not always possible. From the WCM Core you should really list the minimum requirement. One concrete question: Does 2.2.0 run on top of 6.4.1.0 + FP24267 or is there any specific API from SP2 you leverage which is not available in 6.4.1.0 + FP24267.

Issue #340, opened 2018-10-22
[ContentFragment] Provide a list component for content fragments


philipmundt:
## Feature Request

Enable Core List Component to support Content Fragments, or create a separate Content Fragment List Component - to support the use case of gathering multiple content fragments in a component, by folder, or by tag, and delivered against single URL, HTML or JSON.

> As an author, I want to define a list of Content Fragments, so that they can be consumed by my App.

> As a front-end developer, I want to consume the authored list of Content Fragments as JSON, so that I can display them in my App.

The resulting output should be consistent with the already existing single content fragment component, the JSON that is output should be compatible. Ideally, code and API should be re-used.

olimination:
I'm not sure, if I have understood the Asset Collection idea correctly, but would it be an option to use the "Asset Collection" feature? So we could simply point to some Asset Collection and this would load then the associated  content fragment assets?

gabrielwalt:
Great idea @olimination, to pointing to an Asset Collection would be a very meaningful way to define and manage a list, I like it! For now, I don't think that this has been implemented yet, but let's keep this in mind for future list capabilities.

jckautzmann:
The content fragment list component is now part of the Core Components:
https://github.com/adobe/aem-core-wcm-components/pull/502

PR #342, opened 2018-10-23
[ContentFragment] Add content fragment list component (Adobe-Marketing-Cloud#340)

philipmundt:
Fixes/adds #340

The three last commits were added temporarily in order to facilitate reviewing the components.


vladbailescu:
@philipmundt , as discussed offline, it would be great if you could:

- target the main package of the Core Components (as we want to get rid of the extension altogether)
- break up the PR in smaller parts (one for the Content Fragment component and a second one for the Content Fragment List component)
- provide ITs for the new components

Issue #327, opened 2018-10-10
[Carousel] Pause carousel autoplay when carousel document is hidden


richardhand:
So that the visible item in a Carousel is consistent when a user returns to a page it would be beneficial to pause the autoplay of the Carousel when the document containing the Carousel is hidden.

Could be achieved with [1].

[1] https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

(CQ-4254862)

richardhand:
This was fixed with the merge of #358 and is on master. It will be available in the upcoming [2.2.2 release](https://github.com/adobe/aem-core-wcm-components/milestone/9).

Issue #326, opened 2018-10-10
[Carousel] Provide pause/play button when autoplay is enabled


richardhand:
According to the WAI [0][1] it's _compulsory_ to provide a pause/play button when autorotating a carousel / slideshow.

A user may want to pause / stop auto rotation in order to focus on the content of a particular slide. This is currently possible by hovering the carousel, but the feature is somewhat hidden.

[0] https://www.w3.org/WAI/tutorials/carousels/
[1] https://www.w3.org/WAI/EO/Drafts/tutorials/sliders/

(CQ-4254859)

richardhand:
This was fixed with the merge of #358 and is on master. It will be available in the upcoming [2.2.2 release](https://github.com/adobe/aem-core-wcm-components/milestone/9).

PR #358, opened 2018-10-30
#326 #327 #361 [Carousel] Pause/play enhancements

richardhand:
This pull request addresses issues #326 #327 and #361 which are closely related to the autoplay and pause/play bug/features we would like to address for the [2.2.2 milestone](https://github.com/adobe/aem-core-wcm-components/milestone/9).

**General**
- documentation updates
- test case addition and updates

#326 Provide pause/play button when autoplay is enabled
- adds pause and play button markup
- adds JS handling for playing / pausing
- simplifies default site CSS
- adds ability to configure auto-pause on hover via the edit and policy dialogs. It may be a valid use-case for example to disable auto-pause on hover for a full screen Carousel.

#327 Pause carousel autoplay when carousel document is hidden
- uses the page visibility API (https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to only continue autoplay if the document isn't hidden (background tabs, minimised window, OS lock).

#361 Disable and hide autoplay related fields if autoplay isn't enabled
- adds a well in dialog content to group the autoplay fields (delay and auto-pause disabled).
- hides the well if autoplay isn't enabled.
- disables "delay" and "auto-pause disabled" fields if autoplay isn't enabled.

autoplay-group-toggle

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
@@ -40,6 +40,13 @@ public interface Carousel extends Container {
     String PN_DELAY = "delay";
 
     /**
+     * Name of the resource property that indicates whether automatic pause on hovering the carousel is disabled, or not.
+     *
+     * @since com.adobe.cq.wcm.core.components.models 12.6.0
+     */
+    String PN_AUTOPAUSE_DISABLED = "autopauseDisabled";
+
+    /**
      * Indicates whether the carousel should automatically transition between slides or not.
      *
      * @return {@code true} if the carousel should automatically transition slides; {@code false} otherwise
@@ -56,6 +63,16 @@ public interface Carousel extends Container {
      * @since com.adobe.cq.wcm.core.components.models 12.5.0
      */
     default Long getDelay() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Indicates whether automatic pause on hovering the carousel is disabled, or not.
+     *
+     * @return {@code true} if automatic pause on hovering the carousel should be disabled; {@code false} otherwise
+     * @since com.adobe.cq.wcm.core.components.models 12.6.0
+     */
+    default boolean getAutopauseDisabled() {
         throw new UnsupportedOperationException();
     }
 
--- a/content/.stylelintrc.yaml
+++ b/content/.stylelintrc.yaml
@@ -38,7 +38,6 @@ rules:
 
     # property
     property-blacklist: null
-    property-no-vendor-prefix: true
     property-whitelist: null
 
     # declaration
--- a/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/site/css/carousel.less
+++ b/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/site/css/carousel.less
@@ -61,6 +61,8 @@
 }
 
 .cmp-carousel__action--previous {
+    left: 0;
+
     .cmp-carousel__action-icon {
         background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%23fff' viewBox='0 0 8 8'%3E%3Cpath d='M5.25 0l-4 4 4 4 1.5-1.5-2.5-2.5 2.5-2.5-1.5-1.5z'/%3E%3C/svg%3E");
     }
--- a/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/site/css/carousel.less
+++ b/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/site/css/carousel.less
@@ -15,6 +15,7 @@
  */
 
 @cmp-carousel-indicator-size: 10px;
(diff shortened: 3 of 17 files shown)


content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/site/css/carousel.less
```diff
@@ -15,7 +15,6 @@
  */
 
 @cmp-carousel-indicator-size: 10px;
-@cmp-carousel-action-icon-size: 24px;
```
richardhand: this is unused

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/site/js/carousel.js
```diff
@@ -377,11 +477,14 @@
          * @private
          */
         function resetAutoplayInterval() {
-            if (!that._properties.autoplay) {
+            if (that._paused || !that._properties.autoplay) {
                 return;
             }
             clearAutoplayInterval();
             that._autoplayIntervalId = window.setInterval(function() {
+                if (document.visibilityState && document.hidden) {
```
richardhand: See https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API. Here we prevent the interval from pinging a navigation if the page is hidden (tab is background, minimised or OS lock).

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/site/js/carousel.js
```diff
@@ -59,6 +59,19 @@
...
+         * Determines whether automatic pause on hovering the carousel is disabled
+         *
+         * @memberof Carousel
+         * @type {Boolean}
+         * @default false
+         */
+        "autopauseDisabled": {
+            "default": false,
+            "transform": function(value) {
+                return !(value === null || typeof value === "undefined");
```
vladbailescu: What if `value` is `false`?

richardhand: @vladbailescu - this is a boolean attribute [0]. The presence of the `data-autopause-disabled` attribute indicates `true`, the attribute missing indicates `false`.

[0] https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.3.4.2

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
```diff
@@ -39,6 +39,13 @@
      */
     String PN_DELAY = "delay";
 
+    /**
+     * Name of the resource property that indicates whether automatic pause on hovering the carousel is disabled, or not.
+     *
+     * @since com.adobe.cq.wcm.core.components.models 12.6.0
+     */
+    String PN_AUTOPAUSE_DISABLED = "autopauseDisabled";
```
jckautzmann: How about renaming the property `autopause` and invert its values?

richardhand: Because we generally want the defaults to be `false`

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/site/js/carousel.js
```diff
@@ -274,6 +316,73 @@
...
+
+            // If the Carousel is hovered, don't begin auto transitioning until the next mouse leave event
+            var hovered = that._elements.self.parentElement.querySelector(":hover") === that._elements.self;
+            if (that._properties.autopauseDisabled || !hovered) {
+                resetAutoplayInterval();
+            }
+
+            var pause = that._elements["pause"];
+            if (pause) {
+                setActionDisabled(pause, false);
```
jckautzmann: I find this a bit confusing to read. When I play the carousel, I would expect the `pause` action to be disabled, which would translate into: `setActionDisabled(pause, true);`.

jckautzmann: oh I see.. When the carousel is playing, the `pause` action is enabled so that it can be clicked on.
I would add a comment to explain it.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
jckautzmann: Thx for refreshing my memory! Yes I agree that defaults should be unchecked, even if it makes property names sometimes a bit heavy.

richardhand:
Thanks @jckautzmann for the review! Made a (hopefully) last update in 2fe199749c6d6ec9a4ba333525fac14928cbf893 to:

- simplify the default site CSS
- simplify the pause/play logic based on your comments. Hopefully the JSDoc for `setActionDisabled` is enough to know what's going on, and I grouped the logic under `refreshPausePlayActions`.

content/.stylelintrc.yaml
```diff
@@ -38,7 +38,6 @@ rules:
 
     # property
     property-blacklist: null
-    property-no-vendor-prefix: true
```
richardhand: this is only useful if you have some autoprefix logic in the build

PR #374, opened 2018-11-13
#372 [Carousel] Saving dialog with autoplay unchecked, removes delay and disable on pause defaults

### LINKED ISSUE #372: [Carousel] Saving dialog with autoplay unchecked, removes delay and disable on pause defaults
Saving a Carousel dialog (edit or policy) with "Automatically transition slides" unchecked (and without allowing autoplay first and saving), leads to the default values for "Transition Delay" and "Disable automatic pause on hover" missing for any subsequent load of the same dialog.

screen shot 2018-11-08 at 17 01 20

CQ-4257133

richardhand:
Due to the default behaviour of the Granite `com.adobe.granite.ui.components.Value.val()` [0] method if the form isn't fresh, default values are ignored if a property isn't set. As we disable the autoplay fields (delay + disable autopause), they don't have a value set the first time and for subsequent dialog load they therefore aren't provided with a default.

Here we don't disable these so that their default is persisted and available on subsequent load.

[0] - https://helpx.adobe.com/experience-manager/6-4/sites/developing/using/reference-materials/javadoc/com/adobe/granite/ui/components/Value.html

| Fixed Issues?            | `Fixes #372` 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Tests Added + Pass?      |
| Documentation Provided   | (code comments and or markdown)
| Any Dependency Changes?  |

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/editor/js/carousel.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/editor/js/carousel.js
@@ -20,15 +20,11 @@
     var selectors = {
         dialogContent: ".cmp-carousel__editor",
         autoplay: "[data-cmp-carousel-v1-dialog-hook='autoplay']",
-        autoplayGroup: "[data-cmp-carousel-v1-dialog-hook='autoplayGroup']",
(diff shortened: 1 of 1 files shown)


richardhand:
This was fixed with the merge of #374 and is available on master. Will be included in the [2.2.2 release](https://github.com/adobe/aem-core-wcm-components/milestone/9)

PR #379, opened 2018-11-14
#376 [Tabs][Carousel] Panel title should not override child item title

### LINKED ISSUE #376: [Tabs][Carousel] Panel title should not override child item title
For both the Tabs and Carousel components we currently override the `jcr:title` property of the child components when setting an item/panel title. It's therefore not possible to have exclusive titles for the item and for the component if the title is read from the same location. As well as being inflexible, this could be confusing for authors as it's easy to accidentally override the component title via the dialog.

Proposal is to instead write a unique namespaced property `cmp:panelTitle` to the component nodes. The items should read this property and if it doesn't exist fall back to the `jcr:title`.

CQ-4257499

### LINKED ISSUE #378: [Tabs] Redefine active item logic
Change the way that an active item is defined such that it's not based on the path of the child item, the main motivation being a clean JSON output that doesn't expose such paths (see #298). 

Keep the `activeItem` property on the Tabs parent and rather than providing a path, return the resource name (item_2 etc.) of the active item in the Tabs Sling Model.

richardhand:
This pull request addresses issues #376 and #378.

#376 [Tabs][Carousel] Panel title should not override child item title
---------
• Updates children editor and container model to read an item `cq:panelTitle` property and if not present inherit from the `jcr:title`.
• `ResourceListItem` -> `ContainerItem` for dedicated handling of container items. Updates `Container#getItems()` for new return type.
• Major version increase required -> `12.3.0` (tbd) and bumps other references for upcoming release.
• Updates test cases.
• Updates docs.

#378 [Tabs] Redefine active item logic
---------
• Adjusts the `Tabs#getActiveItem()` to return the child item name rather than path.
• Updates rendering script as necessary.
• Added benefit that the `./activeItem` property and `Tabs#getActiveItem()` now match.

| Fixed Issues?            | `Fixes #376, Fixes #378 ` 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/TabsImpl.java
```diff
@@ -38,16 +38,16 @@
     @ValueMapValue(optional = true)
     private String activeItem;
 
-    private String activeItemPath;
+    private String activeName;
```
jckautzmann: I'd call it `activeItem`.

richardhand: we already have `activeItem` read from the valueMap. The separate property is the result of checking whether such an item exists as a child resource and if not, returning `null`. We could also just return the `activeItem` as it is with no such check.

Maybe `activeItemName` is better?

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/README.md
```diff
@@ -33,7 +33,7 @@ The component policy dialog allows definition of allowed components for the Tabs
 ### Edit Dialog Properties
 The following properties are written to JCR for this Tabs component and are expected to be available as `Resource` properties:
 
-1. `./activeItem` - defines the item that is active by default.
+1. `./activeItem` - defines the item that is active by default (the name of the active item `Resource`).
```
jckautzmann: I would phrase it as follows: 
defines the name of the item that is active by default

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/TabsImpl.java
jckautzmann: +1

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/package-info.java
```diff
@@ -34,7 +34,7 @@
  *      version, is bound to this proxy component resource type.
  * </p>
  */
-@Version("12.6.0")
+@Version("13.0.0")
```
vladbailescu: -1, this will break all implementations

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/PanelContainerItem.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.osgi.annotation.versioning.ConsumerType;
+
+/**
+ * Interface for a generic panel container item, used by the {@link Tabs} and {@link Carousel} models.
+ *
+ * @since com.adobe.cq.wcm.core.components.models 12.6.0
+ */
+@ConsumerType
+public interface PanelContainerItem extends ListItem {
```
vladbailescu: This interface only holds a property that could be made available from `com.adobe.cq.wcm.core.components.commons.editor.dialog.childreneditor.Item`

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/AbstractContainerImplTest.java
```diff
@@ -0,0 +1,125 @@
...
+    public void testContainerWithItems() {
+        Container container = getContainerUnderTest(CAROUSEL_1);
+        Object[][] expectedItems = {
+            {"Teaser 1 description", "item_1", "/content/container/jcr:content/root/responsivegrid/container-1/item_1", "Teaser 1"},
+            {"Teaser 2 description", "item_2", "/content/container/jcr:content/root/responsivegrid/container-1/item_2", "Teaser 2"},
+        };
+        verifyContainerItems(expectedItems, container.getItems());
+    }
+
+    private Container getContainerUnderTest(String resourcePath) {
```
jckautzmann: The code might be simplified with the new version of wcm-io-testing. See e.g.:
https://github.com/adobe/aem-core-wcm-components/blob/issue/298/bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/TabsImplTest.java#L81

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/PanelContainerItem.java
richardhand: @vladbailescu - moved it to the implementation https://github.com/adobe/aem-core-wcm-components/pull/379/commits/1360c401d5f194c6fb5643aa8c00bc8693bdd94f

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageListItemImpl.java
```diff
@@ -83,6 +84,9 @@ public String getPath() {
         return page.getPath();
     }
 
+    @Override
+    @JsonIgnore
+    public String getName() { return page.getName(); }
```
vladbailescu: Could you expand this to multiple lines to follow the same style as the rest of the code?

Issue #298, opened 2018-09-26
[Tabs][Carousel] The Json export doesn't work properly for nested Tabs and Carousels


andreeadracea:
The items elements are not exported properly in Json when having nested tabs.
(CQ-4253492)

jckautzmann:
We should first define how the JSON should look like. I see the following options:

**Option 1**

We introduce a `tabs` object with the tab item specific properties (for now only the tab title).
The `:items` object consists of all the nested models JSON.

The JSON would look as follows:
```
{
    "tabs": [
        {
            "title": "Tab 1"
        },
        {
            "title": "Tab 2"
        }
    ],
    ":itemsOrder": [
        "item_1",
        "item_2"
    ],
    ":type": "weretail/components/content/tabs",
    ":items": {
        "item_1": {
            "columnCount": 12,
            "columnClassNames": {},
            "gridClassNames": "aem-Grid aem-Grid--12 aem-Grid--default--12",
            ":items": {},
            ":itemsOrder": [],
            ":type": "wcm/foundation/components/responsivegrid"
        },
        "item_2": {
            "columnCount": 12,
            "columnClassNames": {},
            "gridClassNames": "aem-Grid aem-Grid--12 aem-Grid--default--12",
            ":items": {},
            ":itemsOrder": [],
            ":type": "wcm/foundation/components/responsivegrid"
        }
    }
}
```
Advantages:
- minimal implementation effort
- tab item specific properties are nicely isolated from the other nested models JSON.

Disadvantages:
- it might be confusing to read: which title belongs to which item?

**Option 2**

We introduce a specific node for the tab item which:
- stores the tab item specific properties (today only the title).
- has a specific resource type (e.g. `core/wcm/components/tabs/v1/tabs/item`) and a dedicated model 

The JCR structure would look as follows:
```
tabs
    item1
        responsivegrid
            text
            image
…
```

The JSON would look as follows:
```
{
    ":itemsOrder": [
        "item_1",
        "item_2"
    ],
    ":type": "weretail/components/content/tabs",
    ":items": {
        "item_1": {
            ":type": "weretail/components/content/tabs/item",
            "title": "Tab 1",
            "container": {
                "columnCount": 12,
                "columnClassNames": {},
                "gridClassNames": "aem-Grid aem-Grid--12 aem-Grid--default--12",
                ":items": {},
                ":itemsOrder": [],
                ":type": "wcm/foundation/components/responsivegrid"
            }
        },
[...]

PR #386, opened 2018-11-20
[Tabs][Carousel] The Json export doesn't work properly for nested Tab…

vladbailescu:
…s and Carousels #298

* Updated Tabs and Carousel to implement ContainerExporter


vladbailescu:
Actually, this is a bad approach as we have other places (responsive grid for example) where children of containers are fetched as ComponentExporter and the Sling ModelFactory will not match a model that adapts to ContainerExporter (even though it extends ComponentExporter)

PR #414, opened 2018-12-13
[ContentFragment] fixing in between content / adding new unit tests

### LINKED ISSUE #406: [Content Fragment] GetParagraphs doesn't handle element attributes
If you apply styles to your paragraph, it won't be detected anymore as paragraph, thus the inbetween content will not be displayed as expected (in between)

It is simply due to the current implementation that only split on exact match "p", "h1" tags etc...

Instead it should allow attributes, so maybe parsing the html would be cleaner.

tyge68:
| Fixed Issues?            | `Fixes #406` 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  | Yes added org.apache.sling.commons.html (HTMLParser)

Changed the way it parse the html to extract the paragraphs in the same way it is done in the implementationn of the core contentfragment (since that part is not open)


extension/contentfragment/bundle/src/main/java/com/adobe/cq/wcm/core/components/extension/contentfragment/internal/models/v1/ContentFragmentImpl.java
```diff
@@ -486,8 +485,45 @@ public String getHtml() {
...
+                        // filter whitespaces or carriage returns like characters found
+                        // in between the html elements
+                        if (!StringUtils.containsOnly(text,
+                                ' ',
+                                '\u00A0' /* nbsp */,
+                                '\t',
+                                '\n',
+                                '\r'
+                        )) {
+                            paragraphs.add(text);
```
stefangrimm: This should never happen, as there should be no `TextNode`s with actual content on this level - RTE should ensure that they are nested into container elements like `p`, `h?`, `li`, etc.

As RTE is not infallible, we can probably still keep it.

extension/contentfragment/bundle/src/test/java/com/adobe/cq/wcm/core/components/extension/contentfragment/internal/models/v1/ContentFragmentElementImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+import static org.mockito.Mockito.when;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertNotNull;
+
+public class ContentFragmentElementImplTest {
+
+    @Test
+    public void testElementGetParagraphsNotText() {
+        ContentFragment.Element element = getMockedElement( null, "application/pdf", false);
```
stefangrimm: application/pdf doesn't really make sense ...

extension/contentfragment/bundle/src/test/java/com/adobe/cq/wcm/core/components/extension/contentfragment/internal/models/v1/ContentFragmentElementImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+    public void testStyledHtmlParagraphs() throws IOException {
+        ContentFragment.Element element = getMockedElement( getContent("sample_html_snippet.dat"),"text/html", false);
+        String[] paragraphs = element.getParagraphs();
+        assertNotNull(paragraphs);
+        assertEquals(10, paragraphs.length);
+    }
+
+    @Test
+    public void testPlainParagraphs() throws IOException {
+        ContentFragment.Element element = getMockedElement( getContent("sample_plain.dat"),"text/plain", false);
```
stefangrimm: This test doesn't make sense IMHO because of the way things are mocked.

In reality, RenderService would actually convert the plain text into HTML before parsing it. So the actual result should be 5.

As tests can also be considered documentation, I would remove it, as it provides a wrong impression of how things are supposed to work.

extension/contentfragment/bundle/pom.xml
```diff
@@ -304,6 +306,12 @@
             <artifactId>uber-jar</artifactId>
             <classifier>apis</classifier>
         </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
tyge68: using the Sling Commons HTML parser make sense when we can access the OSGi context or if that is possible to reference it, I will check if that is the case here, if not I guess we have no other way but to use some other libraries.

extension/contentfragment/bundle/src/main/java/com/adobe/cq/wcm/core/components/extension/contentfragment/internal/models/v1/ContentFragmentImpl.java
tyge68: It's mainly to be aligned with the core implementation where it also have this check in multiple places...

extension/contentfragment/bundle/src/test/java/com/adobe/cq/wcm/core/components/extension/contentfragment/internal/models/v1/ContentFragmentElementImplTest.java
tyge68: yes as it is the case when it's not a text, it could be anything not starting with "text/" basically.

extension/contentfragment/bundle/src/test/java/com/adobe/cq/wcm/core/components/extension/contentfragment/internal/models/v1/ContentFragmentElementImplTest.java
tyge68: ok

tyge68:
Major change in the last commit to use Sling Commons HTML Parser , which required  then to implement the Paragraphs Extractor logic, it also highlighted that "text/plain" cannot be tested properly as we don't have the adapter logic in place which is supposed to render it as html when converted to paragraphs. Thus I had to adapt the related tests to use text/html everywhere.

extension/contentfragment/bundle/pom.xml
```diff
@@ -54,6 +54,8 @@
                 </executions>
                 <configuration>
                     <instructions>
+                        <Embed-Dependency>jsoup</Embed-Dependency>
```
vladbailescu: Can we remove this then?

vladbailescu:
@stefangrimm , everything allright now? Can we merge this?

stefangrimm:
@vladbailescu Sorry, we still need to keep this PR open as it depends on other stuff (in AEM) that has not been finished yet.Scheduled for next sprint AFAIK.

vladbailescu:
Closing this PR for now as the contentfragment component has moved from the extension. Please open a new PR when ready!

Issue #274, opened 2018-08-27
Image urls should contain the original filename


kwin:
## Feature Request

**Is your feature request related to a problem? Please describe.**
Currently the image url contains the path to the image component on the page (compare with https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/wiki/The-Adaptive-Image-Servlet#url-formats) but does not contain the actual filename of the referenced asset or an artificial filename set through the editor.

**Describe the solution you'd like**
For SEO purposes a meaningful filename should be contained in the URL. Preferably between the timestamp and the extension. That filename should be derived from the DAM assets node name (or  maybe some metadata property which contains the original filename) or from a locally overwritten property (maintained in the image dialog). Also the filename should be send with an according `Content-Disposition` header (https://tools.ietf.org/html/rfc6266#section-4.3) with a filename set to that value.

Proposed URL format should look like
```<image-component-path>.<selectors>.<extension>/<timestamp>/<filename>.<extension>```

ompandeyinfy:
This certainly needs an enhancement on `AdaptiveImageServlet` because it validates `requestLastModifiedSuffix  `  [1], if it's not a valid timestamp and not  `lastModifiedEpoch`, sets a response with `302`

[1] https://github.com/adobe/aem-core-wcm-components/blob/6ef2ae39a4e2f42a4fa7466ad101643e4979fb7a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L189]

richardhand:
This was added with the merge of #419 and is on master. It will be available in the upcoming [2.3.0](https://github.com/adobe/aem-core-wcm-components/milestone/10)
 release.

Issue #48, opened 2017-07-18
Image: Include a last modification date even in publish image urls


kwin:
To allow to set a very long TTL for image urls it would be beneficial if at least optionally the publish URL (https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/tree/master/content/src/content/jcr_root/apps/core/wcm/components/image/v1/image#url-formats) includes a last modification date of the component's node as well (similar to what we have on the author already). That way it is assured that the URL changes whenever the image changes (which will automatically invalidate the cache entry)

gabrielwalt:
+1, we added that to our roadmap.

kwin:
@bpauli Can you adjust the section in the readme as well (https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/image/v1/image/README.md#url-formats)?

bpauli:
@kwin Sure, thanks for the hint!

PR #419, opened 2018-12-18
Feature/adaptive image seo friendly filename

ompandeyy:
`Fixes Issue-274`

For SEO purposes a meaningful filename should be contained in the image URL. This is an enhancement on the  `AdaptiveImageServlet` 

Image URL called with format `<image-component-path>.<selectors>.<extension>/<timestamp>/<filename>.<extension>`  is served


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -185,8 +185,10 @@ protected void doGet(@Nonnull SlingHttpServletRequest request, @Nonnull SlingHtt
                 lastModifiedEpoch = assetLastModifiedEpoch;
             }
         }
+        
         long requestLastModifiedSuffix = getRequestLastModifiedSuffix(suffix);
-        if (requestLastModifiedSuffix >= 0 && requestLastModifiedSuffix != lastModifiedEpoch) {
+        if (requestLastModifiedSuffix >= 0 && requestLastModifiedSuffix != lastModifiedEpoch
```
kwin: the redirect is always necessary IMHO, the only thing that changes is from where to extract the lastModified timestamp. This should be before the first dot of the suffix! That should be compliant with both the old and the new URL format.

jckautzmann:
@ompandeyy Thx for your contribution! Could you please elaborate how your changes address the requirements described in https://github.com/adobe/aem-core-wcm-components/issues/274 ? E.g. I don't see that the URL has now a meaningful filename derived from DAM. Or are the changes meant to only support a part of the requirements? Thx for clarifying.

ompandeyy:
> @ompandeyy Thx for your contribution! Could you please elaborate how your changes address the requirements described in #274 ? E.g. I don't see that the URL has now a meaningful filename derived from DAM. Or are the changes meant to only support a part of the requirements? Thx for clarifying.

@jckautzmann sorry I have to admit that I didn't understand the requirement completely at first and raised a PR with just the `Servlet` changes. I am right now in the process of enhancing `Image Atom` and the BE part of it. Would update the PR once it's done.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -198,6 +202,7 @@ protected void initModel() {
                 baseResourcePath = resource.getPath();
             }
             baseResourcePath = resource.getResourceResolver().map(request, baseResourcePath);
+            final String assetName = getAssetName();
```
kwin: you need to check getImageName() as that is supposed to overwrite the asset name

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
```diff
@@ -116,6 +119,34 @@ protected void initModel() {
         }
     }
 
+    private String getAssetName() {
```
kwin: v2 extends v1, please call the method from v1 instead of duplicating it.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
```diff
@@ -116,6 +119,34 @@ protected void initModel() {
...
+  
+  /**
+   * Content editors can store DAM assets with whitespaces in the name, this method makes
+   * the asset name SEO friendly and also makes it usable by the {@code AdaptiveImageServlet}
+   * 
+   * @param assetName
+   * @return name of the asset without extension
+   */
+  private String getSeoFriendlyFileName(String assetName) {
+      //check if the image name is overridden at the dialog level
```
kwin: please don't duplicate

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
ompandeyy: The timestamp doesn't exist before the first dot of the suffix in case of a component not as Image resource type. Hence I have used a regex to find it in the suffix string. This regex is checking for a 13 digit numeric string which would be the case of a valid UTC timestamp string converted in milliseconds. 
Redirection logic is now completely untouched!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
ompandeyy: fixed this.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
ompandeyy: fixed this

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
ompandeyy: Sorry, I didn't understand your comment. This method is internally calling `getSeoFriendlyFileName()` which does the overwrite work. I will adjust the comments. Please correct me if there is something else you meant.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -290,4 +296,13 @@ default String getExportedType() {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Returns the image name of the current image, if one exists.
+     *
+     * @return the image name of the current image, if one exists, {@code null} otherwise
+     */
+    default String getImageName() {
```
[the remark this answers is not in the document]
ompandeyy: removed this.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -145,6 +145,12 @@
      * @since com.adobe.cq.wcm.core.components.models 12.4.0
      */
     String PN_MAP = "imageMap";
+    
+    /**
+     * Name of the resource property that defines current image name. This value overrides the image name coming from DAM
```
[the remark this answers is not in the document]
ompandeyy: fixed

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -145,6 +145,12 @@
...
+     *
+     */
+    String IMAGE_NAME = "imageName";
```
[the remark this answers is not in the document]
ompandeyy: fixed

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -238,6 +244,36 @@ protected void initModel() {
...
+     * Content editors can store DAM assets with whitespaces in the name, this method checks if the
+     * image name is overwritten at the image dialog level or extracts the name from DAM. Also makes
+     * the asset name SEO friendly and also makes it usable by the {@code AdaptiveImageServlet}
+     * 
+     * @param assetName
+     * @return name of the asset without extension
+     */
+    private String getSeoFriendlyFileName(String assetName) {
+        //check if the image name is overridden at the dialog level
+    	if(StringUtils.isBlank(imageName)) {
```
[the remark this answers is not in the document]
ompandeyy: not needed anymore as this method is refactored.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -238,6 +244,36 @@ protected void initModel() {
...
+    /**
+     * Content editors can store DAM assets with whitespaces in the name, this method checks if the
+     * image name is overwritten at the image dialog level or extracts the name from DAM. Also makes
+     * the asset name SEO friendly and also makes it usable by the {@code AdaptiveImageServlet}
+     * 
+     * @param assetName
+     * @return name of the asset without extension
+     */
+    private String getSeoFriendlyFileName(String assetName) {
+        //check if the image name is overridden at the dialog level
```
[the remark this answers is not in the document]
ompandeyy: not needed anymore as the method is refactored

ompandeyy:
@jckautzmann I have pushed all the remaining changes. PR is ready for review.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -238,6 +244,32 @@ protected void initModel() {
         }
     }
 
+    protected String getAssetName() {
+        if (StringUtils.isNotBlank(fileReference)) {
+            Resource damResource = request.getResourceResolver().getResource(fileReference);
+            if (damResource != null) {
+              Asset asset = damResource.adaptTo(Asset.class);
+              //check if the image name is overwritten at the dialog level, else set the DAM asset name
+              final String assetName = StringUtils.isNotBlank(imageName) ? imageName : (asset != null ? asset.getName() : "");
```
kwin: Slight performance improvement: Move the check for a set image name to the beginning of the method, then you don't need to get the underlying asset at all!

ompandeyy: thanks for this, I have added this check even one level before now.

kwin:
@jckautzmann Would you mind having a 2nd look, now this should be complete.

richardhand:
@ompandeyy - thanks for the contribution! and sorry to be late in the game here, but I have some general comments. 

• I would first check the DAM asset metadata for a "Title", as set in the DAM properties dialog `./jcr:content/metadata/dc:title` before checking the name. It's likely to be more readable than the name, which can be technical and is easily edited from the DAM UI.
• I would not allow override of the title from the dialog. The only case where I see it being useful is for a locally uploaded file, where we don't auto-fetch the title. We want to move towards a future where we don't have this local file and rather DAM is the source of truth for all metadata. Perhaps you can convince me of a case for an instance override?
• This change breaks images that are added via the template editor, didn't look in any detail as to why, but I assume it's somehow related to this code section: https://github.com/adobe/aem-core-wcm-components/blob/1f5e79cdd0a2757f66cc61544e2cf5a6f6dc9eb2/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L133

cc @kwin

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -238,6 +245,33 @@ protected void initModel() {
...
+    
+    /**
+     * Content editors can store DAM assets with white spaces in the name, this method makes
+     * the asset name SEO friendly and also makes it usable by the {@code AdaptiveImageServlet}
+     * 
+     * @param assetName
+     * @return name of the asset without extension
+     */
+    protected String getSeoFriendlyName(String assetName) {
+        assetName = StringUtils.isNotBlank(assetName)? assetName.trim().replaceAll(" ", "-").toLowerCase() : null;
```
richardhand: • Do we also need to strip non-supported characters here?
• I would remove the extension first before the string manipulation

ompandeyy:
@kwin @richardhand  documentation can be found at 
https://github.com/ompandeyy/aem-core-wcm-components/wiki/The-Adaptive-Image-Servlet

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -145,7 +145,7 @@
      * @since com.adobe.cq.wcm.core.components.models 12.4.0
      */
     String PN_MAP = "imageMap";
-
+
```
kwin: Please remove this change.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServletTest.java
```diff
@@ -543,7 +543,6 @@ public void testImageFromTemplateStructureNode() throws IOException {
         Dimension actualDimension = new Dimension(image.getWidth(), image.getHeight());
         assertEquals("Expected image rendered at requested size.", expectedDimension, actualDimension);
         assertEquals("Expected a PNG image.", "image/png", response.getContentType());
-
```
kwin: Please remove this change

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/_cq_dialog/.content.xml
```diff
@@ -151,7 +151,6 @@
                                                 name="./linkURL"
                                                 rootPath="/content"
                                                 wrapperClass="cmp-image__editor-link"/>
-                                        </items>
```
kwin: This needs to be kept, otherwise you end up with invalid xml

kwin:
@jckautzmann Several comments about your remarks
> • I would first check the DAM asset metadata for a "Title", as set in the DAM properties dialog `./jcr:content/metadata/dc:title` before checking the name. It's likely to be more readable than the name, which can be technical. Plus, this title is more easily found in the DAM UI, unlike the name which requires a move operation.

The problem is that dc:title is not set from the uploaded file's filename. In earlier AEM versions the original asset filename ended up in `jcr:content/cq:name` but now no longer. "cq:name" is set via `AssetManagerImpl.createRevision()`. There it is always derived from the current resource name itself. This can be triggered by creation a new version of the asset via the UI.
locally it was never set (although the dam workflow has been executed). 
Some more observations around names:

When you create the asset via "Create" it always takes node name from the given name in the upload dialog (this will not create cq:name) (initially the value is filled from the uploaded's image filename).Moving an asset around afterwards potentially changes the asset's node name.

In most of the cases dc:title might be a nice thing to expose, e.g. via the title attribute on the image it doesn't seem proper for the filename as there in most cases you are interested in reusing the original assets filename. WDYT?


> I would not allow override of the title from the dialog. The only case where I see it being useful is for a locally uploaded file, where we don't auto-fetch the title. We want to move towards a future where we don't have this local file and rather DAM is the source of truth for all metadata. Perhaps you can convince me of a case for an instance override? cc @gabrielwalt

Although local upload is not widely used, currently it is still allowed. While you have the option I would keep a local filename option, because otherwise it is impossible to influence the filename for locally uploaded images (it seems that the original filename is not transmitted via [granite/ui/components/coral/foundation/form/fileupload](https://helpx.adobe.com/experience-manager/6-4/sites/developing/using/reference-materials/granite-ui/api/jcr_root/libs/granite/ui/components/coral/foundation/form/fileupload/index.html)).

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
ompandeyy: removed.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServletTest.java
ompandeyy: removed

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/_cq_dialog/.content.xml
ompandeyy: Reverted this file as we are not adding a custom field for image name at the component level.

richardhand:
@kwin, @ompandeyy - thanks for your input. I spoke to @gabrielwalt about the requirements again here. We would like to take the following approach:

• Revert the change of using the `dc:title` and instead just take the image name, as we did originally (apologies for this @ompandeyy, I know it's annoying having to keep updating the PR).
• Still not add a field in the dialog. We would like to keep it as simple as possible in the first step and assess later whether this is really required. Again, we may be removing the local upload in the near future.

ompandeyy:
@richardhand  thanks ! 
> • Revert the change of using the `dc:title` and instead just take the image name, as we did originally (apologies for this @ompandeyy, I know it's annoying having to keep updating the PR).

Reverted this back to the getName() method.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
```diff
@@ -110,13 +111,15 @@ protected void initModel() {
             disableLazyLoading = currentStyle.get(PN_DESIGN_LAZY_LOADING_ENABLED, true);
 
             String staticSelectors = selector;
+            final String imageName = getImageNameFromDam();
```
kwin: How does this work now for locally uploaded assets now? From where to extract the name?

kwin:
@richardhand Thanks for the input but I still don't get from where to to retrieve the name in case this is a locally uploaded file, or do you recommend that in that case the filename is not included at all?

richardhand:
@kwin, we won't retrieve a name for local uploads

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
ompandeyy: > @kwin, we won't retrieve a name for local uploads

@kwin  @richardhand  Just to confirm it, in that case the image URL would be with the timestamp as it is in current version.
`<image-component-path>.<selectors>.<extension>/<timestamp>.<extension>`

richardhand: @ompandeyy yes exactly

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -202,6 +205,8 @@ protected void initModel() {
                 baseResourcePath = resource.getPath();
             }
             baseResourcePath = resource.getResourceResolver().map(request, baseResourcePath);
+            //get asset title or name from dam
```
richardhand: I think we can remove this comment altogether now that we don't look up the title. The function call below is itself readable.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -210,7 +215,7 @@ protected void initModel() {
                     smartImages[index] = baseResourcePath + DOT +
                         selector + DOT + jpegQuality + DOT + width + DOT + extension +
                         (inTemplate ? Text.escapePath(templateRelativePath) : "") +
-                        (lastModifiedDate > 0 ? "/" + lastModifiedDate + DOT + extension : "");
+                        (lastModifiedDate > 0 ? "/" + lastModifiedDate + (StringUtils.isNotBlank(imageName)? "/" +imageName :"") + DOT + extension : "");
```
richardhand: Formatting, can we space the ternary operators:
`(StringUtils.isNotBlank(imageName) ? "/" +imageName : "")`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -224,7 +229,8 @@ protected void initModel() {
             } else {
                 src += extension;
             }
-            src += (inTemplate ? Text.escapePath(templateRelativePath) : "") + (lastModifiedDate > 0 ? "/" + lastModifiedDate + DOT +
+            src += (inTemplate ? Text.escapePath(templateRelativePath) : "") + (lastModifiedDate > 0 ? "/" + lastModifiedDate +
+                (StringUtils.isNotBlank(imageName)? "/" +imageName :"") + DOT +
```
richardhand: Same comment as above re. ternary formatting

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -238,6 +244,42 @@ protected void initModel() {
...
+    /**
+     * Extracts the asset name from DAM resource
+     * 
+     * @return image name from DAM
+     */
+    protected String getImageNameFromDam() {
+        String imageName = "";
+        Resource damResource = request.getResourceResolver().getResource(fileReference);
+        if (damResource != null) {
+          Asset asset = damResource.adaptTo(Asset.class);
```
richardhand: Formatting; indentation looks off here

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -238,6 +244,42 @@ protected void initModel() {
...
+     * format using {@code utf-8} encoding scheme.
+     * 
+     * @param assetName
+     * @return name of the asset without extension
+     */
+    protected String getSeoFriendlyName(String imageName) {
+      
+      //Google recommends to use hyphens (-) instead of underscores (_) for seo. See https://support.google.com/webmasters/answer/76329?hl=en 
+      String seoFriendlyName = imageName.replaceAll("[\\ _]", "-").toLowerCase();
+        try {
```
richardhand: Formatting, indentation looks a little off here as well (we should make sure we're using spaces not tabs).

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -238,6 +244,42 @@ protected void initModel() {
...
+        }
+        return getSeoFriendlyName(FilenameUtils.getBaseName(imageName));
+    }
+    
+    /**
+     * Content editors can store DAM assets with white spaces in the name, this method makes
+     * the asset name SEO friendly, Translates the string into {@code application/x-www-form-urlencoded} 
+     * format using {@code utf-8} encoding scheme.
+     * 
+     * @param assetName
```
richardhand: param should be `imageName`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
```diff
@@ -110,13 +111,15 @@ protected void initModel() {
...
             if (smartSizes.length > 0) {
                 // only include the quality selector in the URL, if there are sizes configured
                 staticSelectors += DOT + jpegQuality;
             } 
             srcUriTemplate = baseResourcePath + DOT + staticSelectors +
                     SRC_URI_TEMPLATE_WIDTH_VAR + DOT + extension +
-                    (inTemplate ? templateRelativePath : "") + (lastModifiedDate > 0 ? "/" + lastModifiedDate + DOT + extension : "");
+                    (inTemplate ? templateRelativePath : "") + (lastModifiedDate > 0 ? "/" + lastModifiedDate 
+                        +(StringUtils.isNotBlank(imageName)? "/" +imageName :"") + DOT + extension : "");
```
richardhand: Same formatting comment as in the v1 implementation

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -423,7 +436,7 @@ private void transformAndStreamFile(SlingHttpServletResponse response, ValueMap
                     || "svg".equalsIgnoreCase(mimeTypeService.getExtension(imageType))) {
                 LOGGER.debug("GIF or SVG file detected; will render the original file.");
                 if (is != null) {
-                    stream(response, is, imageType);
+                    stream(response, is, imageType,imageName);
```
richardhand: Space missing:
`stream(response, is, imageType, imageName);`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -277,15 +280,25 @@ protected void doGet(@Nonnull SlingHttpServletRequest request, @Nonnull SlingHtt
...
     }
+    
+    /**
+     * Extracts the image name from image request url suffix. 
+     * 
+     * @param suffix
+     * @return image base name without slashes and extension
+     */
+    private String getImageName(String suffix) {
+      return StringUtils.isNotEmpty(suffix) ? FilenameUtils.getBaseName(suffix) : "";
```
richardhand: indent looks off

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -714,9 +728,15 @@ private Integer getAllowedJpegQuality(@Nonnull ResourceResolver resourceResolver
     private long getRequestLastModifiedSuffix(@Nullable String suffix) {
         long requestLastModified = 0;
         if (StringUtils.isNotEmpty(suffix) && suffix.contains(".")) {
-            String lastMod = suffix.substring(1, suffix.lastIndexOf("."));
+           
+            //check if the 13 digits UTC milliseconds timestamp is present in the suffix
+            Pattern p = Pattern.compile("\\(|\\)|\\d{13}");
+            Matcher m = p.matcher(suffix);
+            if(!m.find()) {
```
richardhand: Space missing:
`if (!m.find()) {`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -502,9 +515,10 @@ private Rendition getAWebRendition(Asset asset) {
         return null;
     }
 
-    private void stream(@Nonnull SlingHttpServletResponse response, @Nonnull InputStream inputStream, @Nonnull String contentType)
+    private void stream(@Nonnull SlingHttpServletResponse response, @Nonnull InputStream inputStream, @Nonnull String contentType, String imageName)
             throws IOException {
         response.setContentType(contentType);
+        response.setHeader( "Content-Disposition", "filename=\"" + imageName + "\"" );
```
richardhand: Shouldn't we also include the file extension here?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -224,7 +229,8 @@ protected void initModel() {
             } else {
                 src += extension;
             }
-            src += (inTemplate ? Text.escapePath(templateRelativePath) : "") + (lastModifiedDate > 0 ? "/" + lastModifiedDate + DOT +
+            src += (inTemplate ? Text.escapePath(templateRelativePath) : "") + (lastModifiedDate > 0 ? "/" + lastModifiedDate +
+                (StringUtils.isNotBlank(imageName)? "/" +imageName :"") + DOT +
                     extension : "");
```
richardhand: this can go on the previous line

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
richardhand: indent; 4 spaces

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -238,6 +244,42 @@ protected void initModel() {
...
+     * 
+     * @param assetName
+     * @return name of the asset without extension
+     */
+    protected String getSeoFriendlyName(String imageName) {
+      
+      //Google recommends to use hyphens (-) instead of underscores (_) for seo. See https://support.google.com/webmasters/answer/76329?hl=en 
+      String seoFriendlyName = imageName.replaceAll("[\\ _]", "-").toLowerCase();
+        try {
+          seoFriendlyName = URLEncoder.encode(seoFriendlyName,"utf-8");
```
richardhand: Is there a constant that we can use for the encoding here rather than the direct string?

richardhand:
@ompandeyy - I added a few final (picky) comments mainly related to formatting. Other than that, I still see a few TODOs:

• Fixing the problem with an image coming from the template:
 
Here is an example for the image URL from the page:
`http://<domain>/content/core-components-examples/library/carousel/_jcr_content/root/responsivegrid/demo_181646866/component/carousel/image.coreimg.svg/1548174499578/adobeexperiencemanager-corecomponents-logo.svg`

And this is what it looks like from a template:
`http://<domain>/content/core-components-examples/library/carousel.coreimg.svg/structure/_jcr_content/root/responsivegrid_1409995310/image_1442382866/1548160386140/adobeexperiencemanager-corecomponents-logo.svg`

We just need to make sure that the handling at [0] is updated. I think `relativeTemplatePath` is probably wrong following the change as the last slash index is now everything including the timestamp, but not the image name.

I couldn't find an example of this in the wknd tutorial, but if you edit a template for any page there and simply add an image component it's quite simple to set up this scenario.

• Ensure this is document in the component README.md for both versions of the component (https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/README.md)

[0] - https://github.com/adobe/aem-core-wcm-components/blob/master/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L140

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
ompandeyy: I have formatted with 4 spaces(Space only)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
ompandeyy: I have formatted with 4 spaces(Space only)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
ompandeyy: I have added it now. Couldn't find a reason for not including this.

ompandeyy:
thanks @richardhand , I have updated the PR. Below are my comments inline:
> @ompandeyy - I added a few final (picky ) comments mainly related to formatting. Other than that, I still see a few TODOs:
> 
> • Fixing the problem with an image coming from the template:
**Fixed**

> We just need to make sure that the handling at [0] is updated. I think `relativeTemplatePath` is probably wrong following the change as the last slash index is now incorrectly everything including the timestamp, but not the image name - the timestamp should also be discarded.
**This is handled now.**

> • Ensure the new pattern is documented in the component README.md for both versions of the component (https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/README.md)
**Both the files are updated with new URL(including `<filename>`)**
> 
> [0] - https://github.com/adobe/aem-core-wcm-components/blob/master/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L140

richardhand:
thanks @ompandeyy for the changes! I installed the update and everything look good now, just one thing: The Content-Disposition header is missing a Disposition Type as per the spec: https://tools.ietf.org/html/rfc6266#section-4.3

ompandeyy:
> thanks @ompandeyy for the changes! I installed the update and everything looks good now, just one thing: The Content-Disposition header is missing a Disposition Type as per the spec: https://tools.ietf.org/html/rfc6266#section-4.2

@richardhand  I have set this with "`attachment`" and also encoded the `filename`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
ompandeyy: `CharEncoding.UTF_8` is used now.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
ompandeyy: fixed

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
ompandeyy: fixed

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
ompandeyy: fixed

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
ompandeyy: fixed

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
ompandeyy: fixed

richardhand:
@ompandeyy - I would have expected the default `inline`

ompandeyy:
> @ompandeyy - I would have expected the default `inline`

@richardhand `"inline"` should be default even if we don't explicitly set it. Let me know if you still would want to have it.

ompandeyy:
> @ompandeyy, I think it's advisable to set it explicitly even if it "works" without, I don't see the type listed as optional in the spec, solely the parameters.

@richardhand  sure, I have set this explicitly now.

bundles/core/src/test/resources/image/v2/exporter-image4.json
```diff
@@ -1,9 +1,9 @@
 {
-    "srcUriTemplate"        : "/core/content/test/_jcr_content/root/image4.coreimg{.width}.png/1494867377756.png",
+    "srcUriTemplate"        : "/core/content/test/_jcr_content/root/image4.coreimg{.width}.png/1494867377756/adobe-systems-logo-and-wordmark.png",
     "widths": [],
     "areas": [],
     "lazyEnabled"           : false,
     "title"                 : "Adobe Systems Logo and Wordmark",
-    "src"                   : "/core/content/test/_jcr_content/root/image4.coreimg.png/1494867377756.png",
+    "src"                   : "/core/content/test/_jcr_content/root/image4.coreimg.png/1494867377756/adobe-systems-logo-and-wordmark.png",
```
jckautzmann: @ompandeyy @richardhand It seems that all the tests have been updated to contain a filename in the URL. Do we still have tests without? Both use cases should be covered by tests.

richardhand: @jckautzmann - good point - I did test both scenarios manually but an automated test would be a good idea.

richardhand: @ompandeyy will you have the bandwidth to add a case the no name scenario? Following that I would recommend we merge here.

ompandeyy: @richardhand Unfortunately not today because of some unexpected Laptop failure in my Office. I can target this by Monday though.

richardhand: Great! Thanks again for your efforts

content/src/content/jcr_root/apps/core/wcm/components/image/v1/image/README.md
```diff
@@ -95,7 +95,7 @@ The images are loaded through the `com.adobe.cq.wcm.core.components.internal.ser
 
 ```
 Author:
-/content/<project_path>/<page_path>/<component_path>/<component_name>.img.<quality>.<width>.<extension>/<timestamp>.<extension>
+/content/<project_path>/<page_path>/<component_path>/<component_name>.img.<quality>.<width>.<extension>/<timestamp>/<filename>.<extension>
 
 Publish:
 /content/<project_path>/<page_path>/<component_path>/<component_name>.img.<quality>.<width>.<extension>
```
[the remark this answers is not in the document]
ompandeyy: I have updated this with the inclusion of  `<timestamp>/<filename>.<extension`>

bundles/core/src/test/resources/image/v2/exporter-image4.json
ompandeyy: @jckautzmann  @richardhand I have added tests for image and template type cases for the Servlet. I am not sure if we really need a test case for a missing name in the `fileReference `for `ImageImpl.java(v1 and v2)`

richardhand:
@ompandeyy no problem we can add this.

richardhand:
@ompandeyy, @jckautzmann - Just pushed 94cd80350bdf8ad7. Realised when working on the issue that the file name is in fact available for local uploads via the `./fileName` property on the image resource [0]. I adjusted the handling to read this property in the local upload case if it available (which is usually should be). Also found a JUnit test case that the local upload scenario and updated it to reflect the new change.

[0] - https://helpx.adobe.com/experience-manager/6-2/sites/developing/using/reference-materials/javadoc/com/day/cq/commons/DownloadResource.html#PN_FILE_NAME

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/README.md
```diff
@@ -62,10 +62,10 @@ The images are loaded through the `com.adobe.cq.wcm.core.components.internal.ser
 
 ```
 Author:
-/content/<project_path>/<page_path>/<component_path>/<component_name>.coreimg.<quality>.<width>.<extension>/<timestamp>.<extension>
+/content/<project_path>/<page_path>/<component_path>/<component_name>.coreimg.<quality>.<width>.<extension>/<timestamp>/<filename>.<extension>
 
 Publish:
-/content/<project_path>/<page_path>/<component_path>/<component_name>.coreimg.<quality>.<width>.<extension>
+/content/<project_path>/<page_path>/<component_path>/<component_name>.coreimg.<quality>.<width>.<extension>/<filename>.<extension>
```
ompandeyy: @richardhand, I just noticed that I’ve missed adding `timestamp` suffixe here. Could you please add that as well if you are planning to merge today? Or I can add it tomorrow.

richardhand: @ompandeyy I was under the impression that the timestamp isn't used on publish

kwin:
That is incorrect: https://github.com/adobe/aem-core-wcm-components/issues/48

richardhand:
Thanks @kwin, I guess I missed that change.

Added b46102217020de728 to:
• Ensure we have the correct pattern documented
• Add a test case for a local upload with missing `./fileName` parameter

Unless there are no further comments, I plan to merge this early tomorrow.

cc @ompandeyy, @jckautzmann, @vladbailescu

richardhand:
Merged, many thanks @ompandeyy for the effort, and @kwin for your input.

PR #438, opened 2019-02-04
Update npm dependencies

abmaonline:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Tests Added + Pass?      | 
| Documentation Provided   | 
| Any Dependency Changes?  | aemfed, npm-run-all and stylelint

Update of the npm dependencies.

Mainly for update of aemfed with better Windows support and because npm doesn't automatically install patches and minor updates when module is still in 0.x.x version range (see [npm semver - Caret Ranges ^1.2.3 ^0.2.5 ^0.0.4](https://www.npmjs.com/package/semver#caret-ranges-123-025-004)).

Not sure what the policy on the package-lock is, so did check it in.

There is also an update for eslint, but it went from 4.x to 5.x, meaning it probably has some breaking changes.

### DIFF
--- a/content/package-lock.json
+++ b/content/package-lock.json
@@ -212,6 +212,12 @@
             "integrity": "sha512-yprFYuno9FtNsSHVlSWd+nRlmGoAbqbeCwOryP6sC/zoCjhpArcRMYp19EvpSUSizJAlsXEwJv+wcWS9XaXdMw==",
             "dev": true
         },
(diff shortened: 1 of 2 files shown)


jckautzmann:
@abmaonline Thx for your contribution! I just merged the changes.

PR #473, opened 2019-02-19
[Library] General improvements

richardhand:
- uses correct proxy page resource type
- fixes display on small devices following viewport meta tag introduction
- hosts Source Sans Pro and Source Code Pro font files - which have problems loading on safari / mobile safari via Google.

Note: The font introduces ~.6MB


jckautzmann:
Approving although I'm not a CSS expert :)

PR #479, opened 2019-02-25
[Separator] the JSON export fails

jckautzmann:
- introduce a Sling Model


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/SeparatorImpl.java
```diff
@@ -0,0 +1,45 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2017 Adobe Systems Incorporated
```
richardhand: 2019

PR #489, opened 2019-03-01
Image: Make the quality selector optionally always take effect

### LINKED ISSUE #472: Image: Make the quality selector optionally always take effect
**Current Behavior**
The quality selector being introduced with #254 only takes effect in case

1. the request width is not larger than the original (https://github.com/adobe/aem-core-wcm-components/blob/8a47281f2831f8a841756e280b4e5cb22f3c8949/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L407) and
2. no rendition exists with exactly the requested width (https://github.com/adobe/aem-core-wcm-components/blob/8a47281f2831f8a841756e280b4e5cb22f3c8949/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L395)

(both only in case there is no additional transformation being added on top)

**Expected behavior/code**
In some cases the quality is reduced to a level, where the result would be even smaller than the already less wide original and also smaller than precompiled renditions. So at least optionally it should be possible to alway recalculate the image (especially when you do aggressive caching in the web server/CDN)

**Environment**
- AEM 6.4.1
- Core Components version 2.2.2

**Possible Solution**
Add an additional flag which forces the `AdaptiveImageServlet` to always rerender the image!

bpauli:
- add design policy property jpegForceQuality with checkbox
- extend quality selector handling with optional f prefix to enforce quality

fixes #472

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -88,6 +88,7 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
     private static final String IMAGE_RESOURCE_TYPE = "core/wcm/components/image";
     static final int DEFAULT_RESIZE_WIDTH = 1280;
     public static final int DEFAULT_JPEG_QUALITY = 82; // similar to what is the default in com.day.image.Layer#write(...)
+    public static final boolean DEFAULT_JPEG_FORCE_QUALITY = false;
     private static final Logger LOGGER = LoggerFactory.getLogger(AdaptiveImageServlet.class);
     private static final String DEFAULT_MIME = "image/jpeg";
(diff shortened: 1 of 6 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -208,8 +209,13 @@ protected void doGet(@Nonnull SlingHttpServletRequest request, @Nonnull SlingHtt
         if (!handleIfModifiedSinceHeader(request, response, lastModifiedEpoch)) {
             Double quality = null;
             Integer jpegQualityInPercentage = null;
+            Boolean forceQuality = false;
             if (selectors.length > 2) {
                 String qualitySelector = selectors[selectors.length - 2];
+                if (StringUtils.startsWith(qualitySelector, "f")) {
```
kwin: As this has an impact on performance the selector should only be considered in case it has been enabled in the content policy.

Issue #493, opened 2019-03-05
[ContentFragment] Move content fragment component from extension to core


philipmundt:
In order to tackle PR #340 and #360 the current content fragment component needs to moved from extension to core.

It was discussed to handle the move the following way:
* CF component code (model + servlets, datasources) and content will be moved to core; package `contentfragment` where applicable
* Paths will need to be adjusted to core (without `extension` – this will also allow for compatibility when someone still has the then legacy extensions installed)
* The "new" component will be handle as version 1 (v1) to prevent confusion in the future (and when the list component will be added)
* The extension module will then become obsolete and will be deleted (and not released in the future anymore)

jckautzmann:
Pull request:
https://github.com/adobe/aem-core-wcm-components/pull/498

jckautzmann:
The content fragment component has been moved from extenion to the Core Components.

PR #498, opened 2019-03-05
[ContentFragment] Move content fragment component from extension to core #493

philipmundt:
Moving the content fragment component to core:
* Move java code from bundle to core (into package `contentfragment` where applicable)
* Move content to core components content 
* Adjust paths to new location
* Fix JSLint issues
* Optimize imports and mass-format code

@adobe/open-source-office


parent/pom.xml
```diff
@@ -595,6 +595,14 @@
                 <scope>provided</scope>
             </dependency>
 
+            <!-- Content fragment API -->
+            <dependency>
+                <groupId>com.adobe.cq.dam</groupId>
+                <artifactId>cq-dam-cfm-api</artifactId>
+                <version>1.1.5-NPR-19008-R002</version>
```
vladbailescu: Is this available in 6.3 as well?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
```diff
@@ -13,22 +13,21 @@
...
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-import javax.annotation.PostConstruct;
-import javax.inject.Inject;
+package com.adobe.cq.wcm.core.components.internal.models.v1.contentfragment;
 
+import com.adobe.cq.dam.cfm.ContentElement;
```
vladbailescu: Trivial stuff: we normally organise imports so that first are `java.*`, then `javax.*`, then `org.*` and only then `com.*`. It would be nice if you could configure your IDE to do the same so that there are less line changes and they won't change back again when somebody touches these files.

parent/pom.xml
philipmundt: NPR's `fixVersion` is set to 6.3.0.2, so I guess this can be answered with a yes (but to be sure I would have to double-check as I simply moved the dependency from the original extension pom).

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
philipmundt: Consider it done (even though I'm not a fan of such convention).

vladbailescu: Thanks!

philipmundt: I quickly checked the other classes to see what the standard is: seems you have empty lines between all the groups. This is what I used in my tmp-profile.

pom.xml
```diff
@@ -70,9 +70,6 @@
         <module>config</module>
         <module>all</module>
         <module>examples</module>
-        <module>extension/contentfragment/bundle</module>
-        <module>extension/contentfragment/content</module>
-        <module>extension</module>
```
jckautzmann: I would still keep the extension folder in case we add stuff in it in the future.

jckautzmann:
@philipmundt I only had one minor comment about keeping the extension folder. Other than that, all good for me!

richardhand:
@philipmundt we should probably update the README.md at the repo root to remove reference to the extension. I also wonder how and where we should document this move for customers who are already using the extension, if I understand it, they would only have to update the resource type:

`core/wcm/extension/components/contentfragment/v1/contentfragment` ->
`core/wcm/components/contentfragment/v1/contentfragment`

pom.xml
philipmundt: Keeping the folder or the module? The latter would mean maintaining the pom + generating an empty content-package (which is only possible by setting the plugin config to `failOnEmptyFilter=false`). The first also doesn't make much sense IMHO. WDYT @jckautzmann @vladbailescu ?

richardhand: +1 for removing both folder and module, if they don't serve a purpose. We can always reintroduce them if there's some need later.

jckautzmann: I'm fine if we remove both folder and module.
@philipmundt Could you please verify that upgrading to the new version of the Core Components does not break existing implementations based on the `core/wcm/extension/components/contentfragment/v1/contentfragment` resource type.

philipmundt:
@richardhand RE your request for updates of the documentation, I totally agree with you. However my available is very limited therefore I agreed with @jckautzmann that he will be taking over these updates in a follow up. Hope that's okay with you!?

jckautzmann:
@richardhand I managed your feedback in a separate PR:
https://github.com/adobe/aem-core-wcm-components/pull/510

cc: @philipmundt

Issue #360, opened 2018-10-31
[ContentFragment] Refactor content fragment component


philipmundt:
In order to reuse most of the API/implementation of the content fragment component in the content fragment list component, some refactoring is necessary:

* Separate content fragment (CFM) API from component API (in order to use the CFM API in the list)
* Fix usages of ComponentExporter and ContainerExporer interface usages (CFM should not implement ContainerExporter – only the single CFM component should)
* Adjust HTML output to reflect:
```html
<article class="cmp-contentfragment" data-cmp-contentfragment-model="project/models/airports">
    <h3 class="cmp-contentfragment__title">EuroAirport</h3>
    <p class="cmp-contentfragment__description">EuroAirport Basel Mulhouse Freiburg</p>
    <dl class="cmp-contentfragment__elements">
        <div class="cmp-contentfragment__element cmp-contentfragment__element--iataCode" data-cmp-contentfragment-element-type="string">
            <dt class="cmp-contentfragment__element-title">IATA Code(s)</dt>
            <dd class="cmp-contentfragment__element-value">
                <ul>
                    <li>EAP</li>
                    <li>BSL</li>
                    <li>MLH</li>
                </ul>
            </dd>
        </div>
        <div ...>...</div>     
    </dl>
</article>
```
* We should also take the chance of a version 2 and move it from extensions to default core component space
* Additional API changes: expose `#getPath()` for `ContentFragment` and hide JSON output of `#getParagraphs()` for a `ContentElement`.

This issue is blocking/a requirement for (adobe/aem-core-wcm-components#340).

jckautzmann:
The refactoring has been done as part of the pull request:
https://github.com/adobe/aem-core-wcm-components/pull/499

PR #499, opened 2019-03-05
[ContentFragment] Refactor content fragment component (#360)

philipmundt:
This PR is based on https://github.com/adobe/aem-core-wcm-components/pull/498

It modifies the newly added content fragment component and
* Separates content fragment (exporter) model logic from component logic
* Extracts common logic into utility class and adds tests
* Refactors tests and mainly extract test classes to simplify
  maintenance
* Improves templates and HTML to reflect requirements from PM

Todos:
- [x] Merge https://github.com/adobe/aem-core-wcm-components/pull/498
- [x] Rebase after integration of https://github.com/adobe/aem-core-wcm-components/pull/498


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
```diff
@@ -440,6 +250,11 @@ public boolean isMultiLine() {
             return contentType != null && contentType.startsWith("text/") && !getData().getDataType().isMultiValue();
```
jckautzmann: You could use `isMultiValue()` for the last condition.

PR #500, opened 2019-03-05
Support for JDK 11

### LINKED ISSUE #491: [General] Support for JDK11
Because of the removed package javax.annotation in JDK11 the core components project can't be build with JDK 11.

For future release of AEM the core components should also work with JDK 11

bpauli:
- switch from jsr305 to org.jetbrains.annotations
- add javax.annotations dependency
- remove powermock
- update mockito and javassist dependency; adapt test cases

this will fix #491


testing/junit/core/src/main/java/com/adobe/cq/wcm/core/components/testing/AbstractModelTest.java
```diff
@@ -1,101 +0,0 @@
...
-import java.util.stream.Collectors;
-
-import com.google.common.collect.ImmutableSet;
-import com.google.common.reflect.ClassPath;
-
-import static org.junit.Assert.fail;
-
-public class AbstractModelTest {
-
-    public void testDefaultBehaviour(String[] packages) throws Exception {
```
vladbailescu: I think we should keep this test.

jckautzmann:
The changes in this PR are making code coverage decrease by 5%. Is it necessary? Are there plans to compensate the decrease?

PR #505, opened 2019-03-14
[Library] update library version and milestone links

richardhand:
Updates library version links for the 2.3.2 release


jckautzmann:
we should also update our wiki page dealing with release so that we don't forget this task for the next release.

Issue #503, opened 2019-03-07
Allow Lanugage Navigation to be more effectively modified by custom extensions


HitmanInWis:
At Bounteous we were considering modifying the behavior of the Language Navigation to make one small changes.

Given this scenario:
- /content/site/en/lvl1/lvl2
- /content/site/fr/lvl1

WCM Core functionality:
- FR link on /content/site/en/lvl1.html goes to /content/site/fr/lvl1.html (same page)
- FR link on /content/site/en/lvl1/lvl2.html goes to /content/site/fr.html (home page)

Desired functionality:
- FR link on /content/site/en/lvl1.html goes to /content/site/fr/lvl1.html (same page)
- FR link on /content/site/en/lvl1/lvl2.html goes to /content/site/fr/lvl1.html (closest parent page)

Now I know we could submit a PR to WCM Core, and add a dialog option to change this behavior, but that would require us to then wait for the next WCM Core release.  Also, not every developer in the wild is looking to contribute back to WCM Core (for various reasons) - instead they want to *leverage* the tool.

The challenge with the current implementation and extension pattern is that to make this very slight change in component functionality a developer is basically forced to rewrite the entire sling model and dependency POJOs as custom code in the client codebase.

Before I discuss the challenges, let's discuss what this *could* look like if sling models were not hidden impl classes.  Note that this is not a discussion on the merits of hidden model impls and the proxy extension pattern, which we've debated on #466.  This is simply to show the difference of how simple this *could* be to implement versus how difficult it is to do so at current.

**Sunshine and bunnies solution**
If `LanguageNavigationImpl.class` was not hidden, custom development of a sling model extending this model could simply `@Override` the `getLocalizedPage()` function (it would need to be `protected` instead of `private`) to return the desired page path.  Everything else would remain as-is.   Clean, simple, done.

**Actual solution (given current constraints)**
With that out of the way, let's consider the "real" solution that can be achieved given the current patterns in WCM Core.

Since the developer is required to use the sling proxy extension pattern, the developer is required to  rewrite the logic to calculate the language links in their own implementation class.  In short, they must basically rewrite the entire `LanguageNavigationImpl` class *or* use the proxied super class to first `getItems()` and then change that `List<NavigationItem>` into a completely new list `NavigationItem` objects (gaining some minimal leverage at best).  Even if we go with that pattern, however, it now becomes even more challenging.

In overwriting `getItems()`, the developer is now forced to also overwrite the following line of code:
```
[...]

PR #534, opened 2019-04-05
Navigation custom templates

niekraaijmakers:
Preliminary PR to see if it's something the community will receive.

Backwards compatible adjustments to navigation component, that makes the component greatly more flexible.
-Adds possibility to choose a custom template per page on a content policy or property level to override the item, itemContent and group templates.
-Adds a secondary tree if called with secondary selector - this is good to render complex flyouts etc where 2 trees are required. 
-Adds possibility to turn off marking an item active server side, you don't want this when caching the menu in HTTP cache. Adds data attributes instead so javascript can do this.

With these adjustments I can make a great showcase example for many clients. WDYT?

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/BreadcrumbImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/BreadcrumbImpl.java
@@ -36,6 +36,8 @@ import com.adobe.cq.wcm.core.components.models.NavigationItem;
 import com.day.cq.wcm.api.Page;
 import com.day.cq.wcm.api.designer.Style;
 
+import static com.adobe.cq.wcm.core.components.internal.models.v1.navigation.StyleUtils.getContentPolicyStyleFromPage;
+
 @Model(adaptables = SlingHttpServletRequest.class,
        adapters = {Breadcrumb.class, ComponentExporter.class},
        resourceType = {BreadcrumbImpl.RESOURCE_TYPE_V1, BreadcrumbImpl.RESOURCE_TYPE_V2})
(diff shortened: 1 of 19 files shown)


richardhand:
@niekraaijmakers - thanks for the pull request! Have a quick look at the [Contributing Guidelines](https://github.com/adobe/aem-core-wcm-components/blob/development/CONTRIBUTING.md#contributing-code-), you will need to sign the [Adobe Contributor License Agreement](http://opensource.adobe.com/cla.html) for us to accept contributions.

I also suggest that you create GitHub issues for each of the three problems you are trying to solve here with some more information (screenshots if possible) explaining the value they bring so that we can provide a more fine-tuned feedback on each.

At first glance:
> Adds possibility to choose a custom template per page on a content policy or property level to override the item, itemContent and group templates.
This seems already available given the ability to override an HTL template in your proxy component.

> Adds a secondary tree if called with secondary selector - this is good to render complex flyouts etc where 2 trees are required.
As I understand, this is about adding a modifier `*--secondary` to the component. Have you considered using the [Style System](https://helpx.adobe.com/experience-manager/6-3/release-notes/style-system-fp.html) - it can apply a style class to the component wrapper div, and should allow you to render this alternative flyout navigation.

vladbailescu:
Hi Niek, I'm not sure we understand exactly how you plan to use the secondary tree, it would be nice if you could add some details.

As for extending the models, I'm afraid with our rule of throwing an `UnsupportedOperationException`, we would need to create a new version. We have an ongoing discussion about keeping the current pattern though (see #503)

niekraaijmakers:
Hey Vlad, sorry for the late reply. Being a consultant and all and currently no bench time whatsoever it's a bit tough to contribute.

Regarding the secondary tree, this is really because most more fancy menu structures these days require this, for comprehensive fly-outs . It makes it much more tough and maintainable for the front-end to code only based on 1 nested structure.
Especially if in the fly-out you have images, stages, trailers etc.

Ok if we have to create a new version, fine by me :)

gabrielwalt:
@niekraaijmakers Do you have the intention to bring this further or are you currently too busy?
If you'd like to continue, we should probably open a corresponding issue where we create some simple specification that describes how we want this to work.

gabrielwalt:
Closing this for now, please reopen if you want to continue to work on this topic, or if you have more information describing the intent or use-case.

Issue #572, opened 2019-05-07
[Carousel] Lack of tabindex on cmp-carousel__indicator


rogemus:
## Bug Report

**Current Behavior**
- Right now the user is unable to navigate to `cmp-carousel__indicator` using only the keyboard. 
- No corresponding slide index on the indicators and because of that is required additional js logic to change slide by clicking on the indicator. 

**Expected behavior/code**
- `tabindex` on `.cmp-carousel__indicator`
- corresponding slide index on `.cmp-carousel__indicator` as data attribute

**Environment**
- Core Components version 2.3.3-SNAPSHOT

**Possible Solution**
```
<li data-sly-repeat.item="${carousel.items}"
    role="tab"
    tabindex="0"
    data-cmp-slide-index="${itemList.index}"
    class="cmp-carousel__indicator${itemList.first ? ' cmp-carousel__indicator--active' : ''}"
    data-cmp-hook-carousel="indicator">
    ${item.title}
</li>
```

richardhand:
@rogemus - Thanks for the report. I'm assuming here that you aren't using the default client library provided with the component `core.wcm.components.carousel.v1` [0] and instead would like to use a 3rd party widget?

The Carousel (and Tabs) JS widgets were designed to manage tabindex such that only one indicator may be tabbed to `tabindex=0`, all others are not tabbable `tabindex=-1`. Navigation between indicators is handled with left (up), right (down) keys, rather than tab keys. This means a user can tab between the main actions of the Carousel easily (left action, right action and indicators).

This seems to be incompatible with the way your 3rd party widget expects the accessibility to work. In this case, I would suggest you add a tab index in your proxy component template to fit your usage.

I have also created a pull request [1] that sets `tabindex` and `aria-selected` attributes in the template to set the initial state and to clarify the default accessibility behaviour.

> No corresponding slide index on the indicators and because of that is required additional js logic to change slide by clicking on the indicator.
It should be fairly simple one-liner to add an index test if you have the htmlElement in your click handler
`[].indexOf.call(element.parentNode.children, element);`. As reading an additional data attribute doesn't provide much more convenience, I think I would rather not expose one.

[0] - https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/clientlibs/site
[1] - https://github.com/adobe/aem-core-wcm-components/pull/575

rogemus:
[...]

PR #575, opened 2019-05-07
[Carousel] Lack of tabindex on cmp-carousel__indicator

richardhand:
- adds `tabindex`, `aria-selected` initial state for Carousel (indicators) and Tabs components.

The JS Widgets for the components setup both properties when loaded, we add them here to the templates for ensuring the initial state and to provide clarity on the expected accessibility behaviour.

fixes: #572


richardhand:
Closing this as the OOTB JS widget already augments the tabs / carousel with a managed tabindex.

PR #593, opened 2019-05-16
[Accordion] Create accordion component

brandonmaynard:
| Fixed Issues?            | N/A 
| Patch: Bug Fix?          | N/A
| Minor: New Feature?      | New component

Added a new Accordion component using the PanelContainer similar to the Carousel and the Tabs components.

Here is an animated gif demoing keyboard and mouse functionality (tab, up, down, left, right, home, end, space & enter):
ScreenRecording20190517at5

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/_cq_dialog/.content.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/_cq_dialog/.content.xml
@@ -79,10 +79,6 @@
                                                 fieldLabel="Heading Type"
                                                 name="./headingType">
                                                 <items jcr:primaryType="nt:unstructured">
-                                                    <h1
-                                                        jcr:primaryType="nt:unstructured"
-                                                        text="H1"
-                                                        value="h1"/>
                                                     <h2
                                                         jcr:primaryType="nt:unstructured"
                                                         text="H2"
--- a/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/_cq_dialog/.content.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/_cq_dialog/.content.xml
@@ -72,6 +72,39 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
+                                            <headingType
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="/libs/granite/ui/components/coral/foundation/form/select"
+                                                fieldDescription="The heading type each accordion's label is."
+                                                fieldLabel="Heading Type"
+                                                name="./headingType">
+                                                <items jcr:primaryType="nt:unstructured">
+                                                    <h1
+                                                        jcr:primaryType="nt:unstructured"
+                                                        text="H1"
+                                                        value="h1"/>
+                                                    <h2
+                                                        jcr:primaryType="nt:unstructured"
+                                                        text="H2"
+                                                        value="h2"/>
+                                                    <h3
+                                                        jcr:primaryType="nt:unstructured"
+                                                        text="H3"
+                                                        value="h3"/>
+                                                    <h4
+                                                        jcr:primaryType="nt:unstructured"
+                                                        text="H4"
+                                                        value="h4"/>
+                                                    <h5
+                                                        jcr:primaryType="nt:unstructured"
+                                                        text="H5"
+                                                        value="h5"/>
+                                                    <h6
+                                                        jcr:primaryType="nt:unstructured"
+                                                        text="H6"
+                                                        value="h6"/>
+                                                </items>
+                                            </headingType>
                                             <expandedItem
                                                 jcr:primaryType="nt:unstructured"
                                                 sling:resourceType="/libs/granite/ui/components/coral/foundation/form/hidden"
--- a/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/_cq_dialog/.content.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/_cq_dialog/.content.xml
@@ -1,0 +1,108 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  ~ Copyright 2019 Adobe Systems Incorporated
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~-->
+<jcr:root xmlns:sling="http://sling.apache.org/jcr/sling/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
+          xmlns:nt="http://www.jcp.org/jcr/nt/1.0" xmlns:cq="http://www.day.com/jcr/cq/1.0"
+          xmlns:granite="http://www.adobe.com/jcr/granite/1.0"
+    jcr:primaryType="nt:unstructured"
+    jcr:title="Accordion"
+    sling:resourceType="cq/gui/components/authoring/dialog"
+    extraClientlibs="[core.wcm.components.commons.editor.dialog.childreneditor.v1,core.wcm.components.accordion.v1.editor]"
+    helpPath="https://www.adobe.com/go/aem_cmp_accordion_v1"
+    trackingFeature="core-components:accordion:v1">
+    <content
+        granite:class="cmp-accordion__editor"
+        jcr:primaryType="nt:unstructured"
+        sling:resourceType="granite/ui/components/coral/foundation/container">
+        <items jcr:primaryType="nt:unstructured">
+            <tabs
+                jcr:primaryType="nt:unstructured"
+                sling:resourceType="granite/ui/components/coral/foundation/tabs"
+                maximized="{Boolean}true">
+                <items jcr:primaryType="nt:unstructured">
+                    <accordionitems
+                        jcr:primaryType="nt:unstructured"
+                        jcr:title="Items"
+                        sling:resourceType="granite/ui/components/coral/foundation/container"
+                        margin="{Boolean}true">
+                        <items jcr:primaryType="nt:unstructured">
+                            <columns
+                                jcr:primaryType="nt:unstructured"
+                                sling:resourceType="granite/ui/components/coral/foundation/fixedcolumns"
+                                margin="{Boolean}true">
+                                <items jcr:primaryType="nt:unstructured">
+                                    <column
+                                        jcr:primaryType="nt:unstructured"
+                                        sling:resourceType="granite/ui/components/coral/foundation/container">
+                                        <items jcr:primaryType="nt:unstructured">
+                                            <accordionitems
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="core/wcm/components/commons/editor/dialog/childreneditor/v1/childreneditor"/>
+                                        </items>
+                                    </column>
+                                </items>
+                            </columns>
+                        </items>
+                    </accordionitems>
+                    <properties
+                        jcr:primaryType="nt:unstructured"
+                        jcr:title="Properties"
+                        sling:resourceType="granite/ui/components/coral/foundation/container"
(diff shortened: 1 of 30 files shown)


richardhand:
Thanks @brandonmaynard - this looks like a really good start! A couple of things - I think you will need to recreate the PR after [signing the Adobe CLA](http://opensource.adobe.com/cla.html), and when doing that it would be great if you could base against the development rather than master branch.

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
```diff
@@ -0,0 +1,43 @@
...
+  ~ limitations under the License.
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
+
+<div data-sly-use.accordion="com.adobe.cq.wcm.core.components.models.Accordion"
+     data-sly-use.templates="core/wcm/components/commons/v1/templates.html"
+     class="cmp-accordion"
+     data-cmp-is="accordion">
+    <ol data-sly-test="${accordion.items && accordion.items.size > 0}"
+        class="cmp-accordion__tablist"
+        role="tablist"
```
jakublamprecht: Roles tablist, tab and tabpanel should not be used with accordions (this refers also to line 28 and 35)

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
```diff
@@ -0,0 +1,43 @@
...
+        aria-multiselectable="false">
+        <li data-sly-repeat.item="${accordion.items}"
+            class="cmp-accordion__item">
+            <div class="cmp-accordion__item--title${item.name == accordion.expandedItem ? ' initially-expanded' : ''}"
```
jakublamprecht: According to accessibility practices the title of accordion should be wrapped with an element with role="button" (e.g `<button>`), which then should be wrapped in element that serves as a heading (e.g. `<h3>` element).

Also the heading in which the title is wrapped should probably be authorable to make sure, that heading structure throughout the page is correct.

sources:
https://davatron5000.github.io/a11y-nutrition-cards/components/accordion
https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
```diff
@@ -0,0 +1,43 @@
...
+                 role="tab"
+                 data-cmp-hook-accordion="item">
+                ${item.title}
+                <span class="cmp-accordion__item--title--icon"></span>
+            </div>
+
+            <div data-sly-resource="${item.name @ decorationTagName='div'}"
+                 role="tabpanel"
```
jakublamprecht: Role on the accordion panel content should be equal to region

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
```diff
@@ -0,0 +1,43 @@
...
+        <li data-sly-repeat.item="${accordion.items}"
+            class="cmp-accordion__item">
+            <div class="cmp-accordion__item--title${item.name == accordion.expandedItem ? ' initially-expanded' : ''}"
+                 role="tab"
+                 data-cmp-hook-accordion="item">
+                ${item.title}
+                <span class="cmp-accordion__item--title--icon"></span>
+            </div>
+
+            <div data-sly-resource="${item.name @ decorationTagName='div'}"
```
jakublamprecht: Right now the content of accordion is not reachable via tab sequence, maybe there should be a tabindex=0 specified?

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
jakublamprecht: Title should be reachable via tab requence, so probably tabindex=0 is missing here

jakublamprecht: According to accessibility, the header of accordion should have an `aria-controls` attribute with value set to ID of the controlled accordion panel

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
jakublamprecht: According to accessibility practices, accordion panel should have `aria-labeledby` attribute with value set to the button that controls this panel (https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html)

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
```diff
@@ -0,0 +1,371 @@
...
+         * @param {Object} itempanel Provided accordion item's itempanel
+         * @param {Object} item Provided accordion item's item
+         */
+        function collapseAccordionItem(itempanel, item) {
+            if (itempanel && item) {
+                itempanel.classList.remove(selectors.expanded.itempanel);
+                itempanel.setAttribute("aria-hidden", true);
+                item.classList.remove(selectors.expanded.item);
+                item.setAttribute("aria-selected", false);
+                item.setAttribute("tabindex", "-1");
```
jakublamprecht: I'm not sure if the inactive accordion headers should have tabindex set to -1 - they should still be reachable via tab sequence, even if their content is collapsed

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
```diff
@@ -0,0 +1,371 @@
...
+         * @private
+         * @param {Object} itempanel Provided accordion item's itempanel
+         * @param {Object} item Provided accordion item's item
+         */
+        function expandAccordionItem(itempanel, item) {
+            if (itempanel && item) {
+                itempanel.classList.add(selectors.expanded.itempanel);
+                itempanel.removeAttribute("aria-hidden");
+                item.classList.add(selectors.expanded.item);
+                item.setAttribute("aria-selected", true);
```
jakublamprecht: Accordions should use aria-expanded instead of aria-selected (https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html)

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/_cq_dialog/.content.xml
```diff
@@ -72,6 +72,39 @@
...
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
+                                            <headingType
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="/libs/granite/ui/components/coral/foundation/form/select"
+                                                fieldDescription="The heading type each accordion's label is."
+                                                fieldLabel="Heading Type"
+                                                name="./headingType">
+                                                <items jcr:primaryType="nt:unstructured">
+                                                    <h1
```
brandonmaynard: I included an `h1` option but I'm wondering if this should be removed for SEO reasons. Does anyone have any thoughts/opinions on this?

richardhand: @brandonmaynard - what SEO problems do you foresee? H1-H6 seems fine to me, unless there's good reason we shouldn't restrict it.

brandonmaynard: @richardhand, I thought for SEO, there should only be one `h1` per page and this property sets all Accordion Item Titles to the same configured heading value

richardhand: Right, I was imagining each accordion item being wrapped by a sectioning element, in which case it would be fine, but since we don't do that I think we can/should omit H1 as you say. Would love an expert SEO opinion though.

brandonmaynard: Ok, I removed `h1`. Thanks for your assistance @richardhand

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
jakublamprecht: I think this still applies after the accessibility changes

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
```diff
@@ -0,0 +1,44 @@
...
+  ~ limitations under the License.
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
+
+<div data-sly-use.accordion="com.adobe.cq.wcm.core.components.models.Accordion"
+     data-sly-use.templates="core/wcm/components/commons/v1/templates.html"
+     class="cmp-accordion"
+     data-cmp-is="accordion">
+    <sly data-sly-test="${accordion.items && accordion.items.size > 0}"
+         data-sly-repeat.item="${accordion.items}">
+        <div class="cmp-accordion__item${item.name == accordion.expandedItem ? ' initially-expanded' : ''}"
```
jakublamprecht: Right now it's an element with role button wrapping the heading, but by standard this should be reversed - heading should be wrapping the button

source: https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html
quote: "Each accordion header element contains a button that controls the visibility of its content panel"

> 

>

richardhand: @brandonmaynard before going ahead with this change, I've made a second version on top of your changes which has the markup structure as @jakublamprecht describes.

Among the changes I switched `getExpandedItem` to `getExpandedItems` so that multiple items can be initially expanded and added an `isSingleExpansion` option so that the accordion can be configured to only allow one item to be expanded at a time. Will create a follow up PR probably tomorrow for us to review.

brandonmaynard:
Reopening to pass CLA check

brandonmaynard:
@richardhand , I signed the CLA and created a new branch off of `development` and opened https://github.com/adobe/aem-core-wcm-components/pull/600. I am closing this one out.

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
brandonmaynard: I don't think we want the Accordion Item's actual expandable/collapsable contents to be accessible via the tab key. I was under the impression that only links and clickable 'areas' should be accessible by the tab key, right? I could be wrong because I am by no means an accessibility expert, but that was my assumption.

richardhand: I agree here, if there's a focusable element in the item panel then it will naturally gain focus. I don't think we don't want focus to shift to the panel if there's nothing focusable there, but I'm also no accessibility expert.

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
richardhand: This one requires a random unique ID generation approach for all the UI components that is consistent and ensures that it doesn't break the HTML document. Would also apply to tabs and carousel.

richardhand: Perhaps it's time to create a `Cmp.commons` object which can provide `Cmp.commons.getID` that returns a namespaced unique ID for any component that asks for one.

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
jakublamprecht: Hey @richardhand, I've found another little issue with the markup related to the aria-expanded attribute - I've added a comment on #600 if you want to take a look at it too

jakublamprecht: Also, do we need the initially-expanded class? Wouldn't it be possible to achieve the same result with just .cmp-accordion__item--expanded and .cmp-accordion__itempanel--expanded classes?

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
jakublamprecht: Sorry guys, you’re right, thought that was said in the w3c recomendations, but it was for the tabs component and Not accordion.

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
richardhand: This should be solved in #601, as we now use `button` controls for which tabindex isn't required.

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
richardhand: @jakublamprecht - I addressed both of your above comments in #601, I replaced the initially-expanded concept with a data attribute configuration on the item - `data-cmp-expanded` and we now provide `cmp-accordion__panel--expanded` (on the panel) and `cmp-accordion__button--expanded` (on the control) classes for styling.

PR #600, opened 2019-05-22
[Accordion] Create accordion component

brandonmaynard:
| Fixed Issues?            | N/A 
| Patch: Bug Fix?          | N/A
| Minor: New Feature?      | New component

Added a new Accordion component using the PanelContainer similar to the Carousel and the Tabs components.

Here is an animated gif demoing keyboard and mouse functionality (tab, up, down, left, right, home, end, space & enter):
ScreenRecording20190517at5

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
@@ -1,0 +1,43 @@
+<!--/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  ~ Copyright 2019 Adobe Systems Incorporated
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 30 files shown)


brandonmaynard:
Created a new branch off of `development` and resubmitted. Original PR was https://github.com/adobe/aem-core-wcm-components/pull/593

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
```diff
@@ -0,0 +1,44 @@
...
+
+<div data-sly-use.accordion="com.adobe.cq.wcm.core.components.models.Accordion"
+     data-sly-use.templates="core/wcm/components/commons/v1/templates.html"
+     class="cmp-accordion"
+     data-cmp-is="accordion">
+    <sly data-sly-test="${accordion.items && accordion.items.size > 0}"
+         data-sly-repeat.item="${accordion.items}">
+        <div class="cmp-accordion__item${item.name == accordion.expandedItem ? ' initially-expanded' : ''}"
+             data-cmp-hook-accordion="item"
+             aria-expanded="false"
```
jakublamprecht: I'm not sure if this should be in the markup tbh. What if a user without JavaScript enabled opens the page? All of the accordion items will have aria-expanded false even though, all of the itempanels should be opened (so that the content is also visible without Javascript enabled). This attribute should probably be added in JavaScript when component is initialized.

richardhand: See #601.

Update this at https://github.com/adobe/aem-core-wcm-components/pull/601/files#diff-c0a2070bfbcc3a3938ac8a648d548d54R28 to set the initial `aria-expanded` and `aria-states hidden` based on the item's configured expanded state.

PR #601, opened 2019-05-23
[Accordion] Accordion component additions

### LINKED ISSUE #583: [Accordion] Create Accordion component
Hi

I really like Core Components idea and ways of working.
In our project we need to deliver Accordion component,  when we look at Core Components we can see very similar component: Tabs. There is only one feature missing, Tabs component doesn’t have an option for managing panels visibility like all-open and all-closed. Tabs have only n-th open.

What is your recommendation in deliver Accordion Component?

CQ-4256161

richardhand:
@brandonmaynard - I merged your initial pull request to a `feature/accordion` branch. Here follows the changes I mentioned yesterday added on top of your (excellent!) initial work. If we review these, I can create a final PR against `development`.

### Changes
- adds `isSingleExpansion()`.
- moves `getExpandedItem()` -> `getExpandedItems()`.
- allows definition of allowed heading elements and default heading element in the policy.
- refactors markup for new functionality and addresses accessibility items raise in #593 and #600 .
- refactors UI widget for new functionality.
- refactors Edit Dialog for new functionality.
- reduces OOTB CSS.
- adds accordion examples to component library.
- adjusts component icon.
- adjusts JUnit tests according to the new functionality.
- adds accordion to main README.md.

### Edit Dialog
Screenshot 2019-05-23 at 17 41 22

Screenshot 2019-05-23 at 17 41 30

### Policy Dialog
Screenshot 2019-05-23 at 17 42 49

### Component Library
Screenshot 2019-05-23 at 16 05 27
Screenshot 2019-05-23 at 16 05 37
Screenshot 2019-05-23 at 17 47 23

| Fixed Issues?            | `Fixes #583` 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Tests Added + Pass?      | 👍 adjusted JUnit tests
| Documentation Provided   | 👍 adjusted `README.md` files (code comments and or markdown)
| Any Dependency Changes?  |


content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
```diff
@@ -13,32 +13,30 @@
...
-                class="cmp-accordion__item__title"
-                role="heading">
-                ${item.title}
-                <span class="cmp-accordion__item__title__icon"></span>
-            </h3>
-        </div>
-
+     data-cmp-is="accordion"
+     data-cmp-single-expansion="${accordion.singleExpansion}">
+    <div data-sly-test="${accordion.items && accordion.items.size > 0}"
```
jakublamprecht: That's cool that we have a wrapper for each item now :) Just wondering, maybe these should be `<li>` tags inside and unordered list now?

richardhand: Does it add value? If not, I think `<div>` are cleaner - they don't require any CSS reset.

jakublamprecht: Well true, on a second thought I will agree.

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
```diff
@@ -13,32 +13,30 @@
...
+         data-sly-repeat.item="${accordion.items}"
+         class="cmp-accordion__item"
+         data-cmp-hook-accordion="item"
+         data-cmp-expanded="${item.name in accordion.expandedItems}">
+        <h3 data-sly-element="${accordion.headingElement @ context='elementName'}"
+            class="cmp-accordion__header">
+            <button class="cmp-accordion__button${item.name in accordion.expandedItems ? ' cmp-accordion__button--expanded' : ''}"
+                    aria-expanded="${item.name in accordion.expandedItems ? 'true' : 'false'}"
```
jakublamprecht: I'm still not sure about whether I like this being in the markup. I think this should be added via JS on initialization, so that if JavaScript is disabled on page load, this, along with the `aria-hidden` on panel does not misdirect the user.

This is also not consistent with tabs component, which does not add those attributes until JavaScript initialization: https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/tabs.html

richardhand: But we do add the `.cmp-accordion__panel--expanded` classes, so these panels _will_ be hidden by default (as long as the correct CSS is provided). So the classes `aria` attributes aren't inconsistent with what the user would experience.

Right, we should align with the tabs component.

jakublamprecht: Hmm, I'm not sure if that's right then. Shouldn't users be able to see all of the content if there's no possibility to open the panels? Maybe instead of `aria-hidden` we could add the `hidden` attribute, which will have the same meaning for screen-readers and hide the panels based on that?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AccordionImpl.java
```diff
@@ -40,26 +48,60 @@
...
+    private String[] expandedItems;
 
-    private String expandedItemName;
+    @ValueMapValue(optional = true)
+    private String headingElement;
+
+    @ScriptVariable
+    private Style currentStyle;
+
+    private String[] expandedItemNames;
```
jckautzmann: Having `expandedItemNames` and `expandedItems` is a bit confusing. I would at least add some docs to clarify.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AccordionImpl.java
```diff
@@ -40,26 +48,60 @@
...
+
+    /**
+     * The {@link com.adobe.cq.wcm.core.components.internal.Utils.Heading} object for the HTML element
+     * to use for accordion headers.
+     */
+    private Utils.Heading heading;
```
jckautzmann: Shouldn't we have a similar approach in the Tabs component then?

richardhand: Tabs is a slightly different concept, it isn't recommended to use heading level elements there:
https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

The accordion design pattern does recommend them: https://www.w3.org/TR/wai-aria-practices/#accordion

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Accordion.java
```diff
@@ -26,22 +26,40 @@
...
     /**
-     * Returns the expanded item.
+     * Name of the configuration policy property that stores the default value for the accordion heading's HTML element.
      *
-     * @return The expanded item
+     * @see #getHeadingElement()
      * @since com.adobe.cq.wcm.core.components.models 12.8.0
      */
-    default String getExpandedItem() {
+    String PN_DESIGN_DEFAULT_HEADING_ELEMENT = "headingElement";
```
jckautzmann: I'd rename it `PN_HEADING_ELEMENT` to map to the JCR property.

richardhand: Yep, but probably `PN_DESIGN_HEADING_ELEMENT` as it's coming from the policy and we use this naming elsewhere.

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
richardhand: @jakublamprecht - thanks! I'll try adding this progressive enhancement such that the `aria` attributes and the css classes are applied by the widget (meaning all panels are open in the no-js situation).

Re. the `hidden` attribute I would rather keep the style classes and accessibility attributes separate for clarity.

richardhand:
merging here and creating a new PR against `development` to make review easier.

PR #626, opened 2019-06-12
[Download] Remove default action text

richardhand:
Removes default action text from download policy configuration.


bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v1/DownloadImplTest.java
```diff
@@ -174,19 +174,6 @@ public void testDownloadWithCustomActionText() throws Exception
...
-    public void testDownloadWithActionTextFromStyle() throws Exception
-    {
-        Resource mockResource = mock(Resource.class);
-        Style mockStyle = new MockStyle(mockResource, new MockValueMap(mockResource, new HashMap() {{
-            put(Download.PN_ACTION_TEXT, STYLE_ACTION_TEST);
-        }}));
-
-        Download download = getDownloadUnderTest(DOWNLOAD_2, mockStyle);
-        assertEquals("Expected action text is not correct", STYLE_ACTION_TEST, download.getActionText());
-        Utils.testJSONExport(download, Utils.getTestExporterJSONPath(TEST_BASE, DOWNLOAD_WITH_ACTION_TEXT_FROM_STYLE));
```
vladbailescu: Wouldn't it make sense to also remove the `DOWNLOAD_WITH_ACTION_TEXT_FROM_STYLE` constant and referenced JSON.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/DownloadImpl.java
```diff
@@ -123,9 +123,6 @@ protected void initModel() {
         descriptionFromAsset = properties.get(PN_DESCRIPTION_FROM_ASSET, descriptionFromAsset);
         inline = properties.get(PN_INLINE, inline);
         if (currentStyle != null) {
-            if (StringUtils.isBlank(actionText)) {
-                actionText = currentStyle.get(PN_ACTION_TEXT, String.class);
```
vladbailescu: Shouldn't you also remove `PN_ACTION_TEXT` constant?

richardhand: This one is still used in the edit dialog

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v1/DownloadImplTest.java
richardhand: Good spot, should have checked that, updated in 551b9b9

richardhand:
Thanks @vladbailescu, addressed both comments.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/DownloadImpl.java
vladbailescu: Hm, you're right, we should then keep the constant in the interface, in case people want to reference the property in their implementation. So you can rollback 303098b, sorry!

PR #628, opened 2019-06-13
[Download] Clean up output markup

bpauli:
- don't display action text markup if action text or download url is empty
- display metadata as description list element
- display download icon as part of the metadata

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/download/v1/download/download.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/download/v1/download/download.html
@@ -21,14 +21,17 @@
         <a data-sly-unwrap="${!download.url}" href="${download.url}">${title}</a>
     </h3>
     <div class="cmp-download__description" data-sly-test.description="${download.description}">${description @ context="html"}</div>
-    <div data-sly-test="${(download.displayFilename && download.filename) || (download.displaySize && download.size) || (download.displayFormat && download.format)}"
-         class="cmp-download__metadata">
-        <span data-sly-test="${download.displayFilename && download.filename}" class="cmp-download__filename">${download.filename}</span>
-        <span data-sly-test="${download.displaySize && download.size}" class="cmp-download__size">${download.size}</span>
-        <span data-sly-test="${download.displayFormat && download.format}" class="cmp-download__format">${download.format}</span>
-    </div>
-    <a data-sly-unwrap="${!download.url}" class="cmp-download__action" href="${download.url}">
-        <span class="cmp-download__action-icon${download.extension ? ' cmp-download__action-icon--${download.extension}' : ''}"></span>
+    <dl class="cmp-download__metadata">
+        <dt data-sly-test="${download.displayFilename && download.filename}" class="cmp-download__filename">${'Filename' @ i18n}</dt>
+        <dd data-sly-test="${download.displayFilename && download.filename}" class="cmp-download__filename">${download.filename}</dd>
+        <dt data-sly-test="${download.displaySize && download.size}" class="cmp-download__size">${'Size' @ i18n}</dt>
+        <dd data-sly-test="${download.displaySize && download.size}" class="cmp-download__size">${download.size}</dd>
+        <dt data-sly-test="${download.displayFormat && download.format}" class="cmp-download__format">${'Format' @ i18n}</dt>
+        <dd data-sly-test="${download.displayFormat && download.format}" class="cmp-download__format">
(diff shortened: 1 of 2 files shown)


content/src/content/jcr_root/apps/core/wcm/components/download/v1/download/download.html
```diff
@@ -21,14 +21,17 @@ <h3 class="cmp-download__title" data-sly-test.title="${download.title}" data-sly
...
     <div class="cmp-download__description" data-sly-test.description="${download.description}">${description @ context="html"}</div>
-    <div data-sly-test="${(download.displayFilename && download.filename) || (download.displaySize && download.size) || (download.displayFormat && download.format)}"
-         class="cmp-download__metadata">
-        <span data-sly-test="${download.displayFilename && download.filename}" class="cmp-download__filename">${download.filename}</span>
-        <span data-sly-test="${download.displaySize && download.size}" class="cmp-download__size">${download.size}</span>
-        <span data-sly-test="${download.displayFormat && download.format}" class="cmp-download__format">${download.format}</span>
-    </div>
-    <a data-sly-unwrap="${!download.url}" class="cmp-download__action" href="${download.url}">
-        <span class="cmp-download__action-icon${download.extension ? ' cmp-download__action-icon--${download.extension}' : ''}"></span>
+    <dl class="cmp-download__metadata">
```
richardhand: @pauli I notice you removed the (long) check for has metadata. Shouldn't we hide the metadata element if no metadata is available?

content/src/content/jcr_root/apps/core/wcm/components/download/v1/download/download.html
```diff
@@ -21,14 +21,17 @@ <h3 class="cmp-download__title" data-sly-test.title="${download.title}" data-sly
...
+        <dt data-sly-test="${download.displayFilename && download.filename}" class="cmp-download__filename">${'Filename' @ i18n}</dt>
+        <dd data-sly-test="${download.displayFilename && download.filename}" class="cmp-download__filename">${download.filename}</dd>
+        <dt data-sly-test="${download.displaySize && download.size}" class="cmp-download__size">${'Size' @ i18n}</dt>
+        <dd data-sly-test="${download.displaySize && download.size}" class="cmp-download__size">${download.size}</dd>
+        <dt data-sly-test="${download.displayFormat && download.format}" class="cmp-download__format">${'Format' @ i18n}</dt>
```
richardhand: I'm not sure the metadata specific BEM elements are required. I would just generalise this to a `cmp-download_metadata-label` and `cmp-download_metadata-content`.

```
    <dl class="cmp-download__metadata">
        <dt class="cmp-download__metadata-label">Filename</dt>
        <dd class="cmp-download__metadata-content">lava-into-ocean.jpg</dd>
        <dt class="cmp-download__metadata-label">Size</dt>
        <dd class="cmp-download__metadata-content">81 KB</dd>
        <dt class="cmp-download__metadata-label">Format</dt>
        <dd class="cmp-download__metadata-content">
            <span class="cmp-download__icon cmp-download__icon-jpeg"></span>image/jpeg
        </dd>
    </dl>
```

richardhand: I would also consider adding the icon span back to the action button, but not applying the extension class. You may still want to add an icon there that indicates download, or view, or whatever you see fit. And it would match the markup we introduced in the button component.

<span class="cmp-download__action-icon"></span>

bpauli: @richardhand  Specific BEM Elements make sense for the metadata as this will give the designer the possibility to apply different styles to different metadata information (e.g. bigger font size to highlight the filename). Specific label and content names are not necessary as this is already implied by the element names `dt/dd`.

richardhand: @pauli, it's more likely a designer would want to style each of these elements the same, and right now they don't have a handle to do that. I don't see a use case in styling these differently.

re. `dt/dd` HTML element names shouldn't be styled based on the element:
```
// bad
dt {
  // my style   
}
```

rather, always via CSS classes:
```
// good
.cmp-download__property-label {
  // my style
}
```

richardhand: See, the below interpretation of this component by Kilian:
Screenshot 2019-06-13 at 15 07 32

PR #632, opened 2019-06-17
[Architecture] Change model annotations to @ProviderType #631

### LINKED ISSUE #631: [Architecture] Change model annotations to @ProviderType
## Feature Request

Currently, the model interfaces are either annotated with `@ConsumerType` or not at all.

O'Reilly describes the difference between OSGi consumer types and provider types as following (from [https://www.oreilly.com/library/view/building-modular-cloud/9781449345143/ch04.html](https://www.oreilly.com/library/view/building-modular-cloud/9781449345143/ch04.html)
> Some interfaces are implemented by, at most, a few implementations but used by many. LogService is an example of this; there are only a handful of libraries that implement the LogService interface, but the interface is used everywhere code needs to log something. We call such interfaces provider types. Other interfaces are used the other way around: they are implemented many times but are only used by a handful of implementations. javax.servlet.Filter is an example of this; many servlet filters are created by developers, but they are only consumed by a handful of application servers. We call such interfaces consumer types.

Based on that description, the core components model interfaces should be of `@ProviderType` because they are usually implemented only once.

This annotation exists only during compile-time and is commonly used for semantic versioning.
It would make sense to either:
 - Remove all occurrences of the `@ConsumerType` annotations since everything is by default considered as a consumer type
- Annotate all models as `@ProviderType`

krakong:
| Fixed Issues?            | Fixes #631  

Removed all `@ConsumerType` annotations. Since all interfaces are by default consumer types, this shouldn't have any effect on the existing tools for version generation.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Accordion.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Accordion.java
@@ -15,14 +15,12 @@
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.models;
 
-import org.osgi.annotation.versioning.ConsumerType;
 
 /**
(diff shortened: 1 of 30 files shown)


gabrielwalt:
Closing this as it is working as designed, and there seems to be no consensus on the different approach so far.

PR #636, opened 2019-06-18
[Title] Default heading size option is not reflected in dialog

### LINKED ISSUE #629: [Title] Default heading size option is not reflected in dialog
## Bug Report

**Current Behavior**
Configuration of default heading size on policy of title component is not reflected on dialog of newly added Title component on a page using this policy.

Title component right after being added correctly uses the default heading size, but once the dialog is opened, the heading size option in this dialog is not the same as the default option, which causes the component to use different heading tag once the dialog is saved.

Reproduction steps:
* Open Title component demo page
* Go to template of the page
* Open policy of Title v2 component
* Change default heading size option to e.g. h5
* Go back to the Title component demo page
* Add new Title v2 component (component uses correct tag)
* Open dialog (different value for heading size is set than the default one set in policy)

**Expected behavior/code**
Default heading size configuration from policy is reflected in heading size option in dialog of newly added component.

**Environment**
- Core Components version 2.4.0

** More info **
This issue seems like something that could happen in more then one component and I think that the solution for this could be reused for other (even new and custom) components. Would you consider any more generic solution? 

Maybe some JS logic added on authoring level which would react once a dialog is opened and which looks for some specific attribute on dialog elements?

bpauli:
- extend TextValueDataResourceSource with selected property
- add default type to data source servlet for allowed headings
fixes #629

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AllowedHeadingElementsDataSourceServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AllowedHeadingElementsDataSourceServlet.java
@@ -51,6 +51,7 @@ public class AllowedHeadingElementsDataSourceServlet extends SlingSafeMethodsSer
     public final static String RESOURCE_TYPE_V1 = "core/wcm/components/commons/datasources/allowedheadingelements/v1";
     public final static String RESOURCE_TYPE_TITLE_V1 = "core/wcm/components/title/v1/datasource/allowedtypes";
     public final static String PN_ALLOWED_HEADING_ELEMENTS = "allowedHeadingElements";
+    public final static String PN_DEFAULT_HEADING_ELEMENT = "headingElement";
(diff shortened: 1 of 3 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AllowedHeadingElementsDataSourceServlet.java
```diff
@@ -52,6 +52,7 @@
     public final static String RESOURCE_TYPE_TITLE_V1 = "core/wcm/components/title/v1/datasource/allowedtypes";
     public final static String PN_ALLOWED_HEADING_ELEMENTS = "allowedHeadingElements";
     public final static String PN_ALLOWED_TYPES = "allowedTypes";
+    public final static String PN_DEFAULT_TYPE = "type";
```
richardhand: @pauli, we should add `PN_DEFAULT_HEADING_ELEMENT = "headingElement";` and test this also works in the accordion component.

Issue #586, opened 2019-05-13
[Navigation][Tabs][Carousel][Button] There's no possibility to add labels to the components


jakublamprecht:
## Bug Report

**Current Behavior**
Navigation component does not allow authors to specify labels for navigation, tabs and carousel (useful if more than one navigation component is present on the page) as W3C suggests here:

https://www.w3.org/WAI/tutorials/menus/structure/#label-menus
https://www.w3.org/TR/wai-aria-practices/#tabpanel

Using `aria-label` would probably require less work, since using `aria-labeledby` would come with adding possibility to choose heading level for the label.

**Expected behavior/code**
Authors can specify labels for navigation, tabs, carousel and button components.

**Environment**
- Core Components version 2.3.0

richardhand:
@jakublamprecht - thanks, this one would require an author defined label per navigation so we would have to put some thought into if and how we would like this to be defined in the dialog.

richardhand:
@gwalt, I see two ways of supporting this, the second being a bit more involved. What do you think?

### Option 1: Allow only a label to be defined by an author for accessibility reasons.

**Edit Dialog**

Field
- Type: Text
- Title: Label
- Description: Label that describes the navigation and distinguishes it from other navigations on the page, for accessibility reasons.

**Markup**
```
<nav aria-label="Main Navigation">
  …
</nav>
``` 

### Option 2: Allow a title to be defined for a navigation, which is displayed on the page and can optionally be displayed as an accessibility label.

**Edit Dialog**

Field 1
- Type: Text
- Title: Title
- Description: Navigation Title

Field 2
- Type: Select
- Title: Title Element
- Options: H1-H6

Field 3
- Type: Checkbox
- Title: Display title as accessibility label
- Description: If checked, displays the navigation title as an accessibility label.
- Default: unchecked

**Markup**
```
<nav aria-labelledby="myNavigationId-title">
  <h3 id="myNavigationId-title">Main Navigation</h3>
    …
</nav>
```

or, if `Display title as accessibility label` is checked:
```
<nav aria-label="Main Navigation">
  …
</nav>
```

jakublamprecht:
@richardhand I changed the title and description of this issue to also include the tabs and carousel components, since it adresses the same issue in all those components.

richardhand:
[...]

PR #672, opened 2019-07-22
[Navigation][Tabs][Carousel] There's no possibility to add labels to the components

jakublamprecht:
| Fixed Issues?            | Fixes: #586 
| Patch: Bug Fix?          | Bugfix
| Minor: New Feature?      | -
| Major: Breaking Change?  | -
| Any Dependency Changes?  |

Added additional property in dialogs of Navigation, Tabs and Carousel components that allows author to specify aria-label for the component

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
@@ -82,17 +82,7 @@ public interface Carousel extends Container {
      * @return The value of label for Carousel component
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
      */
-    default String getCarouselLabel() {
-        throw new UnsupportedOperationException();
-    }
-
-    /**
-     * Returns an accessible label that describes carousel indicators.
-     *
-     * @return The value of label for indicators wrapper
-     * @since com.adobe.cq.wcm.core.components.models 12.9.0
-     */
-    default String getIndicatorsLabel() {
+    default String getLabel() {
         throw new UnsupportedOperationException();
     }
 
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
@@ -79,7 +79,7 @@ public interface Carousel extends Container {
     /**
      * Returns an accessible label that describes the carousel.
      *
-     * @return The value of label for Carousel component
+     * @return an accessible label for carousel
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
      */
     default String getLabel() {
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
@@ -77,8 +77,7 @@ public interface Carousel extends Container {
     }
 
     /**
-     * Returns the value of aria-label for Carousel component.
-     * Improves accessibility of component.
+     * Returns an accessible label that describes the carousel.
      *
      * @return The value of label for Carousel component
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
@@ -88,8 +87,7 @@ public interface Carousel extends Container {
     }
 
     /**
-     * Returns the value of aria-label for Carousel indicators.
-     * Improves accessibility of component.
+     * Returns an accessible label that describes carousel indicators.
      *
      * @return The value of label for indicators wrapper
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
@@ -77,12 +77,22 @@ public interface Carousel extends Container {
     }
 
     /**
-     * Returns the value of label
+     * Returns the value of label for Carousel component
      *
-     * @return The value of label
+     * @return The value of label for Carousel component
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
      */
-    default String getLabel() {
+    default String getCarouselLabel() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Returns the value of label for indicators wrapper
+     *
+     * @return The value of label for indicators wrapper
+     * @since com.adobe.cq.wcm.core.components.models 12.9.0
+     */
+    default String getIndicatorsLabel() {
         throw new UnsupportedOperationException();
     }
 
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
@@ -76,4 +76,14 @@ public interface Carousel extends Container {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Returns the value of label
+     *
+     * @return The value of label
+     * @since com.adobe.cq.wcm.core.components.models 12.9.0
+     */
+    default String getLabel() {
+        throw new UnsupportedOperationException();
+    }
+
 }
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
@@ -77,7 +77,8 @@ public interface Carousel extends Container {
     }
 
     /**
-     * Returns the value of label for Carousel component
+     * Returns the value of aria-label for Carousel component.
+     * Improves accessibility of component.
      *
      * @return The value of label for Carousel component
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
@@ -87,7 +88,8 @@ public interface Carousel extends Container {
     }
 
     /**
-     * Returns the value of label for indicators wrapper
+     * Returns the value of aria-label for Carousel indicators.
+     * Improves accessibility of component.
      *
      * @return The value of label for indicators wrapper
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
@@ -85,7 +85,7 @@ public interface Navigation extends ComponentExporter {
     /**
      * Retrieves an accessible label that describes the navigation.
      *
-     * @return the value of label for navigation
+     * @return an accessible label for navigation
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
      */
     default String getLabel() {
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
@@ -82,4 +82,14 @@ public interface Navigation extends ComponentExporter {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Retrieves the value of label for navigation.
+     *
+     * @return the value of label for navigation
+     * @since com.adobe.cq.wcm.core.components.models 12.9.0
+     */
+    default String getLabel() {
+        throw new UnsupportedOperationException();
+    }
+
 }
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
@@ -83,7 +83,8 @@ public interface Navigation extends ComponentExporter {
     }
 
     /**
-     * Retrieves the value of label for navigation.
+     * Retrieves the value of aria-label for navigation.
+     * Improves accessibility of component.
      *
      * @return the value of label for navigation
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
@@ -83,8 +83,7 @@ public interface Navigation extends ComponentExporter {
     }
 
     /**
-     * Retrieves the value of aria-label for navigation.
-     * Improves accessibility of component.
+     * Retrieves an accessible label that describes the navigation.
      *
      * @return the value of label for navigation
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
@@ -38,7 +38,7 @@ public interface Tabs extends Container {
     /**
      * Returns an accessible label that describes the tabs.
      *
-     * @return The value of label
+     * @return an accessible label for tabs
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
      */
     default String getLabel() {
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
@@ -36,8 +36,7 @@ public interface Tabs extends Container {
     }
     
     /**
-     * Returns the value of aria-label for tabs.
-     * Improves accessibility of component.
+     * Returns an accessible label that describes the tabs.
      *
      * @return The value of label
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
@@ -34,4 +34,14 @@ public interface Tabs extends Container {
     default String getActiveItem() {
         throw new UnsupportedOperationException();
     }
+    
+    /**
+     * Returns the value of label
+     *
+     * @return The value of label
+     * @since com.adobe.cq.wcm.core.components.models 12.9.0
+     */
+    default String getLabel() {
+        throw new UnsupportedOperationException();
+    }
 }
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
@@ -36,7 +36,8 @@ public interface Tabs extends Container {
     }
     
     /**
-     * Returns the value of label
+     * Returns the value of aria-label for tabs.
+     * Improves accessibility of component.
      *
      * @return The value of label
      * @since com.adobe.cq.wcm.core.components.models 12.9.0
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
@@ -34,7 +34,7 @@ public interface Tabs extends Container {
     default String getActiveItem() {
         throw new UnsupportedOperationException();
     }
-    
+
     /**
      * Returns an accessible label that describes the tabs.
      *
--- a/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/_cq_dialog/.content.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/_cq_dialog/.content.xml
@@ -149,10 +149,10 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
-                                            <carouselLabel
+                                            <label
                                                 jcr:primaryType="nt:unstructured"
                                                 sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
-                                                fieldDescription="Value of an aria-label attribute for carousel wrapper which describes content inside carousel."
+                                                fieldDescription="Value of an aria-label attribute for the carousel, which describes the carousel content."
                                                 fieldLabel="Label"
                                                 name="./label"
                                                 value=""/>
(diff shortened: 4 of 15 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
```diff
@@ -82,4 +82,14 @@ default String getExportedType() {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Retrieves the value of label for navigation.
```
richardhand: We should mention the accessibility intention of the label here

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
```diff
@@ -15,14 +15,19 @@
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
 <div data-sly-use.carousel="com.adobe.cq.wcm.core.components.models.Carousel"
      data-sly-use.templates="core/wcm/components/commons/v1/templates.html"
+     aria-label="${carousel.carouselLabel}"
+     aria-roledescription="carousel"
      class="cmp-carousel"
+     role="group"
```
richardhand: Would list the accessibility attributes after the css class, with the role first, something like:
```
class="cmp-carousel"
role="group"
aria-roledescription="carousel"
aria-label="${carousel.carouselLabel}"
```

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
```diff
@@ -15,14 +15,19 @@
...
      data-cmp-is="carousel"
      data-cmp-autoplay="${(wcmmode.edit || wcmmode.preview) ? '' : carousel.autoplay}"
      data-cmp-delay="${carousel.delay}"
      data-cmp-autopause-disabled="${carousel.autopauseDisabled}">
-    <div class="cmp-carousel__content" data-sly-test="${carousel.items && carousel.items.size > 0}" >
+    <div data-sly-test="${carousel.items && carousel.items.size > 0}"
+         class="cmp-carousel__content">
         <div data-sly-repeat.item="${carousel.items}"
              data-sly-resource="${item.name @ decorationTagName='div'}"
+             aria-label="${'Slide {0} of {1}' @ format=[itemList.count, carousel.items.size], i18n}"
```
richardhand: This is great :)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
```diff
@@ -76,4 +76,24 @@ default boolean getAutopauseDisabled() {
...
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Returns the value of label for indicators wrapper
+     *
+     * @return The value of label for indicators wrapper
+     * @since com.adobe.cq.wcm.core.components.models 12.9.0
+     */
+    default String getIndicatorsLabel() {
```
richardhand: Do we need a separate API function for this? Can't we just base the indicators off the main carousel label and prepend / append a default string:

Label: "top holiday destinations"
Indicators: "choose a slide for top holiday destinations"

.. or something similar. I think it would make the authoring experience simpler.

In the example they just demo a generic string, that could also work:
> The set of controls is grouped in a tablist element with an accessible name provided by the value of aria-label that identifies the purpose of the tabs, e.g., "Choose slide to display."

Same would apply to the Tabs component I suppose.

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
```diff
@@ -48,6 +48,15 @@
...
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
+                                            <label
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+                                                fieldDescription="Value of aria-label attribute for the navigation. Should be specified if more than one naviagtion is used on a page."
+                                                fieldLabel="Navigation Label"
+                                                name="./label"
+                                                value=""
+                                                emptyText="Enter value of label."
```
richardhand: I would remove this empty text, doesn't really provide much more info than the label.

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
```diff
@@ -48,6 +48,15 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
+                                            <label
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+                                                fieldDescription="Value of aria-label attribute for the navigation. Should be specified if more than one naviagtion is used on a page."
+                                                fieldLabel="Navigation Label"
```
richardhand: We can just call it "Label"

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
```diff
@@ -48,6 +48,15 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
+                                            <label
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+                                                fieldDescription="Value of aria-label attribute for the navigation. Should be specified if more than one naviagtion is used on a page."
```
richardhand: Would improve this a little:
"Value of an aria-label attribute for the navigation. Should be added there is more than one navigation on the page."

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
```diff
@@ -48,6 +48,15 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
+                                            <label
```
richardhand: This should at the minimum be the last field in the dialog as it's less important than the other options, but I would also consider putting it in a new tab as you have with the other components, even though it contains just one field.

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/_cq_dialog/.content.xml
```diff
@@ -72,6 +72,15 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
+                                            <label
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+                                                fieldDescription="Value of aria-label attribute, which describes purpose of the set of tabs."
+                                                fieldLabel="Tabs Label"
```
richardhand: Let's use just "Label", Tabs is inferred from the context

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/_cq_dialog/.content.xml
```diff
@@ -72,6 +72,15 @@
...
+                                                name="./label"
+                                                value=""
+                                                emptyText="Enter value of label."
```
richardhand: Same comment as with the Navigation, would remove this or provide a good default example.

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
```diff
@@ -48,6 +48,15 @@
...
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+                                                fieldDescription="Value of aria-label attribute for the navigation. Should be specified if more than one naviagtion is used on a page."
+                                                fieldLabel="Navigation Label"
+                                                name="./label"
+                                                value=""
+                                                emptyText="Enter value of label."
+                                                required="{Boolean}false"/>
```
richardhand: Not required is the default, would remove this property

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/_cq_dialog/.content.xml
```diff
@@ -72,6 +72,15 @@
...
+                                                required="{Boolean}false"/>
```
richardhand: Not required is the default, would remove this property.

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/tabs.html
```diff
@@ -18,17 +18,21 @@
...
+        <sly data-sly-test.isActive="${tab.name == tabs.activeItem}"/>
+        <li role="tab"
+            class="cmp-tabs__tab${isActive ? ' cmp-tabs__tab--active' : ''}"
+            tabindex="${isActive ? '0' : '-1'}"
+            data-cmp-hook-tabs="tab">${tab.title}</li>
     </ol>
     <div data-sly-repeat.item="${tabs.items}"
          data-sly-resource="${item.name @ decorationTagName='div'}"
          role="tabpanel"
+         tabindex="0"
```
richardhand: This was already added in https://github.com/adobe/aem-core-wcm-components/commit/29846f754886d6d0ed08eac3e2d080de590bd8dd

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/_cq_dialog/.content.xml
```diff
@@ -72,6 +72,15 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
+                                            <label
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+                                                fieldDescription="Value of aria-label attribute, which describes purpose of the set of tabs."
```
richardhand: Would slightly improve this label:
"Value of an aria-label attribute, which describes the purpose of the set of tabs."

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/_cq_dialog/.content.xml
```diff
@@ -72,6 +72,15 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
+                                            <label
```
richardhand: Same comment as in the Navigation component, this should be the last field in the dialog or in its own accessibility tab.

richardhand:
Thanks @jakublamprecht for the effort, added some comments. I will demo this change to the team and get any additional feedback to you.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
jakublamprecht: Would "Retrieves the value of aria-label for navigation. Improves accessibility of component" sound better?

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
jakublamprecht: I can see that buttons have those in this order: role -> class -> other stuff. Would you want me to change the buttons to have those attributes in the order you mentioned, or maybe make other things have the same order of attributes as those buttons (role before class)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
jakublamprecht: I'm not sure about the prepending/appending of default string - this would cause the aria-label of carousel to be duplicated and possibly read twice to a screen reader user and as far as I know we should avoid duplicating content from the accessibility point of view.

As far as the generic string goes it does sound good, but I'm wondering if there could be any case where author would want the label to be different.

> Same would apply to the Tabs component I suppose.

Sounds similar, but I think the case is a bit different. https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html here they suggest, that the label should describe the purpose of the set of tabs, so I guess a generic label for Tabs would not work as well as here, since the tabs in carousel have one clearly defined purpose.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
richardhand: We have different orderings per component right now and need a clean up task to make sure we're consistent, but I think that order I posted makes the most sense (they're the same order you would see when inspecting an element in the console). We should document it in our HTL guidelines. In full, I think it should be:

```
HTL directives
CSS classes
Role attribute
Aria attributes
Data attributes
```

As a first step it makes sense to make sure we're consistent in this file, then later take on the tidy up for other components.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
richardhand: > As far as the generic string goes it does sound good, but I'm wondering if there could be any case where author would want the label to be different.

I would go with the generic string for now as per the example. We can add the indicators later to the API if there seems to be a need for it.

For Tabs, that's right we can keep the author defined label.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
jakublamprecht: Changed all the Javadocs, let me know if the new ones are ok

richardhand: I probably wouldn't mention the `aria-label` attribute here, it's an implementation detail, you could equally use `aria-labelledby`, for example. Maybe: "Retrieves an accessible label that describes the navigation." is better?

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
jakublamprecht: Sounds good, changed the order of attributes to the order you mentioned

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
jakublamprecht: Removed empty text

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
jakublamprecht: Solved

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
jakublamprecht: Updatede the description

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
jakublamprecht: Moved the label to a separate tab

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/_cq_dialog/.content.xml
jakublamprecht: Solved

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/_cq_dialog/.content.xml
jakublamprecht: Solved by removing

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
jakublamprecht: Removed

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/_cq_dialog/.content.xml
jakublamprecht: Removed

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/tabs.html
jakublamprecht: I'm not sure either, this branch was made using only the development branch

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/_cq_dialog/.content.xml
jakublamprecht: Moved to a separate accessibility tab

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
```diff
@@ -92,6 +92,34 @@
...
+                                margin="{Boolean}true">
+                                <items jcr:primaryType="nt:unstructured">
+                                    <column
+                                        jcr:primaryType="nt:unstructured"
+                                        sling:resourceType="granite/ui/components/coral/foundation/container">
+                                        <items jcr:primaryType="nt:unstructured">
+                                            <label
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+                                                fieldDescription="Value of an aria-label attribute for the navigation. Should be added if there is more than one naviagtion on the page."
```
richardhand: spelling: naviagtion -> navigation

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
jakublamprecht: Done

content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/tabs.html
richardhand: Weird. And if I do a local diff against development, I don't see these changes.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
```diff
@@ -34,4 +34,14 @@
     default String getActiveItem() {
         throw new UnsupportedOperationException();
     }
+
```
richardhand: picky one, but unnecessary spaces are added here.

content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/_cq_dialog/.content.xml
jakublamprecht: Fixed

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Tabs.java
jakublamprecht: Fixed

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/_cq_dialog/.content.xml
```diff
@@ -134,6 +134,34 @@
...
+                            <columns
+                                jcr:primaryType="nt:unstructured"
+                                sling:resourceType="granite/ui/components/coral/foundation/fixedcolumns"
+                                margin="{Boolean}true">
+                                <items jcr:primaryType="nt:unstructured">
+                                    <column
+                                        jcr:primaryType="nt:unstructured"
+                                        sling:resourceType="granite/ui/components/coral/foundation/container">
+                                        <items jcr:primaryType="nt:unstructured">
+                                            <carouselLabel
```
richardhand: `carouselLabel` -> `label`

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/_cq_dialog/.content.xml
```diff
@@ -134,6 +134,34 @@
...
+                                                jcr:primaryType="nt:unstructured"
+                                                sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+                                                fieldDescription="Value of an aria-label attribute for carousel wrapper which describes content inside carousel."
```
richardhand: we should use full sentences for the descriptions:
"Value of an aria-label attribute for the carousel, which describes the carousel content."

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
```diff
@@ -76,4 +76,14 @@ default boolean getAutopauseDisabled() {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Returns an accessible label that describes the carousel.
+     *
+     * @return The value of label for Carousel component
```
richardhand: We should use the same consistent return descriptions for these:
• The value of label for Carousel component
• the value of label for navigation
• The value of label

Would prefer:
• an accessible label for the carousel
• an accessible label for the navigation
• an accessible label for the tabs

richardhand:
@jakublamprecht - this is looking in good shape now, just had a few final comments.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/_cq_dialog/.content.xml
jakublamprecht: Done

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/_cq_dialog/.content.xml
jakublamprecht: This sounds better, changed the description

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Carousel.java
jakublamprecht: Done

richardhand:
Thanks @jakublamprecht!

PR #689, opened 2019-08-09
[Page] Make head more customizable #685

### LINKED ISSUE #685: [Page] Make head more customizable
## Feature Request

When trying to override the title, description and keywords, we have to override the whole https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/head.html file. We would however like to override only the specific parts that we want to customize, so we can still leverage all other code OOTB.

More fine-grained control in overriding parts of the header/footer/...

For example:
* Make the page component, instead of putting the title etc. in head.html, include page-metadata.html (that includes title etc)
* Do the same for other things
* Then we can, in our page component, override page-metadata.html and do our custom title etc over there.

**Are there alternatives?**
* Overriding head.html, but this will make us do that only to customize 3 lines + we will have to maintain that override to be in-sync
* Writing a transformer to remove title, description and keywords, and then use customheaderlibs.html to add our custom title logic -- but this feels so very wrong I cannot even explain it

**Documentation**
Technical

henrykuijpers:
| Fixed Issues?            | Fixes #685 

Added various extension points in the head-logic of the page component. It is now possible to:
* Override the page title (title tag)
* Override the page description (keywords & description)
* Add custom metadata to the head-tag
*

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/head.custommetadata.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/head.custommetadata.html
@@ -1,0 +1,15 @@
+<!--/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  ~ Copyright 2016 Adobe Systems Incorporated
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


henrykuijpers:
I disagree here: the only way for people to influence the outcome of the logic that modified, previously, was to override head.html. 

If anyone did this in their projects, while using the page v2 component, they do still get the same behavior when they install this new version. 

We're only introducing new extension points when the head.html is left intact.

gabrielwalt:
Closing this as there doesn't seem to be a consensus on how the extension points should work.

PR #32, opened 2017-06-14
Restrict image component to rasterized images

kwin:
This closes #31

PR #705, opened 2019-08-21
Remove image/webp media type - 

stefanseifert:
it's supported neither by AEM Assets nor by the Core Component Image component.
i think this was introduced (probably accidentally) in #32 and copied from there to the other places.

| Fixed Issues?            | 
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Documentation Provided   | 


jckautzmann:
I think it's safe to remove the `image/webp` media type as it's not a supported type in AEM:
https://helpx.adobe.com/experience-manager/6-5/assets/using/assets-formats.html

This type is only mentioned in AEM in the context of Dynamic Media which is not supported with the Core Components:
https://helpx.adobe.com/experience-manager/6-5/assets/using/imaging-faq.html

PR #706, opened 2019-08-21
Ensure download component has a valid asset reference or file upload

stefanseifert:
evaluate if download has content by checking the "filename" property - it contains only a value if an asset reference or upload is present and valid
eliminate the checks for title, description and action text - a download component without any of these is fine, but any of these without a valid download does not make sense

| Fixed Issues?            | 
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Documentation Provided   | 
| Any Dependency Changes?  |


bpauli:
@stefanseifert Thank you for this fix!

PR #708, opened 2019-08-24
Migrate unit test to aem mocks 2.7

### LINKED ISSUE #682: Switch to AEM Mocks 2.7.0 with Content Policy Support
we recently added content policy support to [AEM Mocks](https://wcm.io/testing/aem-mock/) - testable with version 2.6.1-SNAPSHOT. as you are using both AEM Mocks and (a homegrown support for) content policies in the unit tests for the core components you may want to give it a try to simplify your unit tests.

usage is really simple - if you want to provide a component with a content policy for testing, just add a mapping for its resource type with a code line like this:

```
// create a content policy with mapping for resource type
context.contentPolicyMapping("app1/componenty/component1",
    "prop1", "value1",
    "prop2", 123);
```

it is then available either using the ContentPolicy APIs, or when injecting a "currentStyle" in a sling model within the unit test.

release is planned in 1-2 weeks.

bpauli:
fixes #682


vladbailescu:
@stefanseifert , after switching to the new AEM Mocks we had issues with our unit tests on CircleCI: the forked VMs crashed for some tests doe to AEM Mocks out-of-memory. I bumped the amount of heap available for surefire forks and that solved the issue but I'm wondering if you have seen this happen before.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v1/TitleImplTest.java
```diff
@@ -1,5 +1,5 @@
 /*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
- ~ Copyright 2017 Adobe Systems Incorporated
+ ~ Copyright 2019 Adobe Systems Incorporated
```
richardhand: @bpauli why bump the copyright year for some classes and not others, as I understand the initial year should remain? I also notice you update the headers for some and not others to remove "Adobe Systems Incorporated", would probably be cleaner to update them all at once in a dedicated task.

stefanseifert:
> @stefanseifert , after switching to the new AEM Mocks we had issues with our unit tests on CircleCI: the forked VMs crashed for some tests doe to AEM Mocks out-of-memory. I bumped the amount of heap available for surefire forks and that solved the issue but I'm wondering if you have seen this happen before.

no, i'm not aware of any problems like this. if you find a configuration to reproduce it locally (e.g. using a certain set of JVM parameters) please give me a notice.

PR #723, opened 2019-09-06
[Teaser] Title changes reset after adding a link

### LINKED ISSUE #719: [Teaser] Title changes reset after adding a link
## Bug Report

**Current Behavior**
When I configure a teaser-component and I edit the title (the checkbox "Get title from linked page" is not activated), it is reseted after I choose a link:

1. Bildschirmfoto 2019-08-30 um 15 20 49
2. Bildschirmfoto 2019-08-30 um 15 21 01
3. Bildschirmfoto 2019-08-30 um 15 22 02
4. Bildschirmfoto 2019-08-30 um 15 22 12
5. Bildschirmfoto 2019-08-30 um 15 20 49

It doesn't matter if I use the picker or just type the link into the field.

**Expected behavior/code**
Changes to the link should not reset the title.

**Environment**
- AEM version: 6.4.2.0
- Core Components: 2.5.0

bpauli:
- update the data attribute for the previous text value after text get changed by the author

fixes #719

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/clientlibs/editor/js/teaser.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/clientlibs/editor/js/teaser.js
@@ -47,16 +47,6 @@
                 retrievePageInfo($dialogContent);
             });
 
-            var $titleTextfield = $dialogContent.find(titleTextfieldSelector);
-            $titleTextfield.on("change", function() {
-                titleTuple.updatePreviousTextValue();
-            });
-
-            var $descriptionTextfield = $dialogContent.find(descriptionTextfieldSelector);
-            $descriptionTextfield.on("change", function() {
-                descriptionTuple.updatePreviousTextValue();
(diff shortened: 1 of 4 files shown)


content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/clientlibs/editor/js/teaser.js
```diff
@@ -47,6 +47,16 @@
                 retrievePageInfo($dialogContent);
             });
 
+            var $titleTextfield = $dialogContent.find(titleTextfieldSelector);
+            $titleTextfield.on("change", function() {
```
richardhand: @bpauli - shouldn't we add this change handling internally to the `checkboxTextfieldTuple`? rather than exposing `CheckboxTextfieldTuple.prototype.updatePreviousTextValue`?

Would be good to also make sure we've covered this case with a UI test and check that the other component dialogs that use the `CheckboxTextfieldTuple` are still working correctly (download and image).

PR #724, opened 2019-09-06
[Download] Improved memory usage

### LINKED ISSUE #700: DownloadServlet loads complete assets into heap
## Feature Request

The DownloadServlet is loading the complete asset into memory (a byte[] array) before it is written to the OutputStream of the response.
This allocation requires a huge memory block, which can lead to massive GC activity with potential OutOfMemory situations.

The DownloadServlet should be improved so it streams the asset directly from the repository to the response, without allocating large portions of memory.

bpauli:
fixes #700

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/DownloadServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/DownloadServlet.java
@@ -15,6 +15,8 @@
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.internal.servlets;
 
+import static com.adobe.cq.wcm.core.components.internal.servlets.DownloadServlet.SELECTOR;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.text.ParseException;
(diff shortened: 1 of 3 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/DownloadServlet.java
```diff
@@ -15,6 +15,8 @@
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.internal.servlets;
 
+import static com.adobe.cq.wcm.core.components.internal.servlets.DownloadServlet.SELECTOR;
```
jckautzmann: why is this needed?

bpauli: Was a shortcut for the selector in the `Component` annotation. But right, maybe it's mor clear to remove the the static import. done in https://github.com/adobe/aem-core-wcm-components/pull/724/commits/87e08a6658f8f16930ae17626490848f8f94dd49

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/DownloadServlet.java
```diff
@@ -72,95 +79,136 @@ protected void doGet(@Nonnull SlingHttpServletRequest request, @Nonnull SlingHtt
...
         }
-        return bytes;
     }
 
-    private byte[] getByteArray(Resource fileResource) throws IOException {
-        byte[] bytes = new byte[]{};
-        InputStream stream = fileResource.getValueMap().get(JcrConstants.JCR_DATA, InputStream.class);
-        if (stream != null) {
+    /**
+     * Determine the size of a binary resource just by using the JCR API; this hopefully avoids to read
```
jckautzmann: use descriptive tense: Determine -> Determines

Issue #713, opened 2019-08-28
[Link Handling] Make link references consistent and flexible


stefanseifert:
## Feature Request

(this is a follow-up from [this dicussion](https://groups.google.com/forum/#!msg/aem-core-components-dev/st_31IWmyhs/fT5ZCXT3DAAJ) in the mailing list)

currently, the storing of link reference data in properties and providing access to link references via the model interfaces is very inconsistent, see [1] for an overview. additionally, it's inflexible as it only allows to store a link url, and no further properties e.g. if the link should be opened in a new tab/window (_blank) or not. PR #625 is an attempt to add this to some components, but in my pov this is not flexible enough.

my proposal is:

* a) introduce an interface for link target reference
* b) all model interface that have a link (including ListItem) extend it
* c) include further properties that may come with the link like: target, SEO, tracking properties in a generic way
* d) change the HTL scripts to generate links always based in the methods of this new interface
* e) create a shared piece of TouchUI dialog definition (e.g. a field set) that is used in all edit dialogs to edit the target link and always looks the same

such an interface may look like this:

```java
interface LinkMixin {
  boolean isLinkValid();
  String getLinkURL();
  Map<String, String> getLinkAttributes();
}
```

the map returns key/value pairs like target=_blank, rel=nofollow etc. it should also contain the main href=xxx attribute. the separate getLinkURL method is only for components that for some reason do not generate an anchor tag but want to do other things with the LinkURL (e.g. putting it in a data attribute).

the HTL would look like:

```html
<a data-sly-unwrap="${!obj.linkValid }" data-sly-attribute="${obj.linkAttributes}">...</a>
```

the inconsistent property names cannot be cleaned up without breaking existing content, so we stick with them. for the interface which use link reference method not in compliance with the new method names a default implementation can be added that calls the new names.

[1] Components and their link properties/methods in Core Components 2.5.0

```
Component     | Property Name  | API Method
--------------+----------------+------------
Button        | link           | getLink
Image         | linkURL        | getLink
Teaser        | linkURL        | getLinkURL
Teaser/Action | link           | getURL
Title         | linkURL        | getLinkURL
ListItem      |                | getURL
```

kwin:
Instead of relying on the HTML specific anchor attributes it would be nice to have some kind of abstraction for common attributes to be completely view agnostic and keep the model/view separation (i.e. dedicated getters per aspect). The problem is though that the list is probably not finite and may grow over time.

[...]

Issue #725, opened 2019-09-08
[Accessibility] Support setting Anchor target attribute for all editable links


stefanseifert:
currently several components allow to define a link target (internal page or external URL) in their edit dialogs, but it's not possible to define a "target" attribute for the anchor tag e.g. to open the link in a new tab/window.

the following components are affected:
* Button
* Image
* Teaser and Teaser Actions
* Title

proposal for the implementation:
* add a common/shared dialog fragment currently consisting of the link URL page field and a select field to select from "self" (default) or "new tab/window" option.
* this shared dialog fragment is included in the edit dialogs of the mentioned components, preferably in a new tab to separate it from the other edit options, and it's possible that new features are added to it in the future or by downstream libraries/applications
* the JS logic that currently hides/displays the link URL field currently in some dialogs needs to be adapted as well
* the window target needs to be stored in the property already added in #713 - then is should work without any further markup changes

this makes #625 obsolete

stefanseifert:
i will come up with a PR for this in the next time.

stefanseifert:
with the new features from #713 underneath this should be easy. what is the exact property you expect to be set on the anchor tag?

jakublamprecht:
I think that the title attribute would be the best way to do this, as it is read additionally, right after the text defined within the link.

richardhand:
As I understand, the `title` attribute is no silver bullet solution here, due to it only being surfaced on pointing devices for visual users and having inconsistent screen reader support (although I don't have a full idea of how that support looks).

The HTML spec [0] discourages its use for this purpose:

> Relying on the title attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g. requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet). 

[0] - https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute

tjameswhite:
@jakublamprecht is correct, there needs to be some mechanism to let assistive technology users know the link opens in a new window/tab. I'm implementing a solution for our custom components and was trying to figure out how to get it on Core as well. 

I'm currently adding `aria-describedby="new-window"` to the `<a>` which then references this block hidden in the page template: 
```
<div hidden>
 <span id="new-window">${'opens in new window' @ i18n}</span>
</div> 
```

[...]

PR #726, opened 2019-09-08
[Link Handling] Make link references consistent and flexible #713

stefanseifert:
* introduce LinkMixin interface for consistent representation of link target information in all model APIs
* make sure default implementations for new API methods fall back to the existing methods to not break anything
* introduce a simple (and internal) LinkHandler implementation to resolve all types of links in all implementations, deprecate Utils methods for the same purpose
* update markup scripts to use linkHtmlAttributes for generating the anchor href (and more attributes in the future), and linkValid to check if a valid link is present
* prepare support for link target in the model implementations

fixes  #713


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ButtonImpl.java
```diff
@@ -46,43 +51,57 @@
...
-
     @ValueMapValue(optional = true)
     private String icon;
 
     @ValueMapValue(optional = true)
     protected String accessibilityLabel;
 
+    @Self
+    private LinkHandler linkHandler;
+    private Link link;
```
kwin: What about a dedicated injector for LinkHandler?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/Link.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public @Nullable Map<String, String> getLinkHtmlAttributes() {
+        return linkHtmlAttributes;
+    }
+
+    /**
+     * @return Target page if the link URL pointed to an internal page.
+     */
+    public Page getTargetPage() {
```
kwin: Nullable?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ButtonImpl.java
stefanseifert: we might think about this later, but it's not as easy as always doing the same for the same model:
* some models have different properties to read the link url from (for historical reasons)
* some models do have additional links which need to be processed with the link handler (e.g. teaser action links, image map area links, links in navigation-like models etc.)

so i currently would prefer to keep it like this until reocurring patterns emerge. after all, it's still an internal API and not yet published, so we can change it at any time.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/Link.java
stefanseifert: +1, forgot it. i'll wait if other review comments are posted and will update it then.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ButtonImpl.java
kwin: I was thinking about an annotation with fields which allow to override the node and the properties from which to read, but I agree, we can do this later on...

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/Link.java
joerghoh: Why not use Optional<Page> to express this in the code as well?

stefanseifert: thought about this as well and even started to use it for this method but refactored back as it's still unusual to be used in the sling/AEM/core component APIs - and it will not work well in HTL (although this Link class is currently not used in HTL at all).
so i'm completly open to use Optional in this case, but it would be the first place in all the existing APIs, and i thought that was not worth it.

joerghoh: HTL support is a good topic, indeed.

But being the first one ... At some point we have to use it. Because I see way too much code with NPEs and people not using linting tools and also not reading JavaDoc. And using Optional forces them to reason about it and makes it really implicit that nothing can be returned (not `null`).

jckautzmann:
Instead of the mixin mechanism, I would prefer that each Model that has a link has a method to return it:
`default Link getLink();`
Accessing a link property in the HTL would be as: `${button.link.URL}`

To achieve this I would:
- rename `LinkMixin` -> `Link` and rename its methods:
`isLinkValid` -> `isValid`
`getLinkURL` -> `getURL`
`getLinkHtmlAttributes` -> `getHtmlAttributes`
- rename `Link` -> `LinkImpl`

WDYT?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -0,0 +1,126 @@
...
+     * Build a link pointing to the given target page.
+     * @param page Target page
+     * @return Link may be invalid, but is never null
+     */
+    @SuppressWarnings("deprecation")
+    public @NotNull Link getLink(@Nullable Page page) {
+        if (page == null) {
+            return getInvalid();
+        }
+        String linkURL = Utils.getURL(request, page);
```
jckautzmann: I wouldn't use a deprecated API. I would prefer to introduce a private method here to do the job.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -0,0 +1,126 @@
...
+     * @return Invalid link, never null
+     */
+    public @NotNull Link getInvalid() {
+        return new Link(null, null, null);
+    }
+
+    @SuppressWarnings("deprecation")
+    private String validateAndResolveLinkURL(String linkURL) {
+        if (!StringUtils.isEmpty(linkURL)) {
+            return Utils.getURL(request, pageManager, linkURL);
```
jckautzmann: same as above.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
stefanseifert: well, i just deprecated the used methods it in the same commit.
i've not removed the Utils.getURL methods completely, but that would be one of the things i would do next and include the code in the link handler itself. just wanted not to throw everything away at once to make reviewing the PRs easier.

and Utils is only an internal class, not an external.

if you prefer i can get rid of the deprecated methods rightaway in this PR as well.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -0,0 +1,126 @@
...
+        if (page == null) {
+            return getInvalid();
+        }
+        String linkURL = Utils.getURL(request, page);
+        return new Link(linkURL, null, page);
+    }
+
+    /**
+     * Build a link pointing to the given target page.
+     * @param page Target page
```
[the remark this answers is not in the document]
stefanseifert: what's wrong?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkNameConstants.java
```diff
@@ -0,0 +1,49 @@
...
+     * Property name for storing link target.
+     */
+    public static final String PN_LINK_TARGET = "linkTarget";
+
+    /**
+     * List of allowed/supported values for link target.
+     * <code>_self</code> is used in the edit dialog but not listed as allowed here as we do not
+     * want to render a target attribute at all when <code>_self</code> is selected.
+     */
+    public static final Set<String> VALID_LINK_TARGETS = ImmutableSet.of("_blank");
```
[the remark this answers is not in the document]
stefanseifert: yes, we can add it as well. at least here in this validation code.
not sure if we want to offer them as options in the edit dialogs later as well (in #725), or if this puzzles the users.

stefanseifert:
> Instead of the mixin mechanism, I would prefer that each Model that has a link has a method to return it:
default Link getLink();
Accessing a link property in the HTL would be as: ${button.link.URL}

yes, good idea (and same API pattern wcm.io Link Handler is currently using).
we just cannot use `getLink` method in all model APIs, as some of them already have an `getLink` method returning the Link URL. but it would be useful for having a method with the same name in all model interfaces.

this enforcing was also the reason i introduced the mixin interface in the first place - this enforces every model used the same method name. on the other hand it may be that in the future some model has two link objects and not only one.

so, what should we do?
1. introduce a `getLink` method for all models where it is possible, use a different name for two models (Button and Image) that have this method already?
2. use a different method name like `getLinkObject` in all models (which is much less nice in HTL)
3. stick with the current mixin interface concept
4. use both concept to introduce a method that returns a link interface and a mixin to enforces the naming of this method in all model interfaces?

(p.s. i was hoping for this principle API discussion to be done in the mailing list a few weeks ago, there it is much easier to switch concepts than to rework the whole PR...)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkNameConstants.java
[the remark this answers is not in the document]
jckautzmann: we might want to have all the possible options selectable in the policy dialog. The edit dialog would then only display the selected options.
@gabrielwalt Could you please chime in to discuss the user experience?

jckautzmann:
+1 for option#1

If a model has a need in the future for multiple links, we'll still be able to introduce a new API like e.g. `Set<Link> getLinks`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
jckautzmann: +1

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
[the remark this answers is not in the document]
bpauli: My IntelliJ is complaining about the JavaDoc.
As the parameter name is `target` the javadoc should be `@param target Page target` or the parameter name needs to be renamed (e.g.

stefanseifert:
please cross-check the new API approach introducing a `Link` instead of a `LinkMixin` interface as described above.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Heading.java
```diff
@@ -0,0 +1,48 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2017 Adobe
```
jckautzmann: Should be 2019.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -0,0 +1,157 @@
...
+@Model(adaptables=SlingHttpServletRequest.class)
+public class LinkHandler {
+
+    @Self
+    private SlingHttpServletRequest request;
+    @ScriptVariable
+    private PageManager pageManager;
+
+    /**
+     * Resolve a link from the properties of the given resource.
```
jckautzmann: We usually use the present tense (not the imperative form) in the javadocs to describe the methods:
e.g. Resolve -> Resolves

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public @Nullable Page getTargetPage() {
+        return targetPage;
+    }
+
+    private static Map<String, String> buildHtmlAttributes(String linkURL, String linkTarget) {
+        if (linkURL == null) {
+            return null;
```
jckautzmann: I'd return an empty Map instead.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
```diff
@@ -0,0 +1,94 @@
...
+        return url != null;
+    }
+
+    @Override
+    public @Nullable String getURL() {
+        return url;
+    }
+
+    @Override
+    public @Nullable Map<String, String> getHtmlAttributes() {
```
jckautzmann: I'd make this method @NotNull instead.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkNameConstants.java
```diff
@@ -0,0 +1,49 @@
...
+package com.adobe.cq.wcm.core.components.internal.link;
+
+import java.util.Set;
+
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * Constants for link handling.
+ */
+public final class LinkNameConstants {
```
jckautzmann: I'd rename the class `LinkConstants`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Button.java
```diff
@@ -38,12 +40,26 @@ default String getText() {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Returns the button link.
+     *
+     * @return the button link
+     * @since com.adobe.cq.wcm.core.components.models 12.10.0
+     */
+    @JsonIgnore  // avoid duplicate URL in JSON, keep old property for backward compatibility
```
jckautzmann: I would display the new property in the JSON as it might contain some information that was not available before (e.g. the link target).

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Button.java
```diff
@@ -38,12 +40,26 @@ default String getText() {
...
+    default @NotNull Link getButtonLink() {
+        // fallback to old method name for backwards compatibility
+        return new LinkImpl(getLink());
```
jckautzmann: I would throw instead an `UnsupportedOperationException`. As it is a new method, backwards compatibility does not apply.
This applies to all the models below.

jckautzmann:
Thx @stefanseifert for the updated PR! I added a few comments inline.

I would also suggest to make `Link`, `LinkHandler` and `LinkConstants` more visible and put them in a new `commons` package as follows:
```
com.adobe.cq.wcm.core.components.commons.Link
com.adobe.cq.wcm.core.components.commons.LinkHandler
com.adobe.cq.wcm.core.components.commons.LinkConstants
```
WDYT?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Button.java
stefanseifert: i thought about this as well - this would result in duplicate information in the JSON, the URL would be listed even thrice.

simplified example:
```javascript
{
  "link": "/content/page1.html":
   "buttonLink": {
      "valid": true,
      "url": "/content/page1.html",
      "htmlAttributes": {
          "href": "/content/page1.html",
          "target": "_blank"
      }
   }
}
```

or for an invalid link:
```javascript
{
   "buttonLink": {
      "valid": false
   }
}
```

is this intended?
is JSON expected to be "view-agnostic"? - in this case at least the "htmlAttributes" map should be hidden, and the information of the target attribute as well.

stefanseifert:
> I would also suggest to make Link, LinkHandler and LinkConstants more visible and put them in a new commons package as follows...

you mean exporting them as part of the core components API? i think this is too early, and the LinkHandler API and implementation is far too simplified and partially naive as it is now. is suits as a replacement for the existing logic so it is at one single place.

but we may think to but the Link interface and probably LinkConstants class into a different package where later the LinkHandler would be problably exported as well when it's ready. but that should be a separate issue and would involve more API discussions.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Button.java
stefanseifert: when a user has overlayed the sling model but not the markup of a model it works out-of-the box with this design.

if we just throw unsupported all models will break until this new method is implemented as well.

i absolutely dislike this "throw unsupportedoperationexception" pattern, it makes it much more complicated for users who overlay/extend part of the models doing their job, as they are not "notified" by the compiler/in their IDE of new interface additions. but this would be a separate topic and separate issue.

for this PR i think it is a great benefit for the users gradually upgrading from core component version to another, and does not harm anything.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Button.java
jckautzmann: you're right. Given that the link is an HTML link and not a generic one, we might want to hide the HTML properties in the JSON. But I would hide it in the Model implementation, not in the Model itself.

jckautzmann:
>you mean exporting them as part of the core components API?
yes. `Link` is currently in the publicly available `com.adobe.cq.wcm.core.components.models` package, which is weird as it is not a Model. `LinkHandler` can be injected in models, so I consider it already publicly available. Making `LinkConstants` available would then also make sense to me.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Button.java
jckautzmann: > if we just throw unsupported all models will break until this new method is implemented as well.

What do you mean with "break"? Do you mean that it won't build? Adding a new method to a model should not break anything. I would not return a default value in the Model. If needed I would do it in the implementing class.

stefanseifert: break example:
* we release a new core component version with this PR that introduces the new get*Link methods and updated markup that references it
* assume a customer projects that have added their own implementation of the sling model of a component using the extender pattern, but did not overlay the markup file.
* when the customer deploys the latest version of the core components he is not aware of this new method as the build does not break
* but the markup is broken, as it references the new get*Link method, which just leads to an "unsupportedoperationexception" unless the customer detects the new method and adds an delegation implementation to this own model.
* he can only detect this new method by looking very closely at the diffs between the two core component releases as in the build and IDE everything seems fine because a default implementation in the impl exists (which if it just throws unsupportedoperationexception is a really unhelpful default implementation)

with my proposal of default implementation the use case described is covered and all runs as expected without further changes.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Button.java
stefanseifert: and to confirm: you want to appear the link URL twice the JSON - one with the old property name (for backwards compatibility), and second with the "link" or "buttonLink" object?
and in case of invalid a link object appears with "valid: false"?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Button.java
jckautzmann: Another option to not break backwards compatibility is to create a new version of the  components affected by this change.
I'm not keen with defining a default value in a new method of the Model that relies on a deprecated method.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Button.java
jckautzmann: @gabrielwalt has some ideas about the JSON rendering: I let him chime in the discussion.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Button.java
stefanseifert: creating a new version of all link-related components would be totally ok for me and a very clean solution - but it did not happen very often until so, so i'm not sure when it's the right time for your team to create a new version (and probably have more maintenance effors with maintaining bugfixes etc. for multiple component versions).

please decide in your team whether you want to create new model version for all link-related components or not. if not i strongly recommend this approach as it's a very simple but helpful solution for the users of this project.

gabrielwalt:
Hi @stefanseifert,

I wanted to let you know that we're currently busy with working on a Data Layer and will likely not be able to make this contribution progress before end of October. I'm really sorry that we lack the capacity to handle such valuable pull requests in a more timely manner. It is however among the top issues in our backlog, so that we can pick it up as soon as we'll have some slack again.

My priority for the JSON output is to have it as lean as possible, so avoid it to contain duplicate values (so avoid any duplicate `link` and `url` and `htmlAttributes.href`). I also kind of dislike having an `htmlAttributes` property in the JSON, which should be html-agnostic. Finally, do we really need a `valid` property (I know it's not new), can we not instead return no link value when it's not valid?

I'd prefer something like that where the `url` is kept separate from the `htmlAttributes`:
```html
<a href="${obj.url}" data-sly-attributes="${obj.htmlAttributes}" data-sly-unwrap="${!obj.url}">...</a>
```
and where the `htmlAttributes` and `valid` don't show up in the JSON:
```javascript
{
    ...
    "url": "/content/page1.html"
    ...
}
```

dabe5651:
Great contribution, before my project goes live in the next few weeks I will either need to overlay core components with temporary throwaway code or be lucky enough that this gets merged in.

richardhand:
@stefanseifert, I've based this against the new `3.0.0` branch as per the discussion in https://groups.google.com/forum/#!topic/aem-core-components-dev/st_31IWmyhs. We should be able to merge there once the discussed updates are available and reviewed.

stefanseifert:
i've created a new PR #862 that replaces this PR and incorporated the architecture changes discussed in https://groups.google.com/forum/#!msg/aem-core-components-dev/st_31IWmyhs/fT5ZCXT3DAAJ

richardhand:
Thanks Stefan!

PR #729, opened 2019-09-09
Allow empty content fragment in JSON

### LINKED ISSUE #720: To allow empty content fragment in JSON
## Feature Request

Problem Statement:
When an empty content fragment is authored in a page and we try to access model.json of same page, Following error message is thrown "There is no content fragment available to delegate to".

Affected use case:
In our environment, we create editable templates and add content fragments as a default component in our template structure.  Sometimes we need to remove the content fragment reference and leave it empty till we create new content fragments. This in-between period, we need that empty content fragment to be there and we need other components JSON value as well. So in order to achieve this, we need the above feature.

1) Remove the "throw new IllegalStateException("There is no content fragment available to delegate to");" from ContentFragmentImpl.java.
2) Do Null check to all getter methods in "ContentFragmentImpl.java" class to return type and outside null check, block return an empty value.

Sample Output in JSON will look like below:
contentfragment: {
   description: " ",
   title: " ",
   model: " ",
   :items: { },
       :itemsOrder: [ ],
       elementsOrder: [ ],
    :type: "my_project/components/content/contentfragment"
}

**Are there alternatives?**
No 

**Documentation**
Maybe a screenshot or design?

bpauli:
- return an empty content fragment in case the path to the content fragment is empty or not valid.

fixes #720

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
@@ -35,8 +35,6 @@ import org.apache.sling.models.annotations.injectorspecific.ValueMapValue;
 import org.apache.sling.models.factory.ModelFactory;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.adobe.cq.dam.cfm.content.FragmentRenderService;
 import com.adobe.cq.dam.cfm.converter.ContentTypeConverter;
@@ -60,8 +58,6 @@ import com.adobe.cq.wcm.core.components.models.contentfragment.DAMContentFragmen
 @Exporter(name = ExporterConstants.SLING_MODEL_EXPORTER_NAME,
           extensions = ExporterConstants.SLING_MODEL_EXTENSION)
 public class ContentFragmentImpl implements ContentFragment {
-
-    private static final Logger LOG = LoggerFactory.getLogger(ContentFragmentImpl.class);
 
     /**
      * The resource type of the component associated with this Sling model.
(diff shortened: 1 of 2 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
```diff
@@ -85,16 +86,20 @@
...
+    @ValueMapValue(name = ContentFragment.PN_VARIATION_NAME,
+                   injectionStrategy = InjectionStrategy.OPTIONAL)
     private String variationName;
 
-    @ValueMapValue(name = ContentFragment.PN_DISPLAY_MODE, injectionStrategy = InjectionStrategy.OPTIONAL)
+    @ValueMapValue(name = ContentFragment.PN_DISPLAY_MODE,
+                   injectionStrategy = InjectionStrategy.OPTIONAL)
     private String displayMode;
 
     private DAMContentFragment damContentFragment;
```
jckautzmann: we should probably also return an empty CF in line 108 and remove the error log

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
```diff
@@ -264,4 +216,45 @@ private DAMContentFragment getDAMContentFragment() {
         // split into paragraphs
         return content.split("(?=(<p>|<h1>|<h2>|<h3>|<h4>|<h5>|<h6>))");
     }
+
+
+    /**
+     * Empty placeholder content fragment.
+     */
+    public static class EmptyContentFragment implements DAMContentFragment {
```
jckautzmann: I'd make the class `private`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
```diff
@@ -146,31 +137,31 @@ public String getGridResourceType() {
...
     public List<DAMContentElement> getElements() {
-        return getDAMContentFragment().getElements();
+        return damContentFragment.getElements();
     }
 
     @Nullable
     @Override
     public List<Resource> getAssociatedContent() {
-        return getDAMContentFragment().getAssociatedContent();
+        return damContentFragment.getAssociatedContent();
```
jckautzmann: I'd also replace `getDAMContentFragment()` with `damContentFragment` in the `getParagraphs()` method.

PR #748, opened 2019-09-19
[Download][Accessibility] Duplicate links, can't add information to link text #737

### LINKED ISSUE #737: [Download][Accessibility] Duplicate links, can't add additional information to link text
## Bug Report

**Current Behavior**
Currently the Download component has a few accessibility issues which makes it not AA compliant:
- both title and the action button link to the same destination which causes duplication and redundancy of links. Either title should not be a link, or the whole card should be treated as a link.
- there's no possibility to add additional information about the link to the action link. More often than not designs will show just a single 'download' word, but for screen reader users, the link should inform the user that this link will cause a resource to be downloaded. The user should also be informed what is being downloaded and what's the size and format of the file.

I provided a few possible approaches to this problem that were mention in WCAG guidelines in the "possible solutions" section. Which one do you think would fit the best? If the component was to be AAA compliant, seems like wrapping the whole component with a link would be the best option.

Resources:

https://www.sitepoint.com/15-rules-making-accessible-links/ - point 6., 7.
https://accessabilly.com/proposal-for-a-more-accessible-download-link/
AA: https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-refs.html
AAA: https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-link.html

**Expected behavior/code**
Download component is compliant with AA accessibility requirements.

**Environment**
- Core Components version (e.g. 2.6.0)

**Possible Solution**
For the first issue:
Removing the link from title or treating the whole "download" component as a link.

For the second issue:
- Wrapping the whole download component with a link (but beware of the length of content inside the link, since users can't skip the text of the link).
- Adding aria-label
- Adding title attribute to the link
- Adding additional text to the link which will be visually hidden, but still available for screen readers.
- using aria-describedby to provide information about filename, filesize and extension by definition list above download link

jakublamprecht:
| Fixed Issues?            | Fixes #737 
| Patch: Bug Fix?          | Enhancement
| Minor: New Feature?      |
| Major: Breaking Change?  | Markup changes
| Documentation Provided   | Changed existing one
| Any Dependency Changes?  |

Changed Download component markup to meet accessibility standards. Now the whole component is a link which prevents link duplication that occurred in previous markup. Treating the whole content as a link allows authors to provide all the required information inside the link (like format and size of the file). Changed css to match the new structure.

This allows the component to be configured and styled in accessible way. Core Components example page styling and design prevents the instances of component used there from being accessible (missing outline on link, hidden labels of properties)

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/download/v1/download/README.md
+++ b/content/src/content/jcr_root/apps/core/wcm/components/download/v1/download/README.md
@@ -55,8 +55,8 @@ The following JCR properties are used:
 ## BEM Description
 ```
 BLOCK cmp-download
+    ELEMENT cmp-download__link
     ELEMENT cmp-download__title
-    ELEMENT cmp-download__title-link
     ELEMENT cmp-download__description
     ELEMENT cmp-download__properties
     ELEMENT cmp-download__property
@@ -64,7 +64,6 @@ BLOCK cmp-download
     ELEMENT cmp-download__property-label
     ELEMENT cmp-download__property-content
     ELEMENT cmp-download__action
-    ELEMENT cmp-download__action-text
 ```
 
 ## Information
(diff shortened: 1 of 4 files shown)


richardhand:
@jakublamprecht - thanks for this effort! These changes will require a new version of the component.

richardhand:
based this against a new `download/v2` branch. Some work is still left here to introduce the download v2 as this is a breaking change.

PR #751, opened 2019-09-19
[Data Layer] PoC

jckautzmann:
Proof of concept for a data layer.

| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Documentation Provided   | N/A (code comments and or markdown)
| Any Dependency Changes?  |


content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,170 @@
...
+        });
+        return state;
+    }
+
+
+    // Augments the push function (only for the data layer object) to also update the data layer state
+    function overridePush() {
+        window.dataLayer.push = function() {
+            var event = arguments[0];
+            updateState(window.dataLayer.state, event);
```
richardhand: we should apply this to all arguments of the push so you can do something like:
window.dataLayer.push({
    "type": "my event 1"
    "data": {}
}, {
    "type": "my event 2"
    "data": {}
});

richardhand: And we could wrap the handling in a single function that can do all the necessary tasks (which will likely be more than simply updating the state, such as event handling, validating the event etc.) before calling the prototype push that adds it to the array.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,170 @@
...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+(function() {
+    "use strict";
+
+    /* eslint no-console: "off" */
+
+    // Initializes the data layer
+    function init() {
+        window.dataLayer = window.dataLayer || [];
+        populateDataLayer();
```
richardhand: we can move this population logic outside of this file to the mocking scripts

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,170 @@
...
+                    }
+                }
+            }
+        });
+    }
+
+    function populateDataLayerAfterOverride() {
+        window.dataLayer.push({
+            "type": "updated",
+            "object": {
```
richardhand: I'm not keen on this "object" key, while not a reserved word in JavaScript I think we can find something better.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,183 @@
...
+        window.dataLayer.state = getComputeState(window.dataLayer);
+        overridePush();
+        populateDataLayerAfterOverride();
+        console.log("data layer script initialized");
+    }
+
+    // Returns the computed state by iterating over all the events stored in the data layer array
+    function getComputeState(dataLayer) {
+        var state = {};
+        dataLayer.forEach(function(event) {
```
richardhand: This won't work on IE. We should decide on browser support.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,183 @@
...
+                    "target": {},
+                    "handler": function(component) {
+                        console.log("clicked on component: " + component);
+                    }
+                }
+            }
+        });
+    }
+
+    function populateDataLayerAfterOverride() {
```
richardhand: for now, we can trigger a custom event on the window object once we've initialized:
```
var readyEvent = new CustomEvent("datalayer:ready");
this.window.dispatchEvent(readyEvent);
```

and add the after population based on that:
```
window.addEventListener("datalayer:ready", function() { console.log("data layer ready"); });
```

will allow us to decouple the mocks you have already from the core handling.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js.txt
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###############################################################################
+
+#base=js
+
+datalayer.js
+consumer.js
```
richardhand: we should really put the mock files before the datalayer. Might even be better to have two clientlibs that aren't embedded:
• `core.wcm.components.commons.datalayer.mocks`
• `core.wcm.components.commons.datalayer`
and include them in that order

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
jckautzmann: Makes sense! I implemented it in a recent commit.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
jckautzmann: According to [0] it is supported in IE9. 
According to [1] IE11 is supported with AEM 6.3. So it should be fine.

[0] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
[1]https://helpx.adobe.com/experience-manager/6-3/sites/deploying/using/technical-requirements.html

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
jckautzmann: how about "data" or "branch"? "details" sounds a bit weird to me.

richardhand: "data" works for me

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
richardhand: I was thinking of the wrong function.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    // Augments the push function (only for the data layer object) to also handle the event
+    function overridePush() {
+        window.dataLayer.push = function() {
+            var pushArguments = arguments;
+            Object.keys(pushArguments).forEach(function(key) {
+                handleEvent(pushArguments[key]);
+            });
+            return Array.prototype.push.apply(this, pushArguments);
```
richardhand: we should make sure here that only events, not listeners are pushed to the array.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,218 @@
...
+        };
+    }
+
+    function handleEvents(dataLayer) {
+        dataLayer.forEach(function(event) {
+            handleEvent(event);
```
richardhand: we should remove listeners from the array when we initially scrape the events that are defined before the script has initialized.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
richardhand: and maybe we should trigger listeners after we've pushed to the array, otherwise `window.dayaLayer` won't yet have the just-pushed event available.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,218 @@
...
+                console.log(this.state);
+            }
+        });
+    }
+
+    function populateDataLayerAfterOverride() {
+        window.dataLayer.push({
+            "type": "page updated",
+            "data": {
+                "page": {
```
sentah: This assume that the data layer object is single layered and not nested ? can we have the values be derived from the component itself ? e.g the pageCategory can be either a page properties or user defined values ?

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,218 @@
...
+    // Initializes the data layer
+    function init() {
+        window.dataLayer = window.dataLayer || [];
+        window.dataLayer.state = {};
+        window.dataLayer._listeners = [];
+        populateDataLayer();
+        handleEvents(window.dataLayer);
+        overridePush();
+        populateDataLayerAfterOverride();
+        console.log("data layer script initialized");
```
lewandowskidawid: should we write the message on console in the final version of the code?

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
jckautzmann: The data layer object can be either single layered or nested as it will be populated by providers. Some providers will be out of the box in AEM or Core Components, some will be custom, e.g. available with custom components. This is the idea for now but it might change as it is for now only a proof of concept.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
jckautzmann: This pull request is a proof of concept. The logs only simulate the behavior and will be removed when the concept is validated.

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,264 @@
...
+        window.dataLayer._listeners.push(listener);
+        console.log("register an event listener: " + listener.on);
+    }
+
+    function triggerListeners(event) {
+        // loop over all the listeners
+        // when a match is found, execute the handler
+        Object.keys(window.dataLayer._listeners).forEach(function(key) {
+            var listener = window.dataLayer._listeners[key];
+            if (listener.on === CHANGE_EVENT || listener.on === event.type) {
```
richardhand: wondering what order we should fire these in:
1. `event.type` first, then `CHANGE_EVENT` events
2. `CHANGE_EVENT` events first, then `event.type` events
3. The order they were registered regardless of type

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -0,0 +1,392 @@
...
+        that.state = {};
+
+        /**
+         * Returns a deep copy of the data layer state.
+         *
+         * @returns {Object} The deep copied state object.
+         */
+        that.dataLayer.getState = function() {
+            // use deep copying technique of JSON stringify and parsing the state.
+            return JSON.parse(JSON.stringify(that.state));
```
grubyak: intention is to clone state, so that if you alternate returned value will now change original object?
asking, because stringify + parse is pretty slow

bdelacretaz: https://github.com/immutable-js/immutable-js might be an alternative to produce immutable objects. My Javascript fu is not sufficient to judge how good that is, I'll leave that to specialists!

richardhand:
Thanks all for your comments on this PoC, development has now moved to https://github.com/adobe/adobe-client-data-layer and this will be closed.

joerghoh:
HI @richardhand , can you quickly explain how the integration of the mentioned project into the core component is planned?

richardhand:
Hi @joerghoh - the current plan is publish that project as an npm package, that can be built into an AEM client library. We're not sure exactly yet where that would sit.

We also plan to enhance the Core Components to write to the data layer with Core Components specific data and events. Of course customer application code is then free to do the same.

PR #753, opened 2019-09-19
[Image] Getting 404 response for SVG assets when image quality is not…

### LINKED ISSUE #679: [Image] Broken image when the quality property is set without smart widths
## Bug Report

**Current Behavior**
When setting only the JPEG quality without setting any smart widths, the image is broken.
Note: when setting one or more smart widths, the image renders correctly.

Steps to reproduce:
1. Go to the We.Retail Experience Page template.
2. Add policy to heroimage component at the top (not to the one inside the Layout Container. Select jpegquality as ‘100’.
3. View content/we-retail/language-masters/en/experience/arctic-surfing-in-lofoten.html after clearing browser cache. Notice Hero Image is broken.
4. If you had 82 for image quality, the image renders correctly.

I was able to reproduce this issue with the v2 image component from the Component Library:
http://localhost:4502/editor.html/content/core-components-examples/library/image.html

**Expected behavior/code**
When setting only the JPEG quality without setting any smart widths, the quality property should be applied to the original image.

**Environment**
- current AEM build
- Core Components version: 2.5.0

This issue was created following some feedback from the following thread in our email list:
https://groups.google.com/d/msgid/aem-core-components-dev/b51aa3db-a151-42e1-87ce-0a238ab06c18%40googlegroups.com?utm_medium=email&utm_source=footer

### LINKED ISSUE #735: [Image] Getting 404 response for SVG assets when image quality is not set to default
## Bug Report

**Current Behavior**
Enable Image component for an editable template (in layout container)
Create a page using the template and Add Image component
Add an SVG Image from DAM to the image component
 - The SVG Image will display
Now go to template, and add a policy for the image component
In the image policy - quality is required field so configure 100 on the quality. 
Reload the page, you'll see the image will not show (it gets a 404)

Upon inspecting the logs, I noticed 

com.adobe.cq.wcm.core.components.internal.servlets.AdaptiveImageServlet The requested quality (82) is not allowed by the content policy.

I updated the JPEG Quality on policy to 82 and the image started showing (not getting a 404 anymore). 

**Expected behavior/code**
The svg image should show regardless of what is set on the quality 

**Environment**
- AEM 6.5.0
- Core Components version: 2.6.1-SNAPSHOT

bpauli:
… set to default

- add support for key / value pairs in AdaptiveImageServlet and Image components

fixes #735, fixes #679

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
@@ -15,12 +15,19 @@
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.internal;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.jetbrains.annotations.NotNull;
 
 import com.day.cq.wcm.api.Page;
 import com.day.cq.wcm.api.PageManager;
+import com.google.common.base.Splitter;
+import com.google.common.collect.Lists;
 
 public class Utils {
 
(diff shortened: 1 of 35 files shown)


jckautzmann:
I agree with @kwin , this is a new mechanism which is not backwards compatible. I wouldn't change the expected format of the image URL. I think we should only fix the SVG issue.

bpauli:
@kwin agree, created a new PR at https://github.com/adobe/aem-core-wcm-components/pull/754 which is backward compatible. Looking forward for your feedback, Thanks!

PR #756, opened 2019-09-20
[Container] Not possible to define policies for allowed components in…

### LINKED ISSUE #736: [Container] Not possible to define policies for allowed components in editable template
## Bug Report
**Current Behavior**
Scenario: A proxy component of the Container v1 component is created like this:
```javascript
{
  "jcr:primaryType": "cq:Component",
  "jcr:title": "Container",
  "componentGroup": "my-component-group",
  "sling:resourceSuperType": "core/wcm/components/container/v1/container",
  "cq:isContainer": true
}
```
It is defined as unlocked structure component in an editable template, and some components are allowed in it's component dialog.
But although the component is unlocked it shows no "button" for the allowed components, and thus does not allow to define policies for the allowed components in the container.

The problem is not visible in the component library because there the container component is not inserted as structure component in the editable template.

**Expected behavior/code**
The container should behave exactly the same as the General / Layout Container component in editable template mode, allowing to define policies for the allowed components.

**Environment**
- AEM 6.5.0
- Core Components 2.6.0

**Possible Solution**
If i statically change the container.html to only include responsiveGrid.html _and_ if i remove the surrounding DIV with class=cmp-container in responsiveGrid.html the allowed component buttons are displayed. The same "hack" does not work with simple.html. And it is no solution anyways as it breaks the markup and background features.

See screenshot for comparison:
container-allowed-components

bpauli:
… editable template

- fix issue with missing allowed components list in structure mode

fixes #736

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/container/v1/container/simple.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/container/v1/container/simple.html
@@ -14,15 +14,17 @@
   ~ limitations under the License.
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
 <template data-sly-template.simple="${ @ container}">
-    <sly data-sly-test="${wcmmode.edit}" data-sly-use.allowed="com.day.cq.wcm.foundation.AllowedComponentList"></sly>
+    <sly data-sly-test="${wcmmode.edit}" data-sly-use.allowed="com.day.cq.wcm.foundation.model.AllowedComponents"></sly>
     <div data-sly-use.templates="core/wcm/components/commons/v1/templates.html"
          id="${container.id}"
-         class="cmp-container"
+         class="cmp-container ${allowed.cssClass}"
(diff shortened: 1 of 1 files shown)


content/src/content/jcr_root/apps/core/wcm/components/container/v1/container/simple.html
```diff
@@ -14,15 +14,17 @@
   ~ limitations under the License.
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
 <template data-sly-template.simple="${ @ container}">
-    <sly data-sly-test="${wcmmode.edit}" data-sly-use.allowed="com.day.cq.wcm.foundation.AllowedComponentList"></sly>
+    <sly data-sly-test="${wcmmode.edit}" data-sly-use.allowed="com.day.cq.wcm.foundation.model.AllowedComponents"></sly>
     <div data-sly-use.templates="core/wcm/components/commons/v1/templates.html"
          id="${container.id}"
-         class="cmp-container"
+         class="cmp-container ${allowed.cssClass}"
```
richardhand: shouldn't we only add these classes in edit mode?

Issue #688, opened 2019-08-09
[Navigation][Accessibility] Inconsistent isActive behavior


jakublamprecht:
## Bug Report

**Current Behavior**
In `NavigationItem.java` file there is a documentation for `isActive` method which says, that it returns true if the current page is represented by this item, and false otherwise:

```
    /**
     * Returns {@code true} if the page contained by this navigation item is active.
     *
     * @return {@code true} if it is the current page, otherwise {@code false}
     * @since com.adobe.cq.wcm.core.components.models 11.0.0; marked <code>default</code> in 12.1.0
     */
```

This is not exactly correct - `isActive` is true if the item represents current page or is a parent of the current page - this is leveraged in the markup of navigation to add the `cmp-navigation__item--active`.

This is a problem for `aria-current` attribute which should be set only on the link to the current page, but currently, because it uses said `isActive` flag, the `aria-current` attribute is present also on any parent of the item.

**Expected behavior/code**

Only the item that represents current page should have `aria-current="page"` attribute.

**Environment**
- Core Components 2.5.0

msagolj:
@jakublamprecht you are right, the implementation is correct but the JavaDoc is wrong, it should say:

'if it is the current page or one of its ancestors'

 and the code for aria-current is using it incorrectly.

msagolj:
(CQ-4277267)

gabrielwalt:
We could just change the `aria-current="page"` attribute to be set only for the current page without requiring a new version of the component.
https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/navigation/v1/navigation/itemContent.html#L18

PR #782, opened 2019-10-05
[Navigation][Accessibility] Inconsistent isActive behavior #688

pqrth:
| Fixed Issues?            | #688 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |

The javadoc of `NavigationItem#isActive`stated that it returns `true` if it is the current page, but in reality it returns `true` if it is the current page or one of its ancestors. So, fixed the javadoc.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/NavigationItem.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/NavigationItem.java
@@ -45,7 +45,7 @@ public interface NavigationItem extends ListItem {
     /**
      * Returns {@code true} if the page contained by this navigation item is active.
      *
-     * @return {@code true} if it is the current page, otherwise {@code false}
+     * @return {@code true} if it is the current page or one of its ancestors, otherwise {@code false}
      * @since com.adobe.cq.wcm.core.components.models 11.0.0; marked <code>default</code> in 12.1.0
      */
     default boolean isActive() {


bpauli:
@pqrth Thank you for providing this PR! Unfortunately the isActive is not strictly implemented. In the BreadcrumbItemImpl the active page is only the current page https://github.com/adobe/aem-core-wcm-components/blob/3994b1e24ed241df70c30b82916d5e4911714878/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/BreadcrumbImpl.java#L96

@gabrielwalt How would we like to deal with this? Should we create a new version of the Breadcrumb component to fix this? We should also think about an extension of the NavigationItem interface by providing an isCurrent method. Which will only return true for the the current page.

bpauli:
On the other side... the Breadcrumb component seems to be correct. Maybe we should keep the javadoc and fix the implementation in the NavigationImpl and LanguageNavigationImpl component and providing an isAncestor method for the NavigationItem.

richardhand:
based this against a `navigation/v2` branch. We will need to add to the PR to introduce a navigation v2 that removes the `aria-current` attribute from ancestor items and introduces the new discussed API, seems like the choice is between either:

• `isActive` and `isCurrent` - where active indicates ancestry
• `isAncestor` and `isActive` - where active indicates current

PR #785, opened 2019-10-07
[Image v2] String function use should be optimized for single characters

kotisyamala:
| Fixed Issues?            | String function use should be optimized for single characters
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |

An indexOf or lastIndexOf call with a single letter String can be made more performant by switching to a call with a char argument.


bpauli:
@kotisyamala Yes, makes sense. Thanks!

PR #786, opened 2019-10-07
Removed role="button" from buttons.

### LINKED ISSUE #783: [Carousel] Button elements have role button
## Bug Report

**Current Behavior**
In the current markup of Carousel, button elements for rotation and slide control have additional attribute `role="button`. This is not required, since buttons get this role by default. This causes W3C validators to report warnings. That's nothing major, but might be worth addressing.

**Expected behavior/code**
No role="button on button elements

**Environment**
- Core Components version 2.6.0

dlemstra:
| Fixed Issues?            | Fixes #783 

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
@@ -33,14 +33,12 @@
              data-cmp-hook-carousel="item"></div>
         <div class="cmp-carousel__actions">
             <button class="cmp-carousel__action cmp-carousel__action--previous"
-                    role="button"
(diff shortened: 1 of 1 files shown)


richardhand:
@jakublamprecht, @dlemstra - should we add `type=button` here? The default type for button is `submit` so I think that makes sense. A little out of scope of this issue, but we could perhaps add it while we're at it.

dlemstra:
That does make sense to add. Will push a new commit to change this during my lunch break.

richardhand:
Thanks @dlemstra! would be good to also have a quick check of `<button>`s used in other components.

dlemstra:
Maybe that should be a new PR instead?

richardhand:
@dlemstra - yep, I agree :)

PR #787, opened 2019-10-08
added 'cmp-image' css class

### LINKED ISSUE #728: images in content fragments not responsive
## Bug Report

**Current Behavior**
When a content fragment has an image, and the content fragment is used on a page, the image is not responsive.

**Expected behavior/code**
I expected the images in the content fragment to be responsive and follow the set lay out of the page.

**Environment**
- GA AEM 6.5
- core.wcm.components.extension.contentfragment.content version 1.0.12
- Java(TM) SE Runtime Environment (build 1.8.0_162-b12)

**Possible Solution**
One possible solution is to add the cmp-image css class to the template that renders the content fragment. You can read details of my workaround [here](https://52lina.wordpress.com/2019/06/03/why-imgs-in-cf-not-resp/).

oswjos:
| Fixed Issues?            |  
| Patch: Bug Fix?          | Fixes #728 
| Tests Added + Pass?      | 
| Documentation Provided   | 

The 'cmp-image' css class will make the images in the content fragment responsive.

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/contentfragment/v1/contentfragment/element.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/contentfragment/v1/contentfragment/element.html
@@ -16,7 +16,7 @@
 <template data-sly-template.element="${@ element='The content fragment element'}">
     <div class="cmp-contentfragment__element cmp-contentfragment__element--${element.name}" data-cmp-contentfragment-element-type="${element.dataType}">
         <dt class="cmp-contentfragment__element-title">${element.title}</dt>
-        <dd class="cmp-contentfragment__element-value">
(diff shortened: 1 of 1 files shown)


richardhand:
Hi @oswjos, thanks for the pull request! I don't think this is the correct fix, as the `.cmp-image` class is private to the image component and adding it for each content fragment element value (which could be of differing types) may have unintended side effects.

A more general approach is required here, such as delegating the rendering of images in content fragments to a particular template or component. Unfortunately the `${element.dataType}` in this case is `String` so I don't see a way currently to switch for image in the content fragment component. It may be that an enhancement is required in the content fragment core logic.

If you don't mind I will close this PR, but leave the current issue open, so we still have the problem tracked and can discuss a solution there.

PR #789, opened 2019-10-08
Issue 788 - trim the richtext html value before saving in data attribute

### LINKED ISSUE #788: [Teaser] Blank lines in teaser description when "Link" or CTA is filled
## Bug Report

**Reproduction steps**
1. Navigate to http://localhost:4502/editor.html/content/core-components-examples/library/teaser.html
2. Go to teaser below "Call To Action Button"
3. Open the dialog
4. Focus on "Description" richtext area after the last character
5. Invoke backspace button 18 times (exact umber of char in "Teaser Description")
6. Confirm the dialog
 

**Current Behavior**
The empty "&nbps;" is present in description

**Expected behavior/code**
No additional white space characters should be visible as the user already deleted all the chars.

**Environment**
- Windows 10, Chrome Version 77.0.3865.90 (Official Build) (64-bit)
- AEM version and patch level - AEM 6.3 SP3 CFP5
- Core Components version 2.6.0
- JRE version: Java(TM) SE Runtime Environment (build 1.8.0_171-b11)

**Possible Solution**
https://github.com/adobe/aem-core-wcm-components/pull/789

teaser

rzasap:
| Fixed Issues?            | Fixes #788 
| Patch: Bug Fix?          | Yes - bugfix
| Major: Breaking Change?  | Should not break anything
| Tests Added + Pass?      | No new tests, changed internal implementation
| Documentation Provided   | No, changed internal implementation only

Added the trimming of richtext html before saving the value in data attribute to prevent unwanted white spaces in teaser description field

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/commons/v1/clientlibs/editor/checkboxTextfieldTuple/checkboxTextfieldTuple.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/commons/v1/clientlibs/editor/checkboxTextfieldTuple/checkboxTextfieldTuple.js
@@ -210,11 +210,16 @@
      * @private
      */
     CheckboxTextfieldTuple.prototype._getTextfieldValue = function() {
+        var result;
         if (this._isRichText) {
-            return $(this._textfield).html();
+            result = $(this._textfield).html();
+            if (result) {
+                result = result.trim();
+            }
         } else {
-            return this._textfield.value;
+            result = this._textfield.value;
         }
+        return result;
     };
 
     /**


richardhand:
Hi @rzasap thanks for the PR! I'm not sure this is the ideal fix, both for its location and behaviour. We could be trimming intentional line breaks here.

Having said that, I'm not sure yet where the fix should be. There is also a workaround of course which is one more backspace to remove the return character.

rzasap:
Hi @richardhand,

Please note that the issue does not occurs for Teasers without "Link" filled or with link pointing to absolute urls. So it looks like checkboxTextfieldTuple.js is messing here.

I think that the #788 is strictly related to checkboxTextfieldTuple.js and it's read method. jQuery html method give us all the whole html including trailing white characters, which we then save in data attribute. From the other side when setting the value the richtext automatically convert the /cr/lf characters into p tag.

I haven't found any better place for the fix. I think it could be either here or in rich text, but for some reasons the rich text is implemented in this way that during setting the value the /cr/lf characters are automatically converted to p tag. (This might be reasonable for example when user copy & paste the text)

I am not sure about the intentional line breaks. The fix affects only rich text fields for which any of trailing white characters does not matter (as far I know, each richtext value is always wrapped either with p, hx or some other tag)

rzasap:
I see that the branch should be based on development branch. If you are ok with the changes, please let me know then I will cherry pick the commit to new branch based on development

richardhand:
Thanks @rzasap for the clarification I agree that this is the right place for the fix. I think a better approach here is to allow the richtext editor to do the processing (which is quite in depth) rather than process the html of the editable div ourselves. I see a couple of options to explore:

**Approach 1:** Read the value from the hidden field associated with the richtext input. This is written to by the richtext editor after it has done some processing of the html.

```javascript
CheckboxTextfieldTuple.prototype._getTextfieldValue = function() {
    if (this._isRichText) {
        var $richTextContainer = $(this._textfield).closest(".cq-RichText");
        if ($richTextContainer) {
            var $richTextInput = $richTextContainer.find("input[type='hidden'][data-cq-richtext-input='true']");
            if ($richTextInput) {
                return $richTextInput.val();
            }
        }
    } else {
        return this._textfield.value;
    }
};
```

**Approach 2:** Use the RTE instance api to get the content
```
CheckboxTextfieldTuple.prototype._getTextfieldValue = function() {
        if (this._isRichText) {
            var rteInstance = $(this._textfield).data("rteinstance");
            if (rteInstance) {
                return rteInstance.getContent();
            }
        } else {
            return this._textfield.value;
        }
    };
```

Approach 2 seems cleaner, but it has a problem, in that the `rteInstance.getContent()` method relies on an `rteInstance.isActive` flag, if the RTE isn't active then it returns `undefined`. BUT, the RTE isn't marked active straight away in the dialog as it does some ajax calls for configurations. The RTE triggers an `editing-start` event on the richtext field `$(this._textfield).on('editing-start')` when it's ready, so we could wrap some handling in that, but I'm not sure where.

msagolj:
@rzasap , let RTE do the work seems to be the right way to go judging from discussion above with @richardhand .

His proposed 'Approach 2' seems to come with its own quirks. So would you be ok to create a new PR against development branch using his 'Approach 1' ?

richardhand:
Alternative proposal in #824

msagolj:
fixed with #824, so we close this one.

PR #793, opened 2019-10-11
Form Container - Resizable parsys #775

### LINKED ISSUE #775: Form Container - Resizable parsys
## Bug Report

**Current Behavior**
Currently, parsys inside Form Container component could be resized that creates redundant nodes on the same level. 
For Responsive Grid component, there is a logic in responsive.Overlay.js for making drop zones not resizable. 

` // the responsive grid drop zones are not resizable
            if (editable.type !== 'wcm/foundation/components/responsivegrid/new') {`

It goes without saying that it doesn't work due to inheritance and different resource type.

**Expected behavior/code**
I shouldn't be able to change the width of parsys as it is currently designed for Layout Container

**Environment**
- AEM 6.4 SP 6.4.5
- Core Components version 2.5

image

andreeadracea:
| Fixed Issues?            | `Fixes #775 
| Patch: Bug Fix?          | Yes - bugfix
| Tests Added + Pass?      | No new tests

   - changed the resourceType of the drop area

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/form/ContainerImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/form/ContainerImpl.java
@@ -123,7 +123,7 @@ public class ContainerImpl implements Container {
             id = FormsHelper.getFormId(request);
         }
         this.name = id;
-        this.dropAreaResourceType += "/new";
+        this.dropAreaResourceType = "wcm/foundation/components/responsivegrid/new";
         if (redirect != null) {
             String contextPath = request.getContextPath();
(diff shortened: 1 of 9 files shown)


bpauli:
@andreeadracea Do we still need the node at https://github.com/adobe/aem-core-wcm-components/tree/master/content/src/content/jcr_root/apps/core/wcm/components/form/container/v1/container/new after this change?

richardhand:
@andreeadracea, what is `container.resourceTypeForDropArea` used for now that we hard code this resource type in the HTL script?

andreeadracea:
@bpauli, we don't need that nodes anymore but I didn't delete them because they can be used by the customers in their implementation.
@richardhand, same for the `container.resourceTypeForDropArea`, we don't use it anymore but I don't know if we should remove it and if that would affect the customers.

PR #798, opened 2019-10-13
Unnecessary imports should be removed

kotisyamala:
| Fixed Issues?            | Unnecessary imports should be removed
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

The imports part of a file should be handled by the Integrated Development Environment (IDE), not manually by the developer.
Unused and useless imports should not occur if that is the case.
Leaving them in reduces the code's readability, since their presence can be confusing.


bpauli:
@kotisyamala Thank you for the PR

PR #800, opened 2019-10-14
Null pointers should not be dereferenced

kotisyamala:
| Fixed Issues?            | Null pointers should not be dereferenced
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

A reference to null should never be dereferenced/accessed. Doing so will cause a NullPointerException to be thrown. At best, such an exception will cause abrupt program termination. At worst, it could expose debugging information that would be useful to an attacker, or it could allow an attacker to bypass security measures

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/editor/dialog/childreneditor/Editor.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/editor/dialog/childreneditor/Editor.java
@@ -62,11 +62,13 @@ public class Editor {
             container = resolver.getResource(containerPath);
             if (container != null) {
                 ComponentManager componentManager = request.getResourceResolver().adaptTo(ComponentManager.class);
-                for (Resource resource : container.getChildren()) {
-                    if (resource != null) {
-                        Component component = componentManager.getComponentOfResource(resource);
-                        if (component != null) {
-                            items.add(new Item(request, resource));
+                if(componentManager != null){
+                    for (Resource resource : container.getChildren()) {
+                        if (resource != null) {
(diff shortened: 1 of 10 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/editor/dialog/childreneditor/Editor.java
```diff
@@ -62,11 +62,13 @@ private void readChildren() {
             container = resolver.getResource(containerPath);
             if (container != null) {
                 ComponentManager componentManager = request.getResourceResolver().adaptTo(ComponentManager.class);
-                for (Resource resource : container.getChildren()) {
-                    if (resource != null) {
-                        Component component = componentManager.getComponentOfResource(resource);
-                        if (component != null) {
-                            items.add(new Item(request, resource));
+                if(componentManager != null){
```
joerghoh: is it possible to come up with a much shorter version using {{Optional}}s?

```
items = container.getChildren().stream()
  .map(res -> componentManager.getComponentOfResource(res)
  .filter(Objects::notNull)
  .map(new Item(request,res)
  .collect(Collectors::toList());
```

(you might want to try to include the earlier statements as well). This reduces the number of branches and makes it easier to understand the code.)

msagolj: Hi @joerghoh I will merge this PR for now as is seems @kotisyamala is not going to change it. There is a pending PR (#814) which actually applies the same fix , so i will try to incorporate your suggestions there.

Issue #804, opened 2019-10-17
Unable to Run Individual jUnit Tests in Eclipse


klcodanr:
## Bug Report

**Current Behavior**
Running individual jUnit tests in Eclipse with JDK 8 results in the following exception:

```java.lang.NoClassDefFoundError: org/junit/platform/commons/PreconditionViolationException
	at org.eclipse.jdt.internal.junit5.runner.JUnit5TestLoader.createUnfilteredTest(JUnit5TestLoader.java:75)
	at org.eclipse.jdt.internal.junit5.runner.JUnit5TestLoader.createTest(JUnit5TestLoader.java:66)
	at org.eclipse.jdt.internal.junit5.runner.JUnit5TestLoader.loadTests(JUnit5TestLoader.java:53)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:525)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:763)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:463)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:209)
Caused by: java.lang.ClassNotFoundException: org.junit.platform.commons.PreconditionViolationException
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 7 more
```

**Expected behavior/code**

Unit tests can be executed.

**Environment**
- N/A
- master branch
- `OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_222-b10)`

**Possible Solution**
Working on a PR, comment to come

richardhand:
The code for this issue is merged to master and will be available in the upcoming [2.8.0 release](https://github.com/adobe/aem-core-wcm-components/milestone/18).

PR #805, opened 2019-10-17
Unable to Run Individual jUnit Tests in Eclipse 

klcodanr:
| Fixed Issues?            | Fixes: #804
| Patch: Bug Fix?          | X
| Minor: New Feature?      | 
| Major: Breaking Change?  |
| Documentation Provided   | N/A
| Any Dependency Changes?  | Yes, see below

Minor fix to add junit-platform-commons version 1.5.2 to provide the missing dependency when running jUnit tests in Eclipse.

### DIFF
--- a/bundles/core/pom.xml
+++ b/bundles/core/pom.xml
@@ -476,6 +476,12 @@
             <groupId>uk.org.lidalia</groupId>
             <artifactId>slf4j-test</artifactId>
         </dependency>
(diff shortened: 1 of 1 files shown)


vladbailescu:
This seems to be a way to avoid an Eclipse bug. I'm wondering if there's a cleaner one or, at least, if we could add this to an Eclipse-specific profile.

PR #821, opened 2019-10-31
String literals should not be duplicated

kotisyamala:
| Fixed Issues?            | String literals should not be duplicated
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

Duplicated string literals make the process of refactoring error-prone, since you must be sure to update all occurrences.
On the other hand, constants can be referenced from many places, but only need to be updated in a single place.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/DownloadServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/DownloadServlet.java
@@ -162,7 +162,7 @@ public class DownloadServlet extends SlingAllMethodsServlet {
                     }
                 }
             } catch (RepositoryException e) {
-                LOG.info ("Cannot determine size of binary at {}", resource.getPath(),e);
+                LOG.error("Cannot determine size of binary at {}", resource.getPath(),e);
             }
         }
         return -1;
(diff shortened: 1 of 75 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/DownloadServlet.java
```diff
@@ -190,7 +190,7 @@ private void sendResponse(InputStream stream, long size, String mimeType, String
             response.setContentLength((int) size);
         }
         if (inline) {
-            response.setHeader(CONTENT_DISPOSITION_HEADER, "inline");
+            response.setHeader(CONTENT_DISPOSITION_HEADER, INLINE_SELECTOR);
```
jckautzmann: I would not use `INLINE_SELECTOR` as it represents a request selector, which has a different meaning than the content disposition header which is used here. You might want to introduce a new constant called e.g. `CONTENT_DISPOSTION_INLINE`.

gabrielwalt:
@kotisyamala, Will you make the requested changes? Else, we'll close this PR.

kotisyamala:
> @kotisyamala, Will you make the requested changes? Else, we'll close this PR.

Will raise the changes to this PR today.

Thanks,
Koti

gabrielwalt:
Closing this for now, I'm not sure this is even a desired contribution.

PR #831, opened 2019-11-05
[ExperienceFragment] "Edit in New Tab" action does not work

### LINKED ISSUE #815: [ExperienceFragment] "Edit in New Tab" action does not work
## Bug Report

**Current Behavior**
When an Experience Fragment component is placed on a page, and a variation is selected, the "Edit" action does not open the experience fragment editor. Instead, it displays the error "The experience fragment doesn't have an associated variation"

**Expected behavior/code**
The edit button should open the experience fragment in the experience fragment editor.
This is the behavior of the OOTB vanilla XF component, and is the obvious intent of the editInNewTab action config.

**Environment**
- AEM version: AEM-6.4.6.0
- Core Components version: 2.7.0
- JRE version: 1.8.0

**Root Cause**
The issue is introduced at [/apps/core/wcm/components/experiencefragment/v1/experiencefragment/cq:editConfig/cq:actionConfigs/editInNewTab](https://github.com/adobe/aem-core-wcm-components/blob/6ab01a4f737b9a397fb8a0f56aa845bd940ac076/content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/_cq_editConfig.xml#L36).

This action config causes the following JS to be called:
/libs/cq/experience-fragments/editor/clientlibs/experiencefragments/js/component/actions/edit.xf.NEWTAB.js

That action explicitly opens in a new tab the fragment referenced by the "fragmentPath" property.
The problem is that the core components XF component does not use the "fragmentPath" property, instead it uses a property named "fragmentVariationPath" (see [dialog](https://github.com/adobe/aem-core-wcm-components/blob/6ab01a4f737b9a397fb8a0f56aa845bd940ac076/content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/_cq_dialog/.content.xml#L54))

**Possible Solution**
There are several options for resolving this issue.
The first possible solution is to just change the property name from "fragmentVariationPath" -> "fragmentPath". If this change is made then the properties mirror the vanilla OOTB XF component properties, and therefore the clientlib for the edit action still works. This is probably the "cleanest" solution, but would be a breaking change because existing components would cease to work unless the content is updated.

Another solution is to copy the edit.xf.NEWTAB.js file into a clientlib and modify the function to load the "fragmentVariationPath" property instead. This JS function would have to have a different name, or the OOTB and core-wcm-components would conflict.

I would lean towards the first solution unless there is a valid reason that the property name is different, or if the desired behavior is different than the vanilla OOTB XF component.

msagolj:
| Fixed Issues?            | Fixes #815 
| Documentation Provided   | no

- created our own edit action, instead of using the one from Experience Fragment 
- instead of using variationPath returned by .json, using localizedVariationPath from .model.json

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/_cq_editConfig.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/_cq_editConfig.xml
@@ -33,7 +33,7 @@
         <editInNewTab
             jcr:primaryType="nt:unstructured"
             text="Edit"
-            handler="Granite.author.experienceFragments &amp;&amp; Granite.author.experienceFragments.actions.editInNewTab"
+            handler="CQ.CoreComponents.experienceFragment.v1.actions.editInNewTab"
             icon="edit"/>
     </cq:actionConfigs>
 </jcr:root>
--- a/content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/_cq_editConfig.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/_cq_editConfig.xml
@@ -33,7 +33,7 @@
         <editInNewTab
             jcr:primaryType="nt:unstructured"
             text="Edit"
-            handler="CQ.CoreComponents.experienceFragment.v1.actions.editInNewTab"
+            handler="CQ.CoreComponents.experiencefragment.v1.actions.edit"
             icon="edit"/>
     </cq:actionConfigs>
 </jcr:root>
--- a/content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/clientlibs/editor/js/experiencefragment.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/clientlibs/editor/js/experiencefragment.js
@@ -1,0 +1,35 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2019 Adobe
+ ~
+ ~ Licensed under the Apache License, Version 2.0 (the "License");
+ ~ you may not use this file except in compliance with the License.
+ ~ You may obtain a copy of the License at
+ ~
+ ~     http://www.apache.org/licenses/LICENSE-2.0
+ ~
+ ~ Unless required by applicable law or agreed to in writing, software
+ ~ distributed under the License is distributed on an "AS IS" BASIS,
+ ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ ~ See the License for the specific language governing permissions and
+ ~ limitations under the License.
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+/* global CQ */
+(function($, ns) {
+    "use strict";
+
+    var MSG_NO_FRAGMENT_PATH = Granite.I18n.get("The experience fragment doesn't have an associated variation");
+
+    ns.experienceFragment.v1.actions.editInNewTab = function() {
+        var ui = $(window).adaptTo("foundation-ui");
+        $.get(this.path + ".model.json")
+            .then(function(response) {
+                var path = response["localizedFragmentVariationPath"];
+                if (typeof path !== "undefined") {
+                    window.open(Granite.HTTP.externalize("/editor.html" +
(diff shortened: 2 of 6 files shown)


content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/clientlibs/editor/js/experiencefragment.js
```diff
@@ -0,0 +1,35 @@
...
+ ~ See the License for the specific language governing permissions and
+ ~ limitations under the License.
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+/* global CQ */
+(function($, ns) {
+    "use strict";
+
+    var MSG_NO_FRAGMENT_PATH = Granite.I18n.get("The experience fragment doesn't have an associated variation");
+
+    ns.experienceFragment.v1.actions.editInNewTab = function() {
```
richardhand: we typically don't camelCase the component namespaces `CQ.CoreComponents.panelcontainer` `CQ.CoreComponents.panelselect` etc. we could be consistent here and instead use `CQ.CoreComponents.experiencefragment`

content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/clientlibs/editor/js/experiencefragment.js
```diff
@@ -0,0 +1,35 @@
...
+ ~ distributed under the License is distributed on an "AS IS" BASIS,
+ ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ ~ See the License for the specific language governing permissions and
+ ~ limitations under the License.
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+/* global CQ */
+(function($, ns) {
+    "use strict";
+
+    var MSG_NO_FRAGMENT_PATH = Granite.I18n.get("The experience fragment doesn't have an associated variation");
```
richardhand: To make this a little clearer, we could perhaps use?
"This experience fragment component doesn't have an associated variation"

content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/clientlibs/editor/js/experiencefragment.js
```diff
@@ -0,0 +1,35 @@
...
+
+    ns.experienceFragment.v1.actions.editInNewTab = function() {
+        var ui = $(window).adaptTo("foundation-ui");
+        $.get(this.path + ".model.json")
+            .then(function(response) {
+                var path = response["localizedFragmentVariationPath"];
+                if (typeof path !== "undefined") {
```
richardhand: `typeof path === "string"` might be a little more robust here

content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/_cq_editConfig.xml
```diff
@@ -33,7 +33,7 @@
         <editInNewTab
             jcr:primaryType="nt:unstructured"
             text="Edit"
-            handler="Granite.author.experienceFragments &amp;&amp; Granite.author.experienceFragments.actions.editInNewTab"
+            handler="CQ.CoreComponents.experienceFragment.v1.actions.editInNewTab"
```
richardhand: Can't we simply call the action `CQ.CoreComponents.experiencefragment.v1.actions.edit`, the new tab part seems redundant.

msagolj: Fixed in follow-up commit

content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/clientlibs/editor/js/experiencefragment.js
msagolj: fixed in follow-up commit + length check

content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/clientlibs/editor/js/experiencefragment.js
msagolj: done

content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/clientlibs/editor/js/experiencefragment.js
msagolj: fixed

PR #837, opened 2019-11-07
[List] Creating a list with tagging requires read rights on tags for …

### LINKED ISSUE #396: [List] Creating a list with tagging requires read rights on tags for everyone
## Bug Report

**Current Behavior**
We are creating a list based on tagging, however as /etc/tags is closed on publish, we need add on the  publish server **jcr:read** rights on the **everyone** group in order for the list to be shown.

**Expected behavior/code**
I would expect for a system user with jcr:read rights /etc/tags rights to get the pages instead of the user accessing the website.

**Environment**
- AEM 6.3 SP3
- Core Components 2.2.0

bpauli:
…everyone

- add service user handling to read tags for anonymous access

fixes #396

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ListImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ListImpl.java
@@ -94,7 +94,7 @@ public class ListImpl implements List {
     private Resource resource;
 
     @OSGiService
-    private ResourceResolverFactory resolverFactory;
+    private ResourceResolverFactory resourceResolverFactory;
 
     @Self
     private SlingHttpServletRequest request;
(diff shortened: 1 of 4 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ListImpl.java
```diff
@@ -244,7 +252,11 @@ private void populateTagListItems() {
         if (ArrayUtils.isNotEmpty(tags)) {
             Page rootPage = getRootPage(PN_TAGS_PARENT_PAGE);
             if (rootPage != null) {
-                TagManager tagManager = resourceResolver.adaptTo(TagManager.class);
+                ResourceResolver tagResourceResolver = getTagResourceResolver();
+                if (tagResourceResolver == null) {
+                    tagResourceResolver = resourceResolver;
+                }
+                TagManager tagManager = tagResourceResolver.adaptTo(TagManager.class);
```
lewandowskidawid: pls close opened resourceResolver when it is not needed to avoid stack trace about unclosed sessions

joerghoh: next to the unclosed resource resolver, I would raise the question if a dedicated service user is required to read the tags. Do we have to assume that the user executing the request is not able to read the tags?

bpauli: @lewandowskidawid, good catch! It's fixed in 228aa5784a710abf8dec4470d53c291a6c801a7e by using the try-with-resources syntax to benefit from the AutoClosable feature introduced in [1].

@joerghoh, as we want to keep the support for AEM 6.3 and 6.4 for the core-components the introduction of using a service resource resolver seems to be the only option as there is an explicit deny for everyone on the /etc path. The user which is used for this service is provided OOTB in AEM. Please let me know if you see another option.

[1] https://issues.apache.org/jira/browse/SLING-4805

joerghoh: I don't think that on 6.3 /etc/ is already locked down, this has been introduced with the repo reorg on 6.4.
And on 6.4 the tags are relocated to /content/, so there shouldn't be any need to have read access to /etc/, thus this service user is not necessary. And if you run 6.4 still with tags in /etc, then you have take care about it on your own.

I do think that we get away without service user. Plus on the other hand side, the creation of a new session on every request will slow down the rendering quite a bit.

bpauli: image
This is a local test publisher environment. Running on AEM 6.3 SP3 CFP4. We have this issue only for AEM 6.3. Maybe we can add an OSGi config option to enable the service user only on demand and use the resource resolver of the request as default?

bpauli:
Closing this PR without merging. Will update #396 with a valid configuration to solve the issue.

PR #841, opened 2019-11-11
[Text] RTE Features in Edit Dialog are not controlled by Component Po…

### LINKED ISSUE #158: [Text] RTE Features in Edit Dialog are not controlled by Component Policies
V1 Text Component does not have an Edit Dialog, it only provides in-line editing with fullscreen view. When customizing a Content Policy for the V1 Text component, all enabled features are available within the fullscreen view.

The V2 Text Component introduced a cq:dialog node which allows the Author to open an Edit Dialog. The Content Policies are not controlling the RTE features available within the Edit Dialog. This produces a conflicting Author Governance experience. 

summary:
I can control the RTE features in in-line fullscreen mode using Content Policies, but I cannot control the Edit Dialog RTE features with Content Policies.

desired results:
I can control the RTE features both within in-line fullscreen mode and within an Edit Dialog.

I'm not sure if controlling Edit Dialog features with Content Policies is even possible or intended, but there's definitely a conflict here. This could be an architectural problem rather than a code problem.

Screenshot of the Content Policy for a V2 Text Component:
screen shot 2018-04-05 at 4 58 40 pm

Screenshot of the in-line fullscreen RTE with features defined in above Policy:
screen shot 2018-04-05 at 4 57 46 pm

Screenshot of the Edit Dialog RTE with incorrect features defined in the above Policy, instead using the defined cq:dialog node for the V2 Text Component:
screen shot 2018-04-05 at 4 56 58 pm

Tested with We.Retail on AEM 6.3.1 with Featurepack-20593-1.8

msagolj:
…licies

| Fixed Issues?            | Fixes #158
| Documentation Provided   | current README is missing description of edit and design props, will create a seaprate documentation issue for that one

Removing the default `<dialogFullScreen>` configuration from core text component makes all its features controllable via content policies.

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/text/v2/text/_cq_dialog/.content.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/text/v2/text/_cq_dialog/.content.xml
@@ -161,16 +161,6 @@
                                                                     ref="paraformat"/>
                                                             </popovers>
                                                         </inline>
-                                                        <dialogFullScreen
-                                                            jcr:primaryType="nt:unstructured"
-                                                            toolbar="[format#bold,format#italic,format#underline,justify#justifyleft,justify#justifycenter,justify#justifyright,lists#unordered,lists#ordered,lists#outdent,lists#indent,links#modifylink,links#unlink,table#createoredit,#paraformat,image#imageProps]">
-                                                            <popovers jcr:primaryType="nt:unstructured">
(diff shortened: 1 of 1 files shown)


jckautzmann:
We should do the same for the other components that also rely on the RTE: Teaser and Download.

kamalc007:
Making this change on my local instance did not fix this issue. Can you please recheck on your end?

Issue #774, opened 2019-09-30
Adaptive Image Renditions for Background Images


HitmanInWis:
The WCM Core `Image` component has automatic adaptive images to ensure that appropriately sized images are used given the current viewport and CSS styling.  However, this same functionality does not apply for components with background images, such as `Container`.  This means if a user uploads a 10MB image for use of a `Container` background, the image will render on the web in its full 10MB glory.

Looking on the online guide for Core components at https://opensource.adobe.com/aem-core-wcm-components/library/container.html it appears there is a *start* of this approach as the image is being loaded with the `coreimg` selector:

```
<div class="cmp-container" style="background-image:url(image/_jcr_content/root/responsivegrid/demo_554582955/component/image.coreimg.jpeg/1561469321769/lava-into-ocean.jpeg);background-size:cover;background-repeat:no-repeat;background-color:#39CCCC;">
```

However, when I try this locally I dont actually get a URL rewritten to the `coreimg` selector, but rather the direct path to my image at /content/dam/path/to/my/image.jpg - is this because the online guide is using a sling mapping/rewrite for /content/dam paths?  If so, is that standard practice and can that rewrite rule be shared?

Assuming the image does get written with the `coreimg` selector, how much effort do we think it would be to get background images using an appropriately sized rendition?

HitmanInWis:
Looking at this more, its pretty simple to get a background image to render with `coreimg` selector by simply using the `Image` model to convert the image URL.  The HTL would also likely need to render a `data-cmp-widths` property.  I see `data-asset` and `data-asset-id` attributes as well, but I dont think those are used for adaptive imaging - perhaps Analytics.  Lastly, we'd maybe need a `data-cmp-is="bgimage"` or something that JS could pick up on.

Once a BG image is rendered I think the changes would all be within `/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js`.   It would need to be able to update a background image URL similar to how the image src is updated today, keying off of the `bgimage` component type from `data-cmp-is`.

One challenge I see with this would be a viable `noscript` solution, but frankly I'm not overly concerned with that especially if we make the feature opt-in via configuration.

Anyone see any huge blockers to this beyond doing the legwork to code it up?

numirov:
[...]

PR #842, opened 2019-11-11
[Container] Adaptive Image Renditions for Background Images #774

froggin:
Q | A
-- | --
Fixed Issues? | Fixes #774
Patch: Bug Fix? |  
Minor: New Feature? | Yes
Major: Breaking Change? | No
Tests Added + Pass? | Yes
Documentation Provided | Yes (code comments and or markdown)
Any Dependency Changes? |  
License | Apache License, Version 2.0

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
@@ -153,7 +153,9 @@
 
             that._elements.container = that._elements.link ? that._elements.link : that._elements.self;
 
-            if (!that._properties.isBgImage) unwrapNoScript();
+            if (!that._properties.isBgImage) {
+                unwrapNoScript();
+            }
 
             if (that._properties.lazy && !that._properties.isBgImage) {
                 addLazyLoader();
@@ -167,7 +169,7 @@
             window.addEventListener("resize", onWindowResize);
             window.addEventListener("update", that.update);
             if (!that._properties.isBgImage) {
-            that._elements.image.addEventListener("cmp-image-redraw", that.update);
+                that._elements.image.addEventListener("cmp-image-redraw", that.update);
             } else {
                 that._elements.self.addEventListener("cmp-image-redraw", that.update);
             }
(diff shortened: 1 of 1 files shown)


content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
```diff
@@ -372,7 +444,7 @@
     if (document.readyState !== "loading") {
         onDocumentReady();
     } else {
-        document.addEventListener("DOMContentLoaded", onDocumentReady);
+        document.addEventListener("DOMContentLoaded", onDocumentReady());
```
HitmanInWis: This line should be reverted - guessing you started from an old version of the file.  This was corrected in February 2019.

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
```diff
@@ -283,11 +333,19 @@
...
+
+                for (var i = 0; i < hooks.length; i++) {
+                    var loopHook = hooks[i];
+                    var loopCapitalized = IS;
+                    loopCapitalized = loopCapitalized.charAt(0).toUpperCase() + loopCapitalized.slice(1);
+                    var loopKey = loopHook.dataset[NS + "Hook" + loopCapitalized];
+                    that._elements[loopKey] = loopHook;
+                }
+            } else {
+                var hook = that._elements.self;
```
HitmanInWis: if we just do `var hooks = [that._elements.self];` and `var capitalized = BG;` for the case of a BG image, then the rest of the code in this function can stay the exact same as it was before instead of copy/pasting it.

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
```diff
@@ -304,6 +362,9 @@
                         if (property && typeof property.transform === "function") {
                             that._properties[key] = property.transform(options[key]);
                         } else {
+                            if (key === "bgsrc") {
```
HitmanInWis: Would `property.transform` ever be set on a BG image?  if so, that would cause this code to not be hit.  Is that okay?

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
HitmanInWis: also, any reason why the image version needs to query for multiple elements, but the BG version of this code does not?

vladbailescu:
Thanks for the contribution. I'm closing this for now as we don't have resources to take this on. If you still want to push this through, please:

1. Resolve merge conflicts
2. Address existing review comments
3. Provide usage sample (preferably in the Library)

Issue #727, opened 2019-09-08
Links to non-existing content pages should be handles as invalid


stefanseifert:
currently, the "link handling" logic in the core components works like this:
* if the link URL is a content path pointing to an existing content page, is is externalized (e.g. adding ".html" extension)
* if not the link URL is assumed to be an external URL and inserted unchanged into the markup

actual: in the case that the link URL is a content path (e.g. starting with "/content") that _does not_ point to a valid page (e.g. the target page was deleted or is not yet published), the link handling logic does not detect this and renders the link pointing to an un-externalized and invalid content path.

expected: the link handling should detect invalid internal content page references and handle those links as invalid.

stefanseifert:
i will come up with a PR based on the work of #713

as this is a behavior change the fix was not included in the PR for #713 but instead should be handled as separate issue.

joerghoh:
The LinkRewriter removes links to non-existing resources, so I don't see the need to have this logic in the model already (unless you want to react differently on the fact, that the link target is invalid). 

To be consistent, we would also have to implement such functionality for RTEs and other components.

stefanseifert:
exactly that is the case - the control how to handle an invalid link should be fully to the model, and not left to some post processor parsing. currently there is already markup in each model how to handle if the link is not present, it should handle the same way if the link is invalid. sometimes it does more than just unwrapping the anchor tag.

and as discussed last week in the sling hackathon (see [mailing list](https://lists.apache.org/thread.html/c80093524461d7203fa9799b79ebbf6bfd1bb3f9795865f4aaf3cd4a@%3Cdev.sling.apache.org%3E)) on the long run we want go get rid completely of such post processing as it is done with the rewriter.

and yes, on the long run this functionality should also be implemented in RTE and others ([wcm.io RichText Handler](https://wcm.io/handler/richtext/) does this already for example).

joerghoh:
Thanks for update. Yes, it makes sense to do it. While on the one hand side any decision on the Sling side should not have an impact right here (unless Adobe starts deprecating the LinkRewriter), we can anticipate this decision already.

tinatenatean:
What happens to internal links in pages if those links point to a page that has later been deleted.  I don't recall the author getting alerted to that fact.

BrianAva:
[...]

PR #862, opened 2019-11-27
[Link Handling] Make link references consistent and flexible #713 - v2

stefanseifert:
- introduce Link interface for consistent representation of link target information in all model APIs
- introduce a simple (and internal) LinkHandler implementation to resolve all types of links in all implementations, deprecate Utils methods for the same purpose
- prepare support for link target in the model implementations
- create new version of all models that are affected by the new link handling
  - breadcrumb v2
  - button v2
  - image v3
  - languagenavigation v2
  - list v3
  - navigation v2
  - teaser v2
  - title v3
  - page v3
- the new model versions contain on the new "link" objects in JSON output instead of the old inconsistently named attributes containing the link URL

the PR also contains two small non-functional improvements in separate commits:
- compare JSON results in unit tests with jsonassert to get better reporting about differences if the test fails
- ignore findbugs plugin execution in eclipse

fixes #713


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return Link URL or null if link is invalid.
+     */
+    @Nullable
+    String getURL();
+
+    /**
+     * Map with Attributes for HTML Anchor tag for this links.
+     * This usually also contains the Link URL as <code>href</code> attribute,
+     * but may contain additional attributes like <code>target</code> and others.
+     * @return Map with HTML-specific anchor attributes, or null if link is invalid
```
kwin: may return null but annotated with `@NotNull`

stefanseifert:
the complaints about the broken builds result that this branch is outdated compared to the development/master branch, and there was a new release in between which added new methods to the public APIs missing in this branch.

once the branch is synchronized with the current development branch these failures go away (but please not this is much more effort than doing a git merge because all changes to the AEM component definitions, HTL files etc. have to be re-applied manually to the copies in the new component versions as warned in this mailing list thread
https://groups.google.com/forum/#!msg/aem-core-components-dev/ITKYKRjqRBI/0x76fZMBCQAJ

kwin:
IIUC then https://github.com/adobe/aem-core-wcm-components/pull/862/files#diff-13ccf10ea7dbf29d5ba83f8ee9bad001R43 won't return the mapped URL. That is necessary though for some use cases which are not processed by the LinkRewriter (e.g. https://github.com/adobe/aem-core-wcm-components/blob/master/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/SearchResultServlet.java#L120). I would propose to add a 2nd method which returns the mapped URL or alternatively a boolean flag as argument for `getUrl()` which determines whether the returned link should be mapped or not!

stefanseifert:
this PR is only the first step to introduced centralized link handling logic in core components at all. once this is done there are lots of more features that should make the link handling more robust and flexible (like your previous comment, and like #725 and #727).

i'm waiting until this PR get's merged (which is spanning over 6 months now including the previous one) until investing more efforts...

stefanseifert:
@bpauli thanks for merging this PR!

are you aware that simply merging it is not enough, all files for the new versions of each affected component needs to be updated as well, all changes to the HTL script, dialog definitions, component definitions which where maid since november on the current versions need to be manually re-applied to the 3.0.0 branch (and this has to be done in the future as well until 3.0.0 is released and the main work focuses on the new version of the components).

ky940819:
Since a lot of new component versions are being added for this issue - would it be possible to implement #1254 with these new versions.

The reason given for not changing the behavior was the possibility of clients having component models that do not implement the Component interface; however, with new versions of the components there should be no concern about this.

This would present a good opportunity for optimization, by eliminating the need to adapt the request -> Component twice.

vladbailescu:
@ky940819 , yes, we'll try to make sure each newly introduced component version is using the `id` from it's own model directly.

Issue #823, opened 2019-11-02
[Teaser] Add Pretitle text


thinkstainless:
## Feature Request

**Is your feature request related to a problem? Please describe.**
It's common to see a pre-title, or category tag used in conjunction in teaser type of content. It would be ideal to see an additional dialog field to include this option for the teaser core component. A few examples as shown:

<img width="1266" alt="Screen Shot 2019-11-01 at 8 29 36 PM" src="https://user-images.githubusercontent.com/4670448/68063443-6560d300-fce7-11e9-9623-040e62c5779e.png">

<img width="1281" alt="Screen Shot 2019-11-01 at 8 28 56 PM" src="https://user-images.githubusercontent.com/4670448/68063441-63970f80-fce7-11e9-97e9-8a597b1fe5a5.png">

<img width="311" alt="Screen Shot 2019-11-01 at 8 31 15 PM" src="https://user-images.githubusercontent.com/4670448/68063446-698cf080-fce7-11e9-8a0d-a6a92e51c377.png">

gabrielwalt:
That would make sense, I agree.

brunob15:
I would like to take this issue. Is it ok?

@gabrielwalt @thinkstainless

msagolj:
@brunob15 please go ahead :) thank you.

msagolj:
@brunob15 are you still working on this, otherwise we would like to tackle this issue.

brunob15:
Yes, I didn't have much time but I'm working on this. Feel free to take it if you prefer.

dymitrs:
Hey, found some time and added the missing field. PR above.

vladbailescu:
Merged to `master` in cc0a04168a7605297033c0f0ef0ffee28073228a

PR #873, opened 2019-12-08
[Teaser] Add Pretitle text

dymitrs:
| Minor: New Feature?      | Pretitle in Teaser's dialog

Added pretitle field in Teaser dialog. Added tests & updated README.
Fixing #823


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Teaser.java
```diff
@@ -164,6 +171,16 @@ default boolean isImageLinkHidden() {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Returns this preteaser's title, if one was defined.
```
richardhand: @dymitrs - "Returns this teaser's pretitle, if one was defined"

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Teaser.java
```diff
@@ -164,6 +171,16 @@ default boolean isImageLinkHidden() {
...
+     *
+     * @return the teaser's title or {@code null}
```
richardhand: `@return the teaser's pretitle or {@code null}`

content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/_cq_dialog/.content.xml
```diff
@@ -67,6 +67,19 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
+                                            <pretitleGroup
```
richardhand: @dymitrs - I don't think this needs to be grouped like the title or description, would remove the well wrapper here.

content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/pretitle.html
```diff
@@ -0,0 +1,18 @@
+<!--/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  ~ Copyright 2018 Adobe
```
richardhand: copyright year 2019

richardhand:
@dymitrs we will have to fix the merge conflicts here as there was a recent restructuring of the component library to `ui.apps` and `ui.content` in https://github.com/adobe/aem-core-wcm-components/commit/68c4071bd6f94de711510d07a8eadd408b2f0dd6

content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/_cq_dialog/.content.xml
dymitrs: Sure, I'll remove it

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
```diff
@@ -126,6 +128,11 @@ private void initModel() {
             targetPage = pageManager.getPage(linkURL);
         }
 
+        if (pretitleHidden) {
+            pretitle = null;
```
jckautzmann: I would not change the value of the `pretitle` if it is hidden: the value should be the one defined by the author, even if the field is hidden.

richardhand: I agree, but we already do the same for the title and description and `pretitleHidden` isn't exposed to the template so the only way to check for display is `? teaser.pretitle`. Perhaps it's best to be consistent here with the current behaviour?

jckautzmann: hm... I agree that we should be consistent and have the same behavior as for `title` and `description`.

dymitrs: Yes, I tried to be consistent and have the same behavior as for the other fields.

PR #899, opened 2020-01-28
#897 - Issue with Navigation.getItems() for non-language live copy

### LINKED ISSUE #897: [Navigation] Issue with Navigation.getItems() for non-language live copy
## Bug Report

**Current Behavior**

I create a live copy of a site where the name of the live copy is not a language code. If I access the root page of the live copy the navigation items provided by the navigation component placed on the page will be the children of the source of the live copy and not the children of the actual live copy.
If I access a child page of the live copy then I get the correct items, only the root page has an issue.
If the name of the live copy is a language code then I get the correct items as well.

So the issue is observed only for the root page of a non-language live copy.

**Expected behavior/code**

If I access the root page of a non-language live copy, the navigation items provided by the navigation component placed on the page should be the children of the actual live copy and not the children of the source of the live copy. This will be consistent with current behavior when accessing a child page of the live-copy, which is the expected behavior.

**Environment**

- AEM 6.6 QS build form the 7/1/2020
- Core Components version 2.8.0
- Java(TM) SE Runtime Environment (build 1.8.0_202-b08)

**Possible Solution**

Change line at [1] from 

`if (currentPage.getPath().startsWith(relationship.getTargetPath() + "/")) {`

to be

`if ((currentPage.getPath() + "/").startsWith(relationship.getTargetPath() + "/")) {`

When the incorrect results are observed in the above comparison currentPage.getPath() is the same as relationship.getTargetPath(). The comparison still fails because relationship.getTargetPath() is appended a path separator. So the condition is true when current page is a child pages but is false when the current page is the root page page. 
The condition will be true for root page as well with the proposed solution.

[1] https://github.com/adobe/aem-core-wcm-components/blob/f9faa9c3b21c245d920f59579b02b3d22719c758/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/NavigationImpl.java#L139

LSantha:
| Fixed Issues?            |  Fixes #897
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

Ensure that by accessing the root page of a non-language live copy, the navigation items provided by the navigation component placed on the page should be the children of the actual live copy and not the children of the source of the live copy.

The solution is to make the condition comparing current page path to live copy path to be true when the two paths are equal, in NavigationImp.getItems().

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/NavigationImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/NavigationImpl.java
@@ -136,7 +136,8 @@ public class NavigationImpl implements Navigation {
                 } else if (liveCopiesIterator != null) {
                     while (liveCopiesIterator.hasNext()) {
                         LiveRelationship relationship = (LiveRelationship) liveCopiesIterator.next();
-                        if ((currentPage.getPath() + "/").startsWith(relationship.getTargetPath() + "/")) {
+                        String currentPagePath = currentPage.getPath() + "/";
+                        if (currentPagePath.startsWith(relationship.getTargetPath() + "/")) {
                             Page liveCopyNavigationRoot = pageManager.getPage(relationship.getTargetPath());
                             if (liveCopyNavigationRoot != null) {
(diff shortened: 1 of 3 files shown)


LSantha:
@vladbailescu , I've added unit test for the fix.

LSantha:
@vladbailescu , does it look OK?

Issue #901, opened 2020-01-29
[Tab][Accordion][Carousel] Panel select action not displayed in the editable template editor


rajeevyadav2:
I am trying to author Tabs/Carousel/Accordion in Editable Template. These components are not giving toggling option on editbar when multiple items are added in these components. If I author on pages, it works fine.

richardhand:
Thanks @rajeevyadav2 for the bug report, I was able to replicate the behaviour on AEM latest with Core Components 2.8.0. The problem seems related to the condition at [0] that adds the panel select action to the toolbar. It should probably be made more general, or to also include template editing modes.

[0] - https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/commons/editor/clientlibs/panelselect/js/panelselect.js#L420

rajeevyadav2:
@richardhand : Thank you for confirmation. Could you please confirm when can we expect the fix for this bug.

richardhand:
@rajeevyadav2 - We don't have a set date yet for the next release of the Core Components, but the best way to ensure this fix gets in is to create a pull request. 

Regarding the solution, it should be enough to expand the logic at [0] to check the current layer against the template editor "structure" and "initial" modes.

[0] - https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/commons/editor/clientlibs/panelselect/js/panelselect.js#L421

rajramo61:
@rajeevyadav2 / @richardhand I can pick this item up if you are okay.

richardhand:
@rajeevyadav2 - that would be great! Thanks

rajeevyadav2:
@rajramo61 Please go ahead.

PR #903, opened 2020-01-30
bugfix-901 panelselect.js

rajramo61:
bugfix-901 show toggle select panel in editable template
https://github.com/adobe/aem-core-wcm-components/issues/901

The container components (Accordion/Tab/Carousel) does not provide option to toggle/select panel item(s) when component is authored directly in editable template (structure or initial). However, this works fine when we author the component on the page.

This PR fix addresses above issue by registering panel select option for layer structure and initial as well.

@richardhand and @rajeevyadav2  pls take a look at the PR.


rajeevyadav2:
@rajramo61 : I tested the PR changes and it seems to be working fine.

richardhand:
Thanks @rajramo61 !

PR #912, opened 2020-02-11
Configurable IDs for all components

### LINKED ISSUE #212: [Title] Add ID to support anchors
It would be useful if titles would support anchors so you can link to a specific section of the page.

### LINKED ISSUE #535: [RTC][Architecture] Configurable IDs for all components
**[Request To Comment]** To anybody who intends to use the Core Components, please provide your feedback about the following requirements. We want to make sure that we build a component that fits your needs, so don't hesitate to tell us how that component should work, so that you will be able to use it.

**User Story:**

As a content author, I want to configure specific unique IDs for all form components (button, container, hidden, text, options) so that we can set up more advanced validation

**Acceptance Criteria:**

- Each form component has a configurable text field for the ID
- If the ID field is empty, the auto-generated ID (current behavior) is used
- IDs are not validated to be unique, it is left to the author to set unique, meaningful IDs
- The ID is validated to contain only allowable characters for an html ID (basically, no spaces)

### LINKED ISSUE #588: [Tabs][Carousel][Accordion][Accessibility] Tabs don't have aria-controls and tabpanels don't have aria-labeledby
## Bug Report

**Current Behavior**
Current markup of Tabs, Accordion and Carousel components does not tell the user about the relationship between tabs and corresponding tabpanels. To achieve that tabs should have an `aria-controls` attribute which value equals to corresponding tabpanels id, and tabpanels should have an `aria-labeledby` attribute which value equals to corresponding tabs id.

More information on this can be found here:
https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel
https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html
https://davatron5000.github.io/a11y-nutrition-cards/components/tabs

**Expected behavior/code**
Markup has `aria-controls` and `aria-labeledby` attributes on respectively tab and tabpanel.

**Environment**
- Core Components version 2.3.0

richardhand:
This pull request adds a general approach to component IDs.

### Summary
• Component ID auto-generation implementation.
• Addition of auto-generated component IDs to all component markup.
• Author component ID configuration for certain components (see Author Configuration below) via their edit dialogs.
• Also fixes #588 to improve the accessibility of container components that rely on a component ID.
• Adds test coverage.
• Updates component documentation.
• Trivial update in the Component Library experience fragment example, to use the v2 text component.

### Auto Generation
The implementation for Component ID auto-generation is in the `AbstractComponentImpl`. The IDs are generated in the form [componentName]-[componentResourcePathShaHex], for example, `title-5fba9a0f26`. For components that are referenced from another page, for example in the experience fragment use case, the form is [componentName]-[(referencingComponentResourcePath+componentResourcePath)ShaHex] to prevent duplicate IDs on the page in that case.

### Author Configuration
Each component has an `./id` property available, but it is only exposed in the edit dialog for the following: accordion, carousel, container, content fragment, download, embed, experience fragment, tabs and title. These components are generally ones for which providing an author defined ID can be useful - for writing more meaningful links to page sections. Note, there are no checks for duplicate IDs if provided by an author, so it is their responsibility to ensure a valid HTML document. Form components already have an auto generated ID strategy and will be updated in their next versions to reuse the logic implemented here.

| Fixed Issues?            | `Fixes #535 , Fixes #588, Fixes #212` 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | 
| Documentation Provided   | Yes (code comments and component readmes)
| Any Dependency Changes?  |

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
@@ -48,7 +48,9 @@ public abstract class AbstractComponentImpl implements Component {
                 id = properties.get(Component.PN_ID, String.class);
             }
             if (StringUtils.isEmpty(id)) {
-                id = getGeneratedId();
+                id = generateId();
+            } else {
+                id = StringUtils.replace(StringUtils.normalizeSpace(StringUtils.trim(id)), " ", "-");
             }
         }
         return id;
@@ -60,7 +62,7 @@ public abstract class AbstractComponentImpl implements Component {
         return resource.getResourceType();
     }
 
-    private String getGeneratedId() {
+    private String generateId() {
         String resourceType = resource.getResourceType();
         String prefix = StringUtils.substringAfterLast(resourceType, "/");
         return prefix + "-" + StringUtils.substring(DigestUtils.shaHex(resource.getPath()), 0, 10);
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
@@ -19,11 +19,17 @@ import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ValueMap;
+import org.apache.sling.models.annotations.injectorspecific.ScriptVariable;
 import org.apache.sling.models.annotations.injectorspecific.SlingObject;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.day.cq.wcm.api.Page;
+import com.day.cq.wcm.api.PageManager;
+import com.day.cq.wcm.api.Template;
+import com.day.cq.wcm.api.components.ComponentContext;
 
 import com.adobe.cq.wcm.core.components.models.Component;
 
@@ -36,6 +42,12 @@ public abstract class AbstractComponentImpl implements Component {
 
     @SlingObject
     protected Resource resource;
+
+    @ScriptVariable
+    protected ComponentContext componentContext;
+
+    @ScriptVariable
+    private Page currentPage;
 
     private String id;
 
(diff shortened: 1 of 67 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
```diff
@@ -50,4 +73,44 @@ public String getId() {
...
+                    inTemplate = (template != null && parentContextResource.getPath().startsWith(template.getPath()));
+                    if (inCurrentPage || inTemplate) {
+                        path = parentContextResource.getPath().concat(resource.getPath());
+                        break;
+                    }
+                }
+                parentContext = parentContext.getParent();
+            }
+        }
+        return prefix + "-" + StringUtils.substring(DigestUtils.shaHex(path), 0, 10);
```
lewandowskidawid: don't we generate the same ID when two the same XF will be placed on the same page?

richardhand: @lewandowskidawid - no we don't, we check for components referenced from other pages (including experience fragment pages) and in that case we hash a combination of the path of the context i.e. the referencing component and the path of the referenced component.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
```diff
@@ -50,4 +73,44 @@ public String getId() {
...
+     */
+    private String generateId() {
+        String resourceType = resource.getResourceType();
+        String prefix = StringUtils.substringAfterLast(resourceType, "/");
+        String path = resource.getPath();
+        PageManager pageManager = currentPage.getPageManager();
+        Page containingPage = pageManager.getContainingPage(resource);
+        Template template = currentPage.getTemplate();
+        Boolean inCurrentPage = (containingPage != null && StringUtils.equals(containingPage.getPath(), currentPage.getPath()));
+        Boolean inTemplate = (template != null && path.startsWith(template.getPath()));
```
jckautzmann: check for path != null

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
jckautzmann: According to the javadocs [0] DigestUtils.shaHex is deprecated:  sha1Hex(String) should be used instead.
[0] https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/digest/DigestUtils.html#shaHex-java.lang.String-

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
richardhand: Path here is from `Resource#getPath`, which isn't nullable so I think we're safe?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
jckautzmann: Taking the first 10 characters of the hash might generate the same ID for two different component paths.

richardhand: Good spot!

richardhand: added a note on the probability of a collision for 10 characters here: https://github.com/adobe/aem-core-wcm-components/blob/33c6bcc0cd4881abaea366a3673dc18f2e1ad3bb/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java#L87

richardhand: Updated in f933ddfa9184ccf6f2b18efcc81b0b6a3acb592b

PR #915, opened 2020-02-13
[Embed] Accessibility issues flagged for Youtube

richardhand:
- adds generic aria label to the YouTube embeddable iframe.

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/embed/v1/embed/embeddable/youtube/youtube.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/embed/v1/embed/embeddable/youtube/youtube.html
@@ -18,4 +18,5 @@
         height="${properties.youtubeHeight || 390 @ context='scriptString'}"
         src="${'https://www.youtube.com/embed/{0}?origin={1}://{2}' @ format=[properties.youtubeVideoId, request.scheme, request.serverName]}"
         frameborder="0"
-        allowfullscreen></iframe>
+        allowfullscreen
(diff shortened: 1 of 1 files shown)


content/src/content/jcr_root/apps/core/wcm/components/embed/v1/embed/embeddable/youtube/youtube.html
```diff
@@ -18,4 +18,5 @@
         height="${properties.youtubeHeight || 390 @ context='scriptString'}"
         src="${'https://www.youtube.com/embed/{0}?origin={1}://{2}' @ format=[properties.youtubeVideoId, request.scheme, request.serverName]}"
         frameborder="0"
-        allowfullscreen></iframe>
+        allowfullscreen
+        aria-label="${'YouTube Video' @ i18n}"></iframe>
```
abhishekagg19: Will it make sense to make title/aria-label as authorable as well? In case it is blank, it can fallback to default value of YouTube Video

richardhand: I wasn't really sure on that one, from a usability point of view it would be better not to have it authorable as it's not really a main concern when configuring the embed, but if from an accessibility standpoint this default label is weak, then maybe we should.

richardhand: We can also start with this default and then have a later improvement to make it authorable if there is a need for it.

PR #917, opened 2020-02-14
[WIP] [PoC] Feature/datalayer inlinescript poc

bartoszglow:
Proof of concept for Data Layer integration with core components using inline scripts. It is still work in progress and concept might be changed in the future.


content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/adobe-client-data-layer.min.js
```diff
@@ -0,0 +1,2 @@
....iterator]();!(n=(a=i.next()).done);n=!0){var u=_slicedToArray(a.value,2),c=u[0],s=u[1];if(y(s,t))return c}}catch(t){r=!0,o=t}finally{try{n||null==i.return||i.return()}finally{if(r)throw o}}}return-1}return{register:function(t){var e=t.event;Object.prototype.hasOwnProperty.call(f,e)?-1===r(t)&&f[t.event].push(t):f[t.event]=[t]},unregister:function(t){var e=t.event;if(Object.prototype.hasOwnProperty.call(f,e))if(t.handler||t.scope||t.path){var n=r(t);-1<n&&f[e].splice(n,1)}else f[e]=[]},triggerListeners:function(i){(function(t){var e=[],n=t.config;t.type===p.itemType.DATA?e.push(p.dataLayerEvent.CHANGE):t.type===p.itemType.EVENT&&(n.event!==p.dataLayerEvent.CHANGE&&e.push(n.event),e.push(p.dataLayerEvent.EVENT),n.data&&e.push(p.dataLayerEvent.CHANGE));return e})(i).forEach(function(t){if(Object.prototype.hasOwnProperty.call(f,t)){var e=!0,n=!1,r=void 0;try{for(var o,a=f[t][Symbol.iterator]();!(e=(o=a.next()).done);e=!0){u(o.value,i,!1)}}catch(t){n=!0,r=t}finally{try{e||null==a.return||a.return()}finally{if(n)throw r}}}})},triggerListener:function(t,e){u(t,e,!0)}}},e.exports=r},{"./DataLayerConstants":7,"lodash.clonedeep":1,"lodash.get":2,"lodash.has":3,"lodash.isequal":4}]},{},[6]);
```
richardhand: This should be delivered by a launch extension, but it's here currently for the PoC

content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/page.html
```diff
@@ -19,6 +19,11 @@
       data-sly-use.footer="footer.html"
       data-sly-use.redirect="redirect.html">
     <head data-sly-call="${head.head @ page = page}"></head>
+
+    <script type="text/javascript">
+        dataLayer.push(${page.dataLayerJson @ context='unsafe'});
```
richardhand: we should probably check here for the dataLayer object:
```
if (window.dataLayer) {
   window.dataLayer.push(${page.dataLayerJson @ context='unsafe'});
}
```

content/src/content/jcr_root/apps/core/wcm/components/page/v2/page/page.html
```diff
@@ -19,6 +19,11 @@
       data-sly-use.footer="footer.html"
       data-sly-use.redirect="redirect.html">
     <head data-sly-call="${head.head @ page = page}"></head>
+
+    <script type="text/javascript">
```
richardhand: Eventually this script will need to be conditionally added based on whether the data layer is enabled, or not. Could be something like:
```
<script data-sly-test=${page.dataLayerJson} type="text/javascript">
   // my script
</script>
```

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/image.html
```diff
@@ -45,4 +45,8 @@
     <span class="cmp-image__title" itemprop="caption" data-sly-test="${!image.displayPopupTitle && image.title}">${image.title}</span>
     <meta itemprop="caption" content="${image.title}" data-sly-test="${image.displayPopupTitle && image.title}">
 </div>
-<sly data-sly-call="${templates.placeholder @ isEmpty = !image.src, classAppend = 'cmp-image cq-dd-image'}"></sly>
\ No newline at end of file
+<sly data-sly-call="${templates.placeholder @ isEmpty = !image.src, classAppend = 'cmp-image cq-dd-image'}"></sly>
+
+<script type="text/javascript">
```
richardhand: Would place the script before the placeholder call. Perhaps it could even be the last child of the component `div` ?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -197,4 +199,13 @@ public boolean hasCloudconfigSupport() {
         }
         return hasCloudconfigSupport;
     }
+
+    @Override
+    public String getDataLayerJson() {
+        JsonObjectBuilder pageData = Json.createObjectBuilder();
+        pageData.add("id", currentPage.getPath());
```
richardhand: TODO define the data format and keys to populate per component.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -299,4 +299,8 @@ default String getExportedType() {
     default boolean isDecorative() {
         throw new UnsupportedOperationException();
     };
+
+    default String getDataLayerJson() {
```
richardhand: This might make more sense in the base `Component` class eventually

content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/adobe-client-data-layer.min.js
jckautzmann: I think that the data layer script should be on a CDN and that we should stay flexible and offer both options for including it on the page:
- through the page component if the data layer is enabled (default)
- though a launch extension
Some customers might want to use the data layer but might not want to use launch.

richardhand:
@bartoszglow - There are build problems due to the v1 image and page components not having implementations of `getDataLayerJson`, for now we can simply add a `JsonIgnore` annotation in the interface for that method to get the build passing. To be decided whether the data layer JSON should be available in the standard JSON output for the component, but would likely results in some repetition.

jckautzmann:
We are investigating inline scripts because it is assumed to be faster than client libs. Before moving forward we should validate this assumption by comparing the performance in both cases.

PR #926, opened 2020-02-20
Feature/experience fragment v2

niekraaijmakers:
Experience Fragment Variation V2: Basically the same as the original XFV. 
In addition:

Exports underlying contents of the experience fragment immediately in the JSON output, using the modelfactory. 

This can be used by the javascript components to render the fragment without performing an AJAX request. 
A internal PR is created for the react spa editor to support the experience fragment using on this resource type.

Extensive tests added, along with basic javadoc.

No breaking changes or existing code modified.


gabrielwalt:
Thank you @niekraaijmakers for your PR, it looks nice! We'll start reviewing it asap.

Can you provide some context why you need CSS class names to be included in the JSON output, and what these class names?

niekraaijmakers:
The CSS class is used by the react (and soon to be angular as well) component to output around the wrapper DIV. This so that there is some control over it using the properties.

habansal:
FYI: 
This PR has been closed in-favour of https://github.com/adobe/aem-core-wcm-components/pull/941 (Adobe CLA thing)

Issue #942, opened 2020-03-06
Deep-linking capability on Accordion and Tabs components


rajeevyadav2:
## Feature Request

**Is your feature request related to a problem? Please describe.**
Content team should be able to anchor URLs to specific item/panel on these components.

**Describe the solution you'd like**
Each Item/Panel in accordion should have unique id.
Javascript should check the anchor and open the specific panel/item in the component and scroll the window to bring user on the anchored panel.

**Are there alternatives?**
No

**Documentation**
If you can, explain how users will be able to use this and possibly write out a version of the docs.
Maybe a screenshot or design?

gabrielwalt:
+1 this is a feature we've also been thinking about for long.

We did a first step towards this feature with #920 ("Configurable IDs for all components"). Now a second step would be to adjust the JS client library of the accordion and tab components to automatically open the corresponding panel when that panel's ID is in the URL hash.

If you'd like to tackle this, @rajeevyadav2, your contribution is welcome!

rajeevyadav2:
@gabrielwalt : Thank you for the information. Do I need to open RTC first before starting working on it or else I should fork the repository and start working on changes?

rajramo61:
@gabrielwalt : I will be working on this issue with @rajeevyadav2 .

rajeevyadav2:
@gabrielwalt : I was looking into ID implementation in tab and accordion component in development branch and see issues with the approach taken.

As per the approach, each item id is based on index number of the item. Now lets say some item is at 3rd index and marketing team configures campaign urls in various channels to point anchor to 3rd item. Post this if any reordering is done in the item, campaign urls will still be pointing to 3rd item which may not be correct. I feel the Id should always remain same after it has been generated and should not change once it moves from one location to another. This will ensure that anchor always points to correct content.

Do let me know your thoughts on this.

bpauli:
@rajeevyadav2 Indeed, the index of the accordion can always change. Do you think it would be sufficient if we use the `ListItem#getName()` as a suffix for the accordion item id?

jckautzmann:
@bpauli Ideally all links (all linkable items) should have a unique ID, similar to the component ID. It will be needed e.g to identify a link in the data layer.

bpauli:
[...]

PR #946, opened 2020-03-11
Deep-linking capability on Accordion component #942

rajramo61:
| Fixed Issues?            | Fixes https://github.com/adobe/aem-core-wcm-components/issues/942
| Patch: Bug Fix?          | Feature enhancement
| Tests Added + Pass?      | No Tests added
Deep linking capability for accordion
Generating item id with component-id and item-name instead of item index
Moved id attribute from inner div to the wrapper dive of each item

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
@@ -22,13 +22,13 @@
          data-sly-repeat.item="${accordion.items}"
          class="cmp-accordion__item"
          data-cmp-hook-accordion="item"
-         id="${accordion.id}-panel-${item.name  @ context='uri'}"
+         id="${accordion.id}-item-${item.name  @ context='uri'}"
          data-cmp-expanded="${item.name in accordion.expandedItems}">
         <h3 data-sly-element="${accordion.headingElement @ context='elementName'}"
             class="cmp-accordion__header">
-            <button id="${accordion.id}-button-${itemList.index}"
+            <button id="${accordion.id}-button-${item.name  @ context='uri'}"
                     class="cmp-accordion__button${item.name in accordion.expandedItems ? ' cmp-accordion__button--expanded' : ''}"
-                    aria-controls="${accordion.id}-panel-${itemList.index}"
+                    aria-controls="${accordion.id}-panel-${item.name  @ context='uri'}"
                     data-cmp-hook-accordion="button">
                 <span class="cmp-accordion__title">${item.title}</span>
                 <span class="cmp-accordion__icon"></span>
@@ -36,9 +36,10 @@
         </h3>
         <div data-sly-resource="${item.name @ decorationTagName='div'}"
              data-cmp-hook-accordion="panel"
+             id="${accordion.id}-panel-${item.name  @ context='uri'}"
              class="cmp-accordion__panel${item.name in accordion.expandedItems ? ' cmp-accordion__panel--expanded' : ' cmp-accordion__panel--hidden'}"
              role="region"
-             aria-labelledby="${accordion.id}-button-${itemList.index}"></div>
+             aria-labelledby="${accordion.id}-button-${item.name  @ context='uri'}"></div>
     </div>
     <sly data-sly-resource="${resource.path @ resourceType='wcm/foundation/components/parsys/newpar', appendPath='/*', decorationTagName='div', cssClassName='new section aem-Grid-newComponent'}"
          data-sly-test="${(wcmmode.edit || wcmmode.preview) && accordion.items.size < 1}"></sly>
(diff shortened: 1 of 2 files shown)


content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
```diff
@@ -109,15 +109,42 @@
...
                 that._elements["panel"] = Array.isArray(that._elements["panel"]) ? that._elements["panel"] : [that._elements["panel"]];
 
-                if (that._properties.singleExpansion) {
-                    var expandedItems = getExpandedItems();
-                    // no expanded item annotated, force the first item to display.
-                    if (expandedItems.length === 0) {
-                        toggle(0);
+                // Expand the item based on deep-link-id if it matches with any existing accordion item id
+                var deepLinkIdIndex = -1;
+                if (config.deepLinkId) {
```
bpauli:
```suggestion
                if (config.deepLinkId && config.deepLinkId.substring("-panel-") !== -1) {
```

rajeevyadav2:
> Other than the small suggestion LGTM! @rajramo61 Thank you!

@rajramo61 : Thank you. 👍

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
rajeevyadav2: @bpauli : Do we need to make this change or else you are taking care of it?

rajramo61:
@gabrielwalt This PR is only for Accordion. I will create another PR for Tab component. So, I modified the title for this PR.

rajramo61:
> Other than the small suggestion LGTM! @rajramo61 Thank you!

Thanks @bpauli, I have accepted the suggestion. I think it's ready to be merged.

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
```diff
@@ -35,7 +36,6 @@
         </h3>
         <div data-sly-resource="${item.name @ decorationTagName='div'}"
              data-cmp-hook-accordion="panel"
-             id="${accordion.id}-panel-${itemList.index}"
```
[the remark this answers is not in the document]
rajramo61: Thank you @richardhand for bringing accessibility point here which is very important. I am not an expert in this area. Thanks for providing the link, I read through the link documentation. It was really helpful to understand your point. I am going to make these changes to keep the accessibility intact.

rajramo61:
@richardhand I have made the changes as per the your suggestion. Please review the changes.

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
```diff
@@ -111,7 +111,7 @@
 
                 // Expand the item based on deep-link-id if it matches with any existing accordion item id
                 var deepLinkIdIndex = -1;
-                if (config.deepLinkId && config.deepLinkId.substring("-panel-") !== -1) {
+                if (config.deepLinkId && config.deepLinkId.substring("-item-item") !== -1) {
```
richardhand: Shouldn't this be `-item` rather than `-item-item`?

rajramo61: @richardhand As we are adding item name at the end of the id. All of the elements, button, panel and item (the wrapper div) all has item in common.
So, to avoid any confusion I added the 'item' which is added on wrapper div and the item coming from name to make it unique.
I am adding a screen shot for how the ids look like.

image

jckautzmann:
IMHO we need a broader concept about IDs as other clickable items (like links) need to have an ID. For example: a call-to-action link in a teaser needs to have an ID that can be tracked with the data layer.
We should wait with merging this PR until we have such a design in place.

@gabrielwalt WDYT?

rajeevyadav2:
@jckaurzmann : Agreed. There need to be a consistent way of id generation. May be separate out is generation in separate service and then use it in component and item models.

rajramo61:
> IMHO we need a broader concept about IDs as other clickable items (like links) need to have an ID. For example: a call-to-action link in a teaser needs to have an ID that can be tracked with the data layer.
> We should wait with merging this PR until we have such a design in place.
> 
> @gabrielwalt WDYT?

True, right now the ID generation logic is sitting in abstract class (tied up with resource path) and can be added to those where we inherit from the abstract class.
To make it widely available, like @rajeevyadav2 suggested; we need to move out this logic as service and can easily be consumed by any other class with association (has a) relationship to this service.
@jckautzmann / @gabrielwalt What are your thoughts?

jckautzmann:
> True, right now the ID generation logic is sitting in abstract class (tied up with resource path) and can be added to those where we inherit from the abstract class.
> To make it widely available, like @rajeevyadav2 suggested; we need to move out this logic as service and can easily be consumed by any other class with association (has a) relationship to this service.
> @jckautzmann / @gabrielwalt What are your thoughts?

@rajramo61 I created a new ticket to track it:
https://github.com/adobe/aem-core-wcm-components/issues/953
Let's continue discussing this very interesting topic there!

PR #947, opened 2020-03-12
Use SHA-256 for digest instead of deprecated SHA-1

vladbailescu:


vladbailescu:
@pauli, done in https://github.com/adobe/aem-core-wcm-components/pull/947/commits/ea759782d06548beb69cf32154e99024b94c2964

Issue #945, opened 2020-03-11
[Navigation] Navigation doesn't show original title of redirecting page


3ch023:
## Bug Report

**Current Behavior**
For redirecting page Navigation shows target page title instead of the original page title.
When PageA redirects to PageB it is displayed in the Navigation as a PageB.
This leads to unexpected behavior when the Author configures the parent page to redirect to a child page.
 
Without redirect:
- Parent
    - Child

With redirect:
- Child
    - Child

**Expected behavior/code**
It should be possible to reflect real Site Structure in Navigation, even when using a redirect.
Redirecting page title is displayed in Navigation, while the link is a redirect target. 

**Environment**
- AEM 6.5.3.0
- Core Components version 2.8.0
- JRE 1.8.0

**Possible Solution**

One of the solutions is to add a checkbox to Navigation component dialog, which will prevent Navigation to treat the redirecting page as a redirect target. Something like "Disable Shadowing".
When the checkbox is checked the Navigation will display:
- Parent 
    - Child

Where parent link target is Child.

**Additional context / Screenshots**
This behavior happens in PageListItemImpl.java, line 45.
I assume it's needed for [Shadow Redirect Pages](https://docs.adobe.com/content/help/en/experience-manager-core-components/using/components/navigation.html#shadow-structure) to work. 

```
public PageListItemImpl(@NotNull SlingHttpServletRequest request, @NotNull Page page) {
        this.request = request;
        this.page = page;
        Page redirectTarget = getRedirectTarget(page);
        if (redirectTarget != null && !redirectTarget.equals(page)) {
            this.page = redirectTarget;
        }
    }
```

vladbailescu:
Thank you for reporting this issue. We would greatly appreciate a contribution in form of a pull request for this as we currently don't have capacity to address this request in the near future.

PR #952, opened 2020-03-13
[Navigation] Navigation doesn't show original title of redirecting page #945

3ch023:
| Fixed Issues?            | [#945](https://github.com/adobe/aem-core-wcm-components/issues/945) 
| Patch: Bug Fix?          |
| Any Dependency Changes?  |

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/BreadcrumbImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/BreadcrumbImpl.java
@@ -36,6 +36,9 @@ import com.adobe.cq.wcm.core.components.models.NavigationItem;
 import com.day.cq.wcm.api.Page;
 import com.day.cq.wcm.api.designer.Style;
 
+import static com.adobe.cq.wcm.core.components.internal.models.v1.PageListItemImpl.PN_DISABLE_SHADOWING;
+import static com.adobe.cq.wcm.core.components.internal.models.v1.PageListItemImpl.PROP_DISABLE_SHADOWING_DEFAULT;
+
 @Model(adaptables = SlingHttpServletRequest.class,
        adapters = {Breadcrumb.class, ComponentExporter.class},
        resourceType = {BreadcrumbImpl.RESOURCE_TYPE_V1, BreadcrumbImpl.RESOURCE_TYPE_V2})
@@ -48,7 +51,6 @@ public class BreadcrumbImpl implements Breadcrumb {
     protected static final boolean PROP_SHOW_HIDDEN_DEFAULT = false;
     protected static final boolean PROP_HIDE_CURRENT_DEFAULT = false;
     protected static final int PROP_START_LEVEL_DEFAULT = 2;
-    private static final boolean PROP_DISABLE_SHADOWING = false;
 
     @ScriptVariable
     private ValueMap properties;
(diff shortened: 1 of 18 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageListItemImpl.java
```diff
@@ -107,4 +108,7 @@ private Page getRedirectTarget(@NotNull Page page) {
         return result;
     }
 
+    private boolean isShadowingEnabled(@NotNull SlingHttpServletRequest request) {
```
vladbailescu: I see you defined both a policy config and a dialog property for disabling shadowing, but you're only using the latter. What's the purpose of defining the policy config property? Am I missing something?

Also, I would prefer you inject this property at component (navigation/breadcrumb) level and pass it on to this logic.

When navigating such a long path of methods it might be safer to null-check.

3ch023: Hi @vladbailescu, thanks for a fast code review :) 

Good point, I will add a fallback to currentStyle like it's done for rest of properties: 
`isShadowingEnabled = !properties.get(PN_DISABLE_SHADOWING, currentStyle.get(PN_DISABLE_SHADOWING, false));`

  I was planning to put the property on Navigation component level initially, but then thought it might be also helpful for other components, that use PageListItemImpl (e.g. List).  Now when I look at it a second time, I see it's more clear way to keep it on a component level. I’ll move it.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/BreadcrumbImpl.java
```diff
@@ -48,6 +48,7 @@
     protected static final boolean PROP_SHOW_HIDDEN_DEFAULT = false;
     protected static final boolean PROP_HIDE_CURRENT_DEFAULT = false;
     protected static final int PROP_START_LEVEL_DEFAULT = 2;
+    private static final boolean PROP_DISABLE_SHADOWING = false;
```
vladbailescu: Please add `_DEFAULT` to constant name so it's easier to understand what it's used for!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
```diff
@@ -72,6 +72,12 @@
      */
     String PN_STRUCTURE_DEPTH = "structureDepth";
 
+    /**
+     * Name of the resource property that for redirecting pages will indicate if original page or redirect target page should be returned.
+     * Property is false by default.
+     * If `true` - original page is returned. If `false` or not configured - redirect target page.
+     */
+    String PN_DISABLE_SHADOWING = "disableShadowing";
```
vladbailescu: I would rather we only define the property once and reuse it in other components.

3ch023: I am not sure where can I move it. There is no dedicated constants class yet, should I create something like ComponentConstants or ComponentConfigConstants?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/BreadcrumbImpl.java
```diff
@@ -36,6 +36,9 @@
 import com.day.cq.wcm.api.Page;
 import com.day.cq.wcm.api.designer.Style;
 
+import static com.adobe.cq.wcm.core.components.internal.models.v1.PageListItemImpl.PN_DISABLE_SHADOWING;
```
dymitrs: It's good to not use static imports.
You could import your PageListItem and refer to the your variable by
PageListItem.PN_DISABLE_SHADOWING.

3ch023: Hi, thank you for the comment. You are right, this is not a good case for static import, I will update the request.

PR #958, opened 2020-03-25
Check for null in PanelContainerItemImpl

phornig:
In `PanelContainerItemImpl` the panel title ist read from the resource value map property `cq:panelTitle` with `jcr:title` as fallback, but `jcr:title` may not exist as well and newer versions of the Sling API don't allow a null value for the default value in `ValueMap.get(@NotNull String name, @NotNull T defaultValue)`, so `jcrTitle` needs to be checked for `null` and we have to fallback to `ValueMap.get(@NotNull String name, @NotNull Class<T> type)` to get the current behavior, i.e. return a `null` title.

| Fixed Issues?            | 

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PanelContainerItemImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PanelContainerItemImpl.java
@@ -29,9 +29,6 @@ public class PanelContainerItemImpl extends ResourceListItemImpl implements List
 
     public PanelContainerItemImpl(@NotNull SlingHttpServletRequest request, @NotNull Resource resource) {
         super(request, resource);
-        ValueMap valueMap = resource.adaptTo(ValueMap.class);
-        if (valueMap != null) {
-            title = valueMap.get(PN_PANEL_TITLE, valueMap.get(JcrConstants.JCR_TITLE, ""));
-        }
+        title = resource.getValueMap().get(PN_PANEL_TITLE, resource.getValueMap().get(JcrConstants.JCR_TITLE, ""));
     }
 }
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PanelContainerItemImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PanelContainerItemImpl.java
@@ -32,7 +32,11 @@ public class PanelContainerItemImpl extends ResourceListItemImpl implements List
         ValueMap valueMap = resource.adaptTo(ValueMap.class);
         if (valueMap != null) {
             String jcrTitle = valueMap.get(JcrConstants.JCR_TITLE, String.class);
-            title = valueMap.get(PN_PANEL_TITLE, jcrTitle);
+            if (jcrTitle == null) {
+                title = valueMap.get(PN_PANEL_TITLE, String.class);
+            } else {
+                title = valueMap.get(PN_PANEL_TITLE, jcrTitle);
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PanelContainerItemImpl.java
```diff
@@ -32,7 +32,11 @@ public PanelContainerItemImpl(@NotNull SlingHttpServletRequest request, @NotNull
         ValueMap valueMap = resource.adaptTo(ValueMap.class);
         if (valueMap != null) {
             String jcrTitle = valueMap.get(JcrConstants.JCR_TITLE, String.class);
-            title = valueMap.get(PN_PANEL_TITLE, jcrTitle);
+            if (jcrTitle == null) {
```
kwin: I think a simple `title = valueMap.get(PN_PANEL_TITLE, valueMap.get(PN_PANEL_TITLE, String))` should suffice here and is way easier to read

kwin:
Is there a real exception caused by this? Although the parameter is annotated as `NotNull` it falls back to a simple `get(JcrConstants.JCR_TITLE)` in case `null` is given as 2nd argument (https://github.com/apache/sling-org-apache-sling-api/blob/e6321c53fc404fc9af80dbe3f0670c139670ee57/src/main/java/org/apache/sling/api/wrappers/ValueMapDecorator.java#L78)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PanelContainerItemImpl.java
phornig: That would be the same as before, just in one line, and with the same problem, `ValueMap.get(@NotNull String name, @NotNull Class<T> type)` is `@Nullable`, but `defaultValue` in `ValueMap.get(@NotNull String name, @NotNull T defaultValue)` is `@NotNull`.

phornig:
No, there is no exception, but static code analysis fails.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PanelContainerItemImpl.java
kwin: Right, but I think in this case just returning `""` instead of `null` does not do any harm (HTL treats both values the same), so I would suggest simplifying to
```
public PanelContainerItemImpl(@NotNull SlingHttpServletRequest request, @NotNull Resource resource) {
        super(request, resource);
        title = resource.getValueMap().get(PN_PANEL_TITLE, resource.getValueMap().get(JcrConstants.JCR_TITLE, ""));
    }
```
For a longer discussion around those nulllness annotations look at https://jira.apache.org/jira/browse/SLING-6609.

phornig: I agree, returning empty string would be preferable. I updated the pull request.

vladbailescu:
@phornig , this is blocked because you did not sign the CLA.

ky940819:
> No, there is no exception, but static code analysis fails.

There is no exception in most cases, not all ValueMap implementations handle a null default value. 
As part of #1029 (which deals with other possible NPEs) I have offered an alternative solution to this issue. 

My preference for dealing with default values in ValueMap.get() is that if the default value isn't a constant non-null value (i.e. it either is possibly null, or requires some work to calculate) then i wrap the call in an optional:
```JAVA
Optional.ofNullable(valueMap.get(PN_SOMETHING, String.class))
    .orElseGet(() -> {
        // calculate the default here, it will only be executed if the property doesn't exist
    });
```

ky940819:
I think this PR can be closed as the issue was resolved as part of the merged #1029

PR #1002, opened 2020-05-11
Move core-components from /apps to /libs for AEM as a cloud

bpauli:
- add new cloud profile to content module


content/pom.xml
```diff
@@ -193,6 +198,67 @@
...
+        <profile>
+            <id>cloud</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-resources-plugin</artifactId>
```
kwin, line 210: Instead of copying I would rather use ‘prefix’ (http://jackrabbit.apache.org/filevault-package-maven-plugin/generate-metadata-mojo.html#prefix) together with adjusting http://jackrabbit.apache.org/filevault-package-maven-plugin/package-mojo.html#jcrRootSourceDirectory

kwin:
The filevault maven plug-in 1.1.4 has been released meanwhile.

PR #1003, opened 2020-05-11
Alternative package using final name and multiple executions

vladbailescu:


bpauli:
adding the approach with executions to https://github.com/adobe/aem-core-wcm-components/pull/1002 and closing this PR

PR #1004, opened 2020-05-11
Refactored datalayer implementation

vladbailescu:
Refactored the datalayer implementation for a better structure and simpler serializations.


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Component.java
```diff
@@ -50,12 +56,14 @@ default String getId() {
...
-     * @return {@link DataLayer} object associated with the component
+     * @return {@link ComponentDataModel} object associated with the component
      *
      * @since com.adobe.cq.wcm.core.components.models 12.14.0
      */
     @Nullable
-    default DataLayer getDataLayer() {
+    @JsonProperty("dataLayer")
+    @JsonSerialize(using = ComponentDataModelSerializer.class)
+    default ComponentDataModel getComponentDataModel() {
```
vladbailescu, line 66: We could keep the previous `getDataLayer()` method name.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Component.java
```diff
@@ -50,12 +56,14 @@ default String getId() {
...
         throw new UnsupportedOperationException();
```
vladbailescu, line 67: To avoid needing to bump up versions for all component we could replace this with a default of `return null`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/ComponentDataModel.java
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * Returns the JSON string of the component's properties used in the data layer
+     *
+     * @return JSON string
+     *
+     * @since com.adobe.cq.wcm.core.components.models.datalayer 1.0.0
+     */
+    @JsonIgnore
+    @Nullable
+    default String getString() {
```
jckautzmann, line 136: I'd rename it `getJsonString()`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/ImageDataModel.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Returns the asset associated with the image.
+     *
+     * @return Asset data model
+     *
+     * @since com.adobe.cq.wcm.core.components.models.datalayer 1.0.0
+     */
+    @JsonProperty("asset")
+    default AssetDataModel getImage() {
```
jckautzmann, line 35: I'd rename it: `getAssetData()`

jckautzmann:
LGTM in general. I would get rid of `Model` suffix in Java classes, to be consistent with the way we name our model classes (e.g. `Image`, not `ImageModel`):
AssetDataModel -> AssetData
ContainerDataModel -> ContainerData
ComponentDataModel -> ComponentData
...
Same for Impl:
PageDataModelImpl -> PageDataImpl

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
```diff
@@ -17,13 +17,13 @@
      id="${accordion.id}"
      class="cmp-accordion"
      data-cmp-is="accordion"
-     data-cmp-data-layer="${accordion.dataLayer.string}"
+     data-cmp-data-layer="${accordion.componentDataModel.string}"
```
jckautzmann, line 20: With the changes that I suggested, it would become:
`accordion.componentData.json` (or `accordion.componentData.jsonString`)
Even better and more concise notation would be:
`accordion.data.json`

vladbailescu: Done in 2d1846a, now we use `component.data.json`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/ImageDataModel.java
vladbailescu: Done in 2d1846a

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/ComponentDataModel.java
vladbailescu: Done in 2d1846a, it's `getJson()` now

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Component.java
vladbailescu: Added `return null` in 5dbfd4e

PR #1005, opened 2020-05-12
Tentative XDM JSON properties

vladbailescu:


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/AssetDataModel.java
```diff
@@ -47,6 +49,7 @@ default String getId() {
      *
      * @since com.adobe.cq.wcm.core.components.models.datalayer 1.0.0
      */
+    @JsonProperty("repo")
```
jckautzmann, line 52: Is `repo` the asset URL or the URL of the repository (store) where the asset is stored?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/ContainerDataModel.java
```diff
@@ -29,6 +31,7 @@
      *
      * @since com.adobe.cq.wcm.core.components.models.datalayer 1.0.0
      */
+    @JsonProperty("xdm:items")
```
jckautzmann, line 34: IIUC `xdm:items` represents all the items of the container which is different from the displayed items.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/ImageDataModel.java
```diff
@@ -31,7 +31,7 @@
      *
      * @since com.adobe.cq.wcm.core.components.models.datalayer 1.0.0
      */
-    @JsonProperty("asset")
+    @JsonProperty("image")
     default AssetDataModel getImage() {
```
jckautzmann, line 35: I'd call this method `getAssetData()`

vladbailescu: Done in 2d1846aedbd65f6c9483a948382dc9ee13d9588b

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/AssetDataModel.java
vladbailescu: That was supposed to be `repo:path` (https://github.com/adobe/xdm/blob/master/docs/reference/external/repo/asset.schema.md#repopath)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/ContainerDataModel.java
vladbailescu: You're right, we only list the active items. I will switch this back to our custom property name.

vladbailescu: Done in f71cfef

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/AssetDataModel.java
vladbailescu: Done in f71cfef

PR #1010, opened 2020-05-19
[Image] Javascript exception when linkURL is removed from the dialog

### LINKED ISSUE #1009: [Image] Javascript exception when linkURL is removed from the dialog
## Bug Report

**Current Behavior**
If you extend the Image component and you remove the linkURL from the dialog (with sling:hideChildren for example), the dialog javascript will throw an "Uncaught TypeError: Cannot read property 'setDisabled' of undefined" exception when the dialog loads and when you check/uncheck the isDecorative checkbox.

**Expected behavior/code**
You shouldn't get any javascript exceptions and the rest of the functionality should work as expected even if you remove the linkURL.

**Environment**
- AEM 6.5.4
- 2.8.0
- `Java(TM) SE Runtime Environment (build 1.8.0_241-b07)`

**Possible Solution**
I will create a PR, there is a simple fix, to check the $linkURLField.length before trying to do ".setDisabled" in toggleAlternativeFieldsAndLink function (apps/core/wcm/components/image/v2/image/clientlibs/editor/js/image.js, line 121)

A workaround is possible by adding an extra "hidden" class to the linkURL in the wrapperClass property.

luciannicolaescu:
| Fixed Issues?            | Fixes #1009

Adds a check that the $linkURLField jquery element is present in the dialog before trying to use the element properties (in this case the .setDisabled function). This way you can remove the linkURL from the dialog without breaking the functionality.

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/editor/js/image.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/editor/js/image.js
@@ -118,7 +118,9 @@
                 $altGroup.show();
                 $linkURLGroup.show();
             }
-            $linkURLField.adaptTo("foundation-field").setDisabled(checkbox.checked);
+            if ($linkURLField.length) {
+                $linkURLField.adaptTo("foundation-field").setDisabled(checkbox.checked);
+            }
             altTuple.hideTextfield(checkbox.checked);
             if (fileReference) {
                 altTuple.hideCheckbox(checkbox.checked);


content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/editor/js/image.js
```diff
@@ -118,7 +118,9 @@
                 $altGroup.show();
                 $linkURLGroup.show();
             }
-            $linkURLField.adaptTo("foundation-field").setDisabled(checkbox.checked);
+            if ($linkURLField.length) {
```
jckautzmann, line 121: how about checking if the field exists?
`if ($linkURLField) {`

luciannicolaescu: It's a jQuery object so it will always exist, except the length will be 0.

jckautzmann:
@luciannicolaescu Thx for your PR!

PR #1016, opened 2020-05-20
Added page id and updated logic for setting parentId of components directly under page

vladbailescu:


content/src/content/jcr_root/apps/core/wcm/components/commons/datalayer/clientlibs/js/datalayer.js
```diff
@@ -38,13 +38,7 @@
...
 
             if (parentElement.tagName === "BODY") {
-                parentData = dataLayer.find(function(element) {
-                    return element.data !== undefined && element.data.page !== undefined;
-                });
-
-                if (parentData !== undefined) {
-                    component[componentID].parentId = Object.keys(parentData.data.page)[0];
-                }
+                component[componentID].parentId = Object.keys(dataLayer.getState("page"))[0];
```
jckautzmann, line 41: I would get the page ID directly:
`component[componentID].parentId = parentElement.id`

vladbailescu: Then it probably makes no sense to tread `body` elements differently, we could use `component[componentID].parentId = parentElement.id` for both cases, right?

jckautzmann: yes, makes sense.

vladbailescu: Done in 8695193

PR #1029, opened 2020-05-31
Fix possible DataLayer NPEs

ky940819:
| Fixed Issues?            | None filed
| Tests Added + Pass?      | None added
| Documentation Provided   | N/A
| Any Dependency Changes?  | N/A

Fixes possible NPEs introduced in the DataLayer.

Specifically:
1) in AccordionImpl `item.getData().getId()` introduces a possible NPE because getData() may return null.
2) in PanelContainerImpl `.findFirst().get();` introduces a possible NPE because findFirst() returns an Optional, Optional.get() causes a NPE if the Optional is empty. It does not return null if the Optional is empty. The null check following the `findFirst().get()` is impossible to be false, because a NPE would occur prior to the null check.
3) in PanelContainerItemImpl the use of a possibly null default value in a call to ValueMap.get() can result in an NPE because not every ValueMap implementation handles a default value of null. For this reason the documentation for ValueMap indicates this value must not be null. 

Additionally, `TabImpl#getDataLayerShownItems()` is cleaned up. There are several unnecessary temporary variables. Streams & Optionals area already in use, might as well follow them through to the end result which introduces less opportunity for error.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AccordionImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AccordionImpl.java
@@ -132,12 +132,10 @@ public class AccordionImpl extends PanelContainerImpl implements Accordion {
             List<String> expandedItemsName = Arrays.asList(expandedItems);
             List<ListItem> items = this.getItems();
 
-            List<String> expandedItemIdsList = items.stream()
+            expandedItemIds = items.stream()
                 .filter(item -> expandedItemsName.contains(item.getName()))
                 .map(item -> item.getData().getId())
-                .collect(Collectors.toList());
-
-            expandedItemIds = expandedItemIdsList.toArray(new String[expandedItemIdsList.size()]);
+                .toArray(String[]::new);
 
         }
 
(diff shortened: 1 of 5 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AccordionImpl.java
```diff
@@ -130,19 +131,13 @@ public String getHeadingElement() {
 
         if (expandedItemIds == null) {
             List<String> expandedItemsName = Arrays.asList(expandedItems);
-            List<ListItem> items = this.getItems();
-
-            if (items != null) {
-                List<String> expandedItemIdsList = items.stream()
-                    .filter(item -> expandedItemsName.contains(item.getName()))
-                    .map(item -> item.getData().getId())
```
ky940819, line 138: Possible NPE due to getData() returning null

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PanelContainerImpl.java
```diff
@@ -33,26 +34,24 @@
...
     }
 
     @Override
     protected Map<String, ComponentExporter> getItemModels(@NotNull SlingHttpServletRequest request,
                                                            @NotNull Class<ComponentExporter> modelClass) {
         Map<String, ComponentExporter> models = super.getItemModels(request, modelClass);
-        models.entrySet().forEach(entry -> {
-            ListItem match = getItems().stream()
-                .filter(item -> item != null && StringUtils.isNotEmpty(item.getName()) && StringUtils.equals(item.getName(), entry.getKey()))
-                .findFirst().get();
```
ky940819, line 51: Possible NPE due to calling Optional.get() without a isPresent() check.

jckautzmann:
@ky940819 Thx for your PR!

PR #1037, opened 2020-06-05
[AMP] AMP config is overwritten by the core config

jckautzmann:
- create the AMP config below /apps/core/wcm/extensions/amp/config


jckautzmann:
The approach used here didn't work because of a limitation in Jackrabbit Vault.

Issue #1039, opened 2020-06-06
[RTC] Table Component


rajeevyadav2:
## Use Case
As an author, I want to insert a table component on a page. 

## Requirements
- Component should be able to populate data from selected source (Resource or DAM file)
- Component should be scalable and extensible. Initially it support Table population from Resource children nodes and CSV but should be extendable to other sources like XLS, JSON etc
- Author should be able to configure the Table Headers that needs to be populated from selected resource

## Edit Design Properties 
- source - Path Browser
- headers - Multi Field

tjameswhite:
Addresses https://github.com/adobe/aem-core-wcm-components/issues/894

rajeevyadav2:
@tjameswhite : FYI.. Me and @rajramo61  have started the table feature code at https://github.com/rajeevyadav2/aem-core-wcm-components/tree/feature/table

rajeevyadav2:
@tjameswhite : I have csv file in test resources and i want to exclude it from rat plugin scan. Can you please confirm which pom.xml need to be updated.

rajeevyadav2:
@jckautzmann @msagolj @gabrielwalt 
I want to exclude below file from rat plugin scan.
https://github.com/rajeevyadav2/aem-core-wcm-components/blob/feature/table/bundles/core/src/test/resources/table/test-content.csv

I tried adding this file in root and parent plugin but I am still getting rat error.
 Can you please assist in fixing this issue.

rajeevyadav2:
@jckautzmann @msagolj @gabrielwalt @vladbailescu 
@rajramo61  and I have raised PR for  this enhancement. PR link is below:
https://github.com/adobe/aem-core-wcm-components/pull/1058

Pls review and let us know if any changes needed.

rajeevyadav2:
@msagolj Thanks for your details regarding the PR. Below are my POV:
1. We can keep table name open or else enhance this component for manual authoring of table in future but IMHO it is cumbersome to allow table authoring especially for large number of table rows. In my experience, authors prefer a way to upload the table data through various sources so that it can be used to display in table. I am open to suggestions on this.

2. Regarding use case of children nodes: There are cases where authors want to display imported data in a table eg products, stocks etc where importer data is stored as children nodes under any resource. Most of the time, developers end up creating a separate component to display that data into a table. This component provides flexibility to select any type of data based on properties into a table.

[...]

PR #1058, opened 2020-06-13
Feature/table

rajeevyadav2:
| Fixed Issues?            | #1039 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Yes
| Any Dependency Changes?  |

Feature Enhancement https://github.com/adobe/aem-core-wcm-components/issues/1039

Description:
This is new table component. It allows authors to create table from CSV or AEM Resource children nodes based on headers selected in dialog.  Table Resource Processor services scalable and additional processors can be added to support different mime types.

Main Contributors for this PR : @rajramo61 & @rajeevyadav2


rajeevyadav2:
Adding Table component in Core

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Table.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.IOException;
+import java.util.List;
+
+public interface Table extends Component {
+
+    /**
+     * Returns formatted property names by removing jcr: from the property name.
+     * @return
+     */
+    default List<String> getFormattedHeaderNames() {
```
ky940819, line 27: Should indicate if return value is `@Nullable` or `@NotNull`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Table.java
```diff
@@ -0,0 +1,48 @@
...
+package com.adobe.cq.wcm.core.components.models;
+
+import java.io.IOException;
+import java.util.List;
+
+public interface Table extends Component {
+
+    /**
+     * Returns formatted property names by removing jcr: from the property name.
+     * @return
```
ky940819, line 25: Should adhere to Javadoc requirements.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Table.java
```diff
@@ -0,0 +1,48 @@
...
+ ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ ~ See the License for the specific language governing permissions and
+ ~ limitations under the License.
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+package com.adobe.cq.wcm.core.components.models;
+
+import java.io.IOException;
+import java.util.List;
+
+public interface Table extends Component {
```
ky940819, line 21: Interface should be documented

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/table/ResourceProcessor.java
```diff
@@ -0,0 +1,27 @@
...
+ ~ limitations under the License.
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+package com.adobe.cq.wcm.core.components.services.table;
+
+import org.apache.sling.api.resource.Resource;
+
+import java.io.IOException;
+import java.util.List;
+
+public interface ResourceProcessor {
```
ky940819, line 23: Needs Javadocs

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableImpl.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @Override
+    public List<String> getFormattedHeaderNames() {
+        return formattedTableHeaderNames;
+    }
+
+    @Override
+    public List<List<String>> getItems() throws IOException {
+        if (nonNull(resourceProcessors)) {
```
ky940819, line 110: field `this.items` is updated on every call but not used.
Should either eliminate the object field `items` and just return directly; or, more likely, this is intended to only process the data once, in which case it should only be run if `this.items == null`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableImpl.java
```diff
@@ -0,0 +1,127 @@
...
+    adapters = {Table.class, ComponentExporter.class},
+    resourceType = TableImpl.RESOURCE_TYPE)
+
+@Exporter(name = ExporterConstants.SLING_MODEL_EXPORTER_NAME,
+    extensions = ExporterConstants.SLING_MODEL_EXTENSION)
+public class TableImpl extends AbstractComponentImpl implements Table {
+
+    public static final String RESOURCE_TYPE = "core/wcm/components/table/v1/table";
+    @ValueMapValue(name = "source", injectionStrategy = InjectionStrategy.OPTIONAL)
+    private String source;
```
ky940819, line 53: Should annotate optional fields as `@Nullable` and resolve any possible NPEs that might result from the field being empty.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableImpl.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @PostConstruct
+    public void initModel() {
+        formatHeaderNames();
+    }
+
+    private String getSourceResourceMimeType(Resource sourceResource) {
+        if (nonNull(sourceResource)) {
+            if (DamUtil.isAsset(sourceResource)) {
+                Asset asset = DamUtil.resolveToAsset(sourceResource);
```
ky940819, line 79: asset can, in theory, be null because `DamUtil.resolveToAsset` uses `adaptTo` which doesn't guarantee a non-null return.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/table/DefaultResourceProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public List<List<String>> processData(Resource resource, String[] headerNames) {
+        List<List<String>> rows = new ArrayList<>();
+        if(!resource.hasChildren()) {
+            return rows;
+        }
+
+        for (Resource child : resource.getChildren()) {
+            ValueMap props = child.adaptTo(ValueMap.class);
```
ky940819, line 43: just using `child.getValueMap()` would guarantee `props` is non-null.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableImpl.java
```diff
@@ -0,0 +1,127 @@
...
+    private List<String> formattedTableHeaderNames;
+    private List<List<String>> items;
+
+
+    @PostConstruct
+    public void initModel() {
+        formatHeaderNames();
+    }
+
+    private String getSourceResourceMimeType(Resource sourceResource) {
```
ky940819, line 76: Since the default return is an empty string, and the default processor accepts any mime type of empty string, this will cause the default processor to accept any resource or asset for which we were unable to derive a mime type.

This includes if the resource is null, which causes a NPE in the default processor.
The processor interface should probably be marked as expecting a `@NotNull` resource, and double check that this is the behaviour you want.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/services/table/DefaultResourceProcessorTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+class DefaultResourceProcessorTest {
+
+    private static final String[] HEADER_NAMES = {"email", "name", "gender", "status"};
+    private static final String[] PROP_VALUES = {"test@test.com", "test", "male", "active"};
+    @InjectMocks
+    private DefaultResourceProcessor defaultResourceProcessor;
+
+    @Mock
+    Resource resource;
```
ky940819, line 46: I would suggest not mocking the resource, it becomes very implementation-specific.
Why not just load a resource from JSON or use the resource builder as part of the aem testing library?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableImpl.java
rajramo61: @ky940819 DamUtil.isAsset(sourceResource) check is to ensure that the sourceResource is an asset. If it is Asset then it should adaptTo Asset with no issues. It can throw null if it is not adaptable to Asset which is not the case here.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/services/table/DefaultResourceProcessorTest.java
rajeevyadav2: @ky940819 : I was facing issues with json loader to test resource. Would you mind to update this test in the PR. :)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableImpl.java
ky940819: `resolveToAsset` also performs an `isAsset` check; by performing your own check you are only preventing the parent resolution if the current resource is a child of an asset. 

Either way, `isAsset` only confirms the resource type is "dam:Asset". It makes no promise that `Resource.adaptTo(Asset.class)` will succeed. 
It is never safe to assume that `Adaptable.adaptTo()` will return a non-null value.

You and I both know that a resource of type "dam:Asset" will --under normal conditions, in most cases -- return an Asset. 
However, the Sling API explicitly states that it is acceptable for `adaptTo()` to return null, so we always have to assume that will happen at some point and check for null.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableImpl.java
rajramo61: Marked it as Required.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableImpl.java
rajramo61: Added null check in getItems to avoid NPE.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableImpl.java
rajramo61: Thanks for the quick responses, I really appreciate it. Certainly adding Null check is the safest way,  I will add it here.
Do you think any scenario, where asset type is `dam:Asset` and fails to adapt to Asset?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableImpl.java
rajramo61: Removed the items object. Returning empty list for negative scenarios.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/table/DefaultResourceProcessor.java
rajramo61: Changed the `adaptTo(ValueMap.class)` to `getValueMap()`. Without streams also the code was fine and was easy to read. However, I changed it as per the request.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Table.java
rajramo61: done.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Table.java
rajramo61: done

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Table.java
rajramo61: done

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/table/ResourceProcessor.java
rajramo61: done

rajeevyadav2:
@bpauli @richardhand @vladbailescu : could you please review this PR and let us know if any findings

content/src/content/jcr_root/apps/core/wcm/components/table/v1/table/.content.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jcr:root xmlns:sling="http://sling.apache.org/jcr/sling/1.0" xmlns:cq="http://www.day.com/jcr/cq/1.0"
+          xmlns:jcr="http://www.jcp.org/jcr/1.0"
+          cq:icon="table"
+          jcr:primaryType="cq:Component"
+          jcr:title="Table (v1)"
+          componentGroup=".wcm-core"/>
```
msagolj, line 7: should be `.core-wcm` not` .wcm-core`

content/src/content/jcr_root/apps/core/wcm/components/table/v1/table/_cq_dialog/.content.xml
```diff
@@ -0,0 +1,73 @@
...
+                    <column
+                        jcr:primaryType="nt:unstructured"
+                        sling:resourceType="granite/ui/components/foundation/container">
+                        <items jcr:primaryType="nt:unstructured">
+                            <id
+                                jcr:primaryType="nt:unstructured"
+                                sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+                                fieldDescription="HTML ID attribute to apply to the component."
+                                fieldLabel="ID"
+                                name="./id"/>
```
msagolj, line 34: the `id` field usually goes to the bottom of the dialog

rajeevyadav2: Done

content/src/content/jcr_root/apps/core/wcm/components/table/v1/table/.content.xml
rajeevyadav2: done

content/src/content/jcr_root/apps/core/wcm/components/table/v1/table/_cq_dialog/.content.xml
msagolj: @rajeevyadav2 i have more general comments from usability/author perspective so iÎ added my comment to the issue itself, see
https://github.com/adobe/aem-core-wcm-components/issues/1039#issuecomment-646173257

vladbailescu:
Code is being added to the contrib repository in https://github.com/adobe/aem-contrib-wcm-components/pull/8

PR #1059, opened 2020-06-15
[List] Use local-sensitive collator when sorting by title #1055

### LINKED ISSUE #1055: [List] List component should use Collator for sorting by title
## Feature Request

When sorting a list by title, the list is sorted using a straight sting comparison.
While this works fine for most text, when you have accented forms of the same base letter, they do not sort properly.
Example: "e" vs "é".

The sorting of text, in general, a locale-sensitive exercise.
This is what the [Collator](https://docs.oracle.com/javase/8/docs/api/java/text/Collator.html) comparator is designed for.

The proposed solution is to 

1. determine the locale for the current page
2. construct a Collator for that locale inside of the list comparator
3. use that Collator for title comparisons

Here is an example of why this matters: https://www.onlinegdb.com/HJBaodeT8
If that doesn't work, this is the output of the example:

- String Compare:                 A, a, b, e, À, é
- English (United Kingdom):       A, À, a, b, e, é 

**Are there alternatives?**
No

**Documentation**
N/A

----
Sidenote: I have a PR for this that I am holding for when other changes to the ListImpl are merged so as to avoid conflicts.

ky940819:
| Fixed Issues?            | Fixes #1055 

Updates the ListImpl to use a locale-sensitive collator when
sorting list items by title.

This corrects an issue where list items with accents are sorted
out of order.

### DIFF
--- a/bundles/core/src/test/resources/list/test-content.json
+++ b/bundles/core/src/test/resources/list/test-content.json
@@ -77,6 +77,21 @@
                 "/content/list/pages/page_2"
             ]
         },
+        "staticOrderByTitleListTypeWithAccent": {
+            "jcr:primaryType"     : "nt:unstructured",
+            "sling:resourceType"  : "core/wcm/components/list",
+            "listFrom"            : "static",
+            "showThumbnail"       : "true",
+            "linkItems"           : "true",
+            "showDescription"     : "true",
+            "showModificationDate": "true",
+            "orderBy"             : "title",
+            "pages"               : [
+                "/content/list/pages/page_1",
+                "/content/list/pages/page_5",
+                "/content/list/pages/page_2"
+            ]
+        },
         "staticOrderByModificationDateListType"    : {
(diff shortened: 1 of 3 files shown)


bundles/core/src/test/resources/list/test-content.json
```diff
@@ -77,6 +77,21 @@
...
+            "sling:resourceType"  : "core/wcm/components/list",
+            "listFrom"            : "static",
+            "showThumbnail"       : "true",
+            "linkItems"           : "true",
+            "showDescription"     : "true",
+            "showModificationDate": "true",
+            "orderBy"             : "title",
+            "pages"               : [
+                "/content/list/pages/page_1",
+                "/content/list/pages/page_5",
```
vladbailescu, line 91: Just nitpicking but I would put `page_5` last here, makes it more clear that the order defined here is not the expected one.

ky940819: Done

PR #1080, opened 2020-06-28
[Navigation] Fix inconsistent `isActive` behaviour caused by redirect chains + cleanup

### LINKED ISSUE #1079: [Navigation] Navigation item active state inconsistent depending on length of redirect chain
## Bug Report

**Current Behavior**
A navigation item is marked as active if the item is either:

- The current page; or,
- A parent of the current page; or,
- Redirects to the current page (one level redirect only)

The actual navigation item will follow a chain of redirects when determining the effective redirect target.
Consider the following navigation structure:

- navRoot
  - page1 (current page)
  - page2 (redirect to page1)
  - page3 (redirect to page2)

`page1` item will be active because it is the current page.
`page2` item will be active because it redirects to the current page.
`page3` item will not be active.

In the above example, assuming shadowing is enabled, all three navigation items will have the identical target path and title. 

**Expected behavior/code**
The expected behaviour would be that the active state is set based on the effective navigation item page, following any number of redirects, such that if two items in the navigation are effectively identical - but reached via different redirect chains - they should not appear different in the output.

**Environment**
- AEM 6.5.5.0
- CC 2.10.0
- JDK 11

**Possible Solution**
The `PageListItemImpl` already has the logic for following a redirect chain.
This is the logic that is actually used when determining the navigation items.
This logic should simply be reused in `NavigationImpl` to ensure that they are consistent.

I will submit a PR

In the future, further enhancements could be made to ensure that all navigation items behave in a consistent manner, and avoid the requirement to compute the redirect target twice (once to determine if the item is active, then again in the `PageListItemImpl` constructor to get the effective page).

If this is not addressed, then perhaps it would be worth considering when addressing #688.

ky940819:
| Fixed Issues?            | Fixes #1079 

* Modifies `NavigationImpl` to reuse redirect target logic from `PageListItemImpl` to ensure consistent `isActive` behaviour, regardless of length of redirect chain.
* Adds javadoc comments to private members of `NavigationImpl` to aid future development (mostly for me :) )
* Reworks `NavigationImpl` to avoid the need for the inner class `NavigationImpl.NavigationRoot`
* Splits the logic for getting the effective navigation root out of the `NavigationImpl#getItems()` method to make the behaviour for generating the navigation items/tree easier to understand. 
* Removes various injected fields that can be obtained from other injected fields
* Update `PageListItemImpl` constructor to eliminate determining the redirect target if shadowing is disabled (minor performance optimization)

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/NavigationImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/NavigationImpl.java
@@ -24,7 +24,6 @@ import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import javax.annotation.PostConstruct;
-import javax.jcr.RangeIterator;
 
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.resource.Resource;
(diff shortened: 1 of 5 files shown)


vladbailescu:
@ky940819 , thanks for the contribution. Do you think it would be helpful to include a test case for this as well?

ky940819:
> @ky940819 , thanks for the contribution. Do you think it would be helpful to include a test case for this as well?

Hello, yes it's probably a good idea to explicitly test this, I have added the tests.
see https://github.com/adobe/aem-core-wcm-components/pull/1080/commits/1c7355c2f4d1699a7bae130ac565958abe80fd5b

PR #1091, opened 2020-07-03
[Image] Allow lazy threshold to be set in design policy #1090

### LINKED ISSUE #1090: [Image] Lazy threshold should be configurable in policy
## Feature Request

Lazy loading images only load when the image is visible.
There is a property `LAZY_THRESHOLD = 0` in the v2 image.js that can be used to configure this such that values >0 allow the image to begin preloading before the image is in view.

Sometimes it is useful to allow images to begin to load slightly before coming into view so as to anticipate them coming into view without a jarring delay.

Allow the value of this threshold to be configured in the policy for the image component.

**Are there alternatives?**
Copy the JS and override the value of `LAZY_THRESHOLD`.

ky940819:
| Fixed Issues?            | Fixes #1090 

Allows template author to set the lazy threshold value so that lazy-loading images can be loaded slightly in advance of being visible on the page in order to reduce delay and jarring image loads and give a more seamless/smooth experience.

Note: someone may want to review the wording of the dialog field name/description to make sure that it makes sense.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageAreaImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageAreaImpl.java
@@ -17,16 +17,57 @@ package com.adobe.cq.wcm.core.components.internal.models.v1;
 
 import com.adobe.cq.wcm.core.components.models.ImageArea;
 
-public class ImageAreaImpl implements ImageArea {
+/**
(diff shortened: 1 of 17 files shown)


vladbailescu:
Thanks @ky940819 for this contribution. The code looks good to me, I think you should also update the readme with the name and description of the new property.

bpauli:
I will add the changes to the readme directly on the development branch.

@ky940819 Thank you!

PR #1103, opened 2020-07-15
https://github.com/adobe/aem-core-wcm-components/issues/1102 : [Teaser]

### LINKED ISSUE #1102: [Teaser] Title is not clickable or hyperlink when external link (outside AEM) is configured in the call to actions link
## Feature Request

Title of the teaser component is clickable or has hyperlink when the first call to actions link is configured with internal link/page within AEM.
If the first call to actions link configured is external link like https://www.adobe.com/ then the title is not clickable or is not hyperlink
This is inconsistent experience on the page where we have multiple teaser component configured with both internal or external link configured in the first call to actions link.

Update the TeaserImpl.java file to support external link configured in the first call to actions link.

**Are there alternatives?**
As work-around, it has been handled through CSS  but this would be custom implementation that needs to be implemented whoever needs this feature.

**Documentation**
Maybe a screenshot or design?
There is no change in content authoring or dialog. Only change is in TeaserImpl.java which would take care to support external link in call to actions link.

satishkumarjadhav:
Title is not clickable or hyperlink when external link (outside AEM) is
configured in the call to actions link

| Fixed Issues?            | Fixes #1102 

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java has been updated to support the change.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
@@ -125,7 +125,14 @@ public class TeaserImpl extends AbstractImageDelegatingModel implements Teaser {
             if (!actions.isEmpty()) {
                 ListItem firstAction = actions.get(0);
                 if (firstAction != null) {
-                    targetPage = pageManager.getPage(firstAction.getPath());
+                	LOGGER.debug("firstaction URL : "+firstAction.getURL());
+                	if(firstAction.getPath().startsWith("/content")) {
+                		LOGGER.debug("internal link");
+                		targetPage = pageManager.getPage(firstAction.getPath());
+                	} else if(firstAction.getPath().startsWith("https")) {
+                		LOGGER.debug("external link");
+                		linkURL=firstAction.getURL();
+                   	}
                 }
             }
         } else {
@@ -185,8 +192,9 @@ public class TeaserImpl extends AbstractImageDelegatingModel implements Teaser {
         if (hasImage) {
             setImageResource(component, request.getResource(), hiddenImageResourceProperties);
         }
-        if (targetPage != null) {
+        if ((targetPage != null) && (targetPage.getPath().startsWith("/content"))) {
             linkURL = Utils.getURL(request, targetPage);
+            LOGGER.debug("internal link url : "+linkURL);
         }
     }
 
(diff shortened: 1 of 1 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
```diff
@@ -125,7 +125,14 @@ private void initModel() {
             if (!actions.isEmpty()) {
                 ListItem firstAction = actions.get(0);
                 if (firstAction != null) {
-                    targetPage = pageManager.getPage(firstAction.getPath());
+                	LOGGER.debug("firstaction URL : "+firstAction.getURL());
```
ky940819, line 128: I wouldn't assume that an action is only an internal page link if the path starts with `/content`.
I also wouldn't assume that it's external if it starts with `https`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
```diff
@@ -185,8 +192,9 @@ private void initModel() {
         if (hasImage) {
             setImageResource(component, request.getResource(), hiddenImageResourceProperties);
         }
-        if (targetPage != null) {
+        if ((targetPage != null) && (targetPage.getPath().startsWith("/content"))) {
```
ky940819, line 195: Would it make more sense to just pull this up to L130?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
```diff
@@ -333,7 +341,8 @@ private Action(Resource actionRes, String parentId) {
             ctaTitle = ctaProperties.get(PN_ACTION_TEXT, String.class);
             ctaUrl = ctaProperties.get(PN_ACTION_LINK, String.class);
             ctaPath = actionRes.getPath();
-            if (ctaUrl != null && ctaUrl.startsWith("/")) {
+            if (ctaUrl != null) {
```
ky940819, line 344: Why remove the path restriction? I don't know of a situation in which the page manager will return a non-null value for a path that starts with anything other than `/`.

jckautzmann:
@satishkumarjadhav Thx for your contribution. We decided to go with #1106 as it is a more generic fix for the issue.

PR #1106, opened 2020-07-17
[Teaser] Fix inconsistent `linkURL` behaviour + cleanup

### LINKED ISSUE #1102: [Teaser] Title is not clickable or hyperlink when external link (outside AEM) is configured in the call to actions link

ky940819:
| Fixed Issues?            | Fixes #1102 

- Modifies the `TeaserImpl` class such that, when the title link is derived from the first CTA, it does not matter if the CTA link URL points to an internal or external page (alternative solution to PR #1103).
- Adds javadoc comments to private members of TeaserImpl.
- Significant rewrite of field initialization/getters so that the initialization logic for each field is contained within its own getter instead of all in the `initModel()`. The intent of this change is to make it easier to see how a field is set at a glance instead of having to trace through the entire `initModel()` method where fields may be set and subsequently unset and reset.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
@@ -324,22 +324,21 @@ public class TeaserImpl extends AbstractImageDelegatingModel implements Teaser {
 
         private static final String CTA_ID_PREFIX = "cta";
 
+        @NotNull
         private final Resource ctaResource;
         private final String ctaTitle;
         private final String ctaUrl;
-        private final String ctaPath;
         private final Page ctaPage;
         private final String ctaParentId;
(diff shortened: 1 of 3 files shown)


jckautzmann:
Thx @ky940819 for your contribution! Very much appreciated!

PR #1113, opened 2020-07-17
[LayoutContainer] Fix possible NPE in `LayoutContainerImpl` + cleanup

### LINKED ISSUE #1112: [LayoutContainer] Possible NPE in LayoutContainer model initializer
## Bug Report

**Current Behavior**
There exists a possible NPE in the `LayoutContainer` model initializer.
The NPE is encounters only when the following conditions are met:

1. The layout type is not defined in the component properties; and,
2. The layout type is not defined in the style properties; and,
3. The value map provided by the current resource does not support using null as a default value

**Expected behavior/code**
No NPE should occur. The default layout type of `SIMPLE` should be used.

**Environment**
- AEM version: 6.5.5.0
- Core Components version: 2.10.0
- JRE version: JDK 11

**Possible Solution**
Will submit a PR.

ky940819:
| Fixed Issues?            | Fixes #1112 
| Tests Added + Pass?      | None added, all pass

- Fixes a possible NPE in the `LayoutContainerImpl.initModel()` method.
- Clean up `ResourceListItemImpl` by removing the unused `url` field and unused `request` constructor parameter.
- For classes that were touched in this PR, javadocs are added for any undocumented members.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/LayoutContainerImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/LayoutContainerImpl.java
@@ -29,16 +29,31 @@ import org.jetbrains.annotations.NotNull;
 
 import com.adobe.cq.wcm.core.components.models.LayoutContainer;
 
+/**
+ * Layout container model implementation.
+ */
 @Model(adaptables = SlingHttpServletRequest.class, adapters = LayoutContainer.class, resourceType = LayoutContainerImpl.RESOURCE_TYPE_V1)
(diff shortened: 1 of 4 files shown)


vladbailescu:
Thanks for the contribution. It would be great if you could include a test case for the fixed issue.

ky940819:
> Thanks for the contribution. It would be great if you could include a test case for the fixed issue.

It's a challenging issue to test; the issue does not cause an NPE in most standard use cases.
To effectively test this would require contriving the situation by creating a new ValueMap implementation that doesn't support null defaults. 

The requirement for the default value to not be null is documented:
https://sling.apache.org/apidocs/sling10/org/apache/sling/api/resource/ValueMap.html#get-java.lang.String-T-

PR #1115, opened 2020-07-20
[DataLayer] Add `DataLayerBuilder` public utility

### LINKED ISSUE #1114: [DataLayer] Data Layer Builder
## Feature Request

It is difficult for custom components (non core-components) to implement the data layer.
The data layer objects, and most of the methods that do some heavy lifting (example: generating an ID) are not exported and thus cannot be accessed outside of the project.

There are two issues:

1. It is challenging to implement the data layer for components that are not part of core components
2. it is challenging to extend/modify the data layer for custom component models that are overlays/wrappers for a core components model

The current implementation requires that a client who wishes to provide data layer for their own components must either re-implement the data layer interfaces and all associated logic; or, if their component has a super type that is a core component, then in their model they may adapt the request to the super type model and get the data layer, and then wrap that data layer to add/remove/modify anything they want while keeping in mind that the data layer object returned from the super model does not take into account any of the custom logic they may have overridden in their implementation. 

I propose that a `DataLayerBuilder` be made available to provide an easy API for people to deliver data layer integration from their custom components. For ease of use, the builder should use the Fluent Interface design such that IDE autocompletion can largely guide it's use.

An example of what this would look like:
``` JAVA
ComponentData data = DataLayerBuilder.forComponent()
    .withId(() -> /* generate ID */)
    .withTitle(this::getTitle)
    .withDescription(this::getDescription)
    .withText(() -> "some text")
    .withLastModifiedDate(this::getLastModifiedDate)
    .withLinkUrl(this::getLinkUrl)
    .withType(resource::getResourceType)
    .build();
```

**Are there alternatives?**
Require that clients fully implement custom classes that match the `*Data` interfaces.

**Documentation**
Will provide JavaDocs in PR; however, I would appreciate input on technical documentation. 

For further information please see discussion on the google group:
https://groups.google.com/g/aem-core-components-dev/c/kvhg-OQITIY

ky940819:
| Fixed Issues?            | Fixes #1114 
| Documentation Provided   | Yes (code comments / javadoc)

- Adds `DataLayerBuilder` utility to allow custom (non-core-components) models to easily provide `DataLayer` compliant data.
- Makes some previously internal `DataLayer` related utility methods public (specifically, those used for generating the ID)
- Updates existing internal models to use the builder


ky940819:
Is someone able to tell me why the builds failed?
I don't use CircleCI and am unable to grant it access to my private repositories just to see these results.

vladbailescu:
@ky940819 , the error is from the `baseline` check:

`[ERROR] com.adobe.cq.wcm.core.components.models.datalayer: Version increase required; detected 1.0.0, suggested 1.1.0`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
```diff
@@ -381,25 +382,18 @@ private boolean smartSizesSupported() {
...
-
-    @Override
-    public String getDataLayerLinkUrl() {
-        return getLink();
+    @NotNull
+    protected ImageData getComponentData() {
+        return DataLayerBuilder.extending(super.getComponentData()).asImageComponent()
+            .withTitle(this::getTitle)
+            .withLinkUrl(this::getLink)
+            .withAssetData(() ->
```
vladbailescu, line 390: What if the `Supplier<AssetData>` would be just an anonymous function that returns a locally defined `AssetData` instance?

vladbailescu: Or just use an `AssetData` as parameter of `withAssetData`?

ky940819: My thought here was that if you already have an `Asset` then you can use this to get the AssetData for that asset.
```JAVA
AssetData data = DataLayerBuilder.forAsset(Asset).build()
```
This is just a typical implementation, you override any of the suppliers before doing the `build()` if there is some way that your model should differ from what this helper provides.

If you do not have an Asset then you can use this
```JAVA
AssetData data = DataLayerBuilder.forAsset()
  .withId(/**/)
  ...
  .build()
```

Regardless of how you get the AssetData, including it in the component model will be like this:
```JAVA
ImageData imageData = DataLayerBuilder.forImage()
  .withAssetData(() -> assetData)
  .build();
```

This pattern encourages you to use the builder for creating AssetData, but if you really wanted to you could create some class that implements `AssetData` and then supply an instance of that to the `.withAssetData(() -> customAssetDataInstance)`. That said, i don't think there is any reason to do this since the builder can create the data you want.

vladbailescu: That's exactly what I was suggesting: creating `AssetData` once and mapping it directly in `withAssetData` instead of mapping a function that recreates it on every call of `getAssetData`.

ky940819: Hello,

I agree with you that I don't like the amount of re-work that is done if the `ComponentDataImpl#getAssetData()` method is called twice.

### Proposal 1
The computation of the `AssetDataImpl` object could be moved outside of the supplier, and cached similar to how other getters are done in models - such that the block is only executed once.

```JAVA
    private AssetData getAssetData() {
        if (this.assetData == null) {
            this.assetData = Optional.ofNullable(this.fileReference)
                .map(reference -> this.request.getResourceResolver().getResource(reference))
                .map(assetResource -> assetResource.adaptTo(Asset.class))
                .map(DataLayerBuilder::forAsset)
                .map(AssetDataBuilder::build)
                .orElse(null);
        }
        return this.assetData;
    }

    @Override
    @NotNull
    protected ImageData getComponentData() {
        return DataLayerBuilder.extending(super.getComponentData()).asImageComponent()
            .withTitle(this::getTitle)
            .withLinkUrl(this::getLink)
            .withAssetData(this::getAssetData)
            .build();
    }
```

However, this isn't fully optimal because, while it does eliminate (potentially) duplicate

- resolution of the asset resource
- adaptation of that resource to an Asset object
- creation/building of the data layer builder (not very expensive since no heavy lifting is done)

none of the data builder methods are caching. 
In the above solution, repeated calls to `ComponentDataImpl#getAssetData()` would always return the same object; however, subsequent repeated calls to the returned object getters (e.g. `AssetData#get*()`) would still re-incur the work of calculating the values. It may be worth nothing that the current [`AssetDataImpl`](https://github.com/adobe/aem-core-wcm-components/blob/2f24bf5/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/AssetDataImpl.java) does not cache the values of the getters and recomputes them on-demand, so the above change would result in the same behaviour as currently exists. 

As it stands now (in this PR), the builder does not produce an object which actually contains the computed values, it is more of just setting up the wiring to know how to get the values when needed. In this way the creation of the component/asset data is fairly trivial, the hard work is only done on-demand.

---
### Proposal 2
A solution that would eliminate this issue across the board, in its entirety, would be to cache the values produced by the suppliers inside the `AssetDataImpl` and `ComponentDataImpl`. Then the documentation could make a simple statement that the values are not computed until they are requested, but once requested are never re-computed.

This would change the `ComponentDataImpl#getAssetData()` method from:
```JAVA
    @Override
    @Nullable
    public AssetData getAssetData() {
        return this.getDataLayerSupplier()
            .getAssetData()
            .map(Supplier::get)
            .orElse(null);
    }
```


TO
```JAVA
    @Override
    @Nullable
    public AssetData getAssetData() {
        if (this.assetData == null) {
            this.assetData = this.getDataLayerSupplier()
                .getAssetData()
                .map(Supplier::get)
                .orElse(null);
        }
        return this.assetData;
    }
```

Similar updates would then be done on all the other `Asset/ComponentDataImpl#get*()` methods, which would entirely eliminate the need to worry about the caching issue. 

This isn't as technically complex as the original solution of allowing the caching strategy to be set, instead it's an always cache strategy. 

I favor this solution because it solves the issue but allows for a consistent expectation across the entire data builder, regardless of what is being built. I also think it most likely matches the expectations of the user (developer).

---
### Proposal 3

There are other options, and I think that this might be more in line with what you are thinking.
The other option is the restore the original [`AssetDataImpl`](https://github.com/adobe/aem-core-wcm-components/blob/2f24bf5/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/AssetDataImpl.java) that can be constructed from an `Asset`; and then there would be two `AssetData` implementations (one backed by the builder suppliers, the other is just an object)

In this solution, the `DataLayerBuilder#forAsset(Asset)` method would be changed thusly:
From:
```JAVA
    public static AssetDataBuilder forAsset(@NotNull final Asset asset) {
         return DataLayerBuilder.forAsset()
            .withId(asset::getID)
            ...;
    }
```
To:
```JAVA
    public static AssetData forAsset(@NotNull final Asset asset) {
        new AssetDataImpl(asset);
    }
```

Note that the `DataBuilder#forAsset(Asset)` no longer returns a builder, it returns the asset data object directly.
Any extension or override of that would have to be done by
```JAVA
    DataLayerBuilder.extending(DataLayerBuilder.forAsset(asset))
    ...
    .build()
```

but anything provided after extending the asset data would, again, be lazy and re-computed on every call. 
Also note that the original [`AssetDataImpl`](https://github.com/adobe/aem-core-wcm-components/blob/2f24bf5/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/AssetDataImpl.java) recomputes it's values on every call. As far as I can tell, this solution doesn provide any benefits over **Solution 1** unless we also add caching to [`AssetDataImpl`](https://github.com/adobe/aem-core-wcm-components/blob/2f24bf5/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/AssetDataImpl.java), which would have the same effect as adding it to the supplier backed `AssetDataImpl` provided in this branch (effectively **Solution 2** (_bold just to match title, not for emphasis_)).

---

p.s.: Thank you for taking the time to review this in depth and find these issues; I appreciate the feedback.

ky940819:
Hello,
@vladbailescu I'd appreciate some feedback on the open conversation whenever you are able to.
This branch has been sitting open for awhile now and is getting stale.

There are some other tickets I'd like to tackle as well, but I know they can't be completed until this is merged (without causing significant conflicts).

kwin:
I would move the `datalayer` package from `com.adobe.cq.wcm.core.components.models.datalayer` to `com.adobe.cq.wcm.core.components.datalayer`. Also the implementation should be moved from `com.adobe.cq.wcm.core.components.internal.models.v1.datalayer` to `com.adobe.cq.wcm.core.components.internal.datalayer.v1` and . IMHO the `models` package should contain only Sling Models

ky940819:
> I would move the `datalayer` package from `com.adobe.cq.wcm.core.components.models.datalayer` to `com.adobe.cq.wcm.core.components.datalayer`. Also the implementation should be moved from `com.adobe.cq.wcm.core.components.internal.models.v1.datalayer` to `com.adobe.cq.wcm.core.components.internal.datalayer.v1` and . IMHO the `models` package should contain only Sling Models

Unfortunately this cannot be renamed because the `c.a.c.q.c.c.models.datalayer` package already exists, and there changes in this PR maintain backwards compatibility of those interfaces.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
```diff
@@ -15,41 +15,38 @@
...
-    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractComponentImpl.class);
-
+    /**
+     * The current resource.
+     */
     @SlingObject
     protected Resource resource;
 
+    /**
+     * The component context.
```
jckautzmann, line 48: It should be: The component.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
```diff
@@ -89,76 +92,6 @@ public String getExportedType() {
...
-     * m = 16^n - for a hexadecimal string, where n is the number of characters.
-     *
-     * For i = 1000 components with the same name, and n = 10 characters:
-     *
-     * c ~= (1000^2)/(2*(16^10))
-     * c ~= 0.00000045
-     *
-     * @return the auto generated component ID
-     */
-    private String generateId() {
```
jckautzmann, line 115: I would keep this method here as the ID is specific to a component, not just to compute the data for the data layer

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
```diff
@@ -58,27 +55,33 @@
...
-            if (resource != null) {
-                ValueMap properties = resource.getValueMap();
-                id = properties.get(Component.PN_ID, String.class);
-            }
-            if (StringUtils.isEmpty(id)) {
-                id = generateId();
-            } else {
-                id = StringUtils.replace(StringUtils.normalizeSpace(StringUtils.trim(id)), " ", ID_SEPARATOR);
-            }
+            this.id = DataLayerUtils.getPropertyOrGeneratedId(this.resource, this.currentPage, this.componentContext);
```
jckautzmann, line 84: See my comment below.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ResourceListItemImpl.java
```diff
@@ -95,18 +95,4 @@ public String getPath() {
     public String getName() {
         return name;
     }
-
```
jckautzmann, line 98: Shouldn't this class have a `getComponentData()` method like other components?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
ky940819: I think the functionality for generating the ID was moved so that It did not reside in the internal package such that it could be used by client projects.

If this remains internal to the core components then any custom components would need to safely recreate this logic.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/ComponentDataImpl.java
```diff
@@ -15,101 +15,209 @@
...
+            .map(Supplier::get)
+            .orElse(null);
+    }
 
-        if (lastModified != null) {
-            return lastModified.getTime();
-        }
+    @Override
+    @Nullable
+    public AssetData getAssetData() {
```
jckautzmann, line 147: This methode is specific to an image component. I would re-introduce the removed ImageDataImpl class.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/builder/DataLayerSupplierImpl.java
```diff
@@ -0,0 +1,491 @@
...
+        return supplier;
+    }
+
+    /**
+     * Construct a DataLayerSupplier from existing AssetData.
+     *
+     * @param data The existing asset data.
+     * @return A new DataLayerSupplier that uses the field values from the provided asset data.
+     */
+    public static DataLayerSupplier extend(@NotNull final AssetData data) {
```
jckautzmann, line 186: Should we also introduce an `extend()` method for container, page and image components?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/builder/DataLayerBuilder.java
```diff
@@ -0,0 +1,147 @@
...
+     * Get a ImageComponentDataBuilder for an image component.
+     *
+     * @return An empty ImageComponentDataBuilder.
+     */
+    public static IdRequiredDataBuilder<ImageComponentDataBuilder, ImageData> forImageComponent() {
+        return new IdRequiredDataBuilder<>(new ImageComponentDataBuilder(EMPTY_SUPPLIER));
+    }
+
+    /**
+     * Get a PageDataBuilder for an page.
```
jckautzmann, line 82: typo: for a page

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/LayoutContainerImpl.java
```diff
@@ -75,6 +75,11 @@ protected void initModel() {
             .collect(Collectors.toList());
     }
 
+    @Override
+    public String[] getDataLayerShownItems() {
```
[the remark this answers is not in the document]
ky940819: Is a layout container not a container?

Currently, in the master branch, the method `getDataLayerShownItems` is defined in `AbstractComponentImpl`.
The `LayoutContainerImpl` used to create a `ContainerDataImpl` object in it's `getComponentData` method.
That would functionally be the same as it is now. 

So the only difference is that the `getDataLayerShownItems` was changed from a concrete implementation on `AbstractComponentImpl` to an abstract method on `AbstractContainerImpl`, which necessitated the implementation of the method in `LayoutContainerImpl`

jckautzmann:
hi @ky940819 , thx for your PR. It looks really cool! Sorry it took so long to review. I got a few free cycles to review it. Hopefully you'll have some time to discuss our comments so that we can include it in the next release.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ResourceListItemImpl.java
ky940819: All of the list item implementations share the same `getComponentData()` method (defined in `AbstractListItemImpl`)
The exceptions are the `LanguageNavigationItemImpl` and `PageListItemImpl` which return PageData.

The `getComponentData()` method could be removed from `AbstractListItem` and pushed down to each of the implementations, but with the exception of those two, it would just be duplication.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/builder/DataLayerSupplierImpl.java
ky940819: AssetData is the odd one out, it doesn't extend ComponentData (not being a component).
The previous method `public static DataLayerSupplier extend(@NotNull final ComponentData data)` is used for all other component data.

There is functionality for extending component data using `DataLayerBuilder.extending(componentData.asX()` (where X is the data type `Image`, `Page`, `Container`, etc. 

One issue is that the component models do not tell you what kind of data is going to be returned. I.e. the `Image` component model does not indicate that it will return `ImageData`. Therefore, if a custom model is created which extends the `Image` model via sling delegation pattern, it is impossible for the builder to know that the data from the extended model is `ImageData`. Thus the implementer is required to specify the type, e.g. 
```JAVA
DataLayerBuilder.extending(parentModel.getComponentData()).asImageComponent()
```

This is also the reason why the `instanceof` checks are needed in the `extend(ComponentData)` method. 
I really hate `instanceof` checks here (and pretty much everywhere) - but if the different types of `ComponentData` were separated back into their own implementation classes then someone extending a core component model would need to make a cast, and that cast wouldn't be safe. 

To resolve this issue in a way that doesn't feel like a hack, the component model interfaces should be explicit about the type of data they are going to return; however, that is a breaking change. 

This rational also covers previous comments as to why the different data types are all lumped into one class (`ComponentDataImpl`). Also, while reviewing this, please keep in mind that `ComponentDataImpl` is not exposed - outside of this package you can only interact with the `DataLayerBuilder`, which will only present you with the methods applicable for the specified data type.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/DataLayerUtils.java
```diff
@@ -0,0 +1,148 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Optional;
+
+import static com.adobe.cq.wcm.core.components.internal.Utils.ID_SEPARATOR;
+
+/**
+ * Utility helper functions for Data Layer.
+ */
+public final class DataLayerUtils {
```
jckautzmann, line 38: I see your point in creating a utility class that can be useful when creating custom components (https://github.com/adobe/aem-core-wcm-components/pull/1115#discussion_r499668347).
As the methods that relate to ID generation can be used in context that are not necessarily specific to the data layer (like HTML ID), I would move this class into `com.adobe.cq.wcm.core.components.models` and rename it `Utils`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/LayoutContainerImpl.java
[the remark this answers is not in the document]
jckautzmann: Thx for clarifying. I agree that this change makes sense. `getDataLayerShownItems` should not be part of `AbstractComponentImpl` but of `AbstractContainerImpl`. And yes the layout container is a container :)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ResourceListItemImpl.java
jckautzmann: ok, thx for clarifying. It's fine if this class inherits the `getComponentData()` method from `AbstractListItemImpl`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/builder/DataLayerSupplierImpl.java
jckautzmann: ok, thanks for your explanations, it makes sense to keep it as it is.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
jckautzmann: My preference goes to your proposal#2, which goes more towards what we've tried to achieve in the model implementations.
I would not only cache the return value of `ComponentDataImpl.getAssetData()`method but of all methods of `ComponentDataImpl` and `AssetDataImpl`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/DataLayerUtils.java
vladbailescu: I would rather we put the utility method in something like `com.adobe.cq.wcm.core.components.util.Component`. We could have more utility methods and classes in the future, seems wise to separate them.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
vladbailescu: Sorry for the late answered, I hoped I would get a chance to checkout the changes and maybe come up with another proposal. *Solution 2* is my favourite too.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/DataLayerUtils.java
jckautzmann: how about `com.adobe.cq.wcm.core.components.util.ComponentUtils`, which makes it clear what it is when using it? E.g.: `String id = ComponentUtils.generateId(...)`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ImageImpl.java
ky940819: Thank you, will implement.
For future reference, if the Guava dependency is ever updated to a newer version, there is a MemoizingSupplier class in there that would resolve this issue cleanly, and would also ensure a single execution (i.e. don't try again if it returned null).

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/ComponentDataImpl.java
```diff
@@ -15,101 +15,209 @@
...
+        return this.getDataLayerSupplier()
+            .getAssetData()
+            .map(Supplier::get)
+            .orElse(null);
+    }
 
-        return null;
+    @Override
+    @Nullable
+    public String getTemplatePath() {
```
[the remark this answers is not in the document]
ky940819: Does the explanation [below](https://github.com/adobe/aem-core-wcm-components/pull/1115#discussion_r499713421) clear up this review item, or do you still think changes are required?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/ComponentDataImpl.java
ky940819: Does the explanation [below](https://github.com/adobe/aem-core-wcm-components/pull/1115#discussion_r499713421) clear up this review item, or do you still think changes are required?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
ky940819: `DataLayerUtils` has been moved and renamed to `ComponentUtils`.
Will mark this as resolved.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/DataLayerUtils.java
ky940819: I renamed it to it as `com.adobe.cq.wcm.core.components.util.ComponentUtils` as suggested by @jckautzmann;
I prefer having `Utils` in the name because it is unambiguous that this is a utility class, and naming the class `Component` would cause a lot of class name conflicts, as this utility is likely to be used in conjunction with the `Component` model.  

If you convinced that this is the appropriate solution (as I am), then please feel free to resolve the conversation.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
```diff
@@ -0,0 +1,148 @@
...
+     * Get the ID property value if set (using {@link #getPropertyId(Resource)},
+     * otherwise generate a new ID (using {@link #generateId(Resource, Page, ComponentContext)}.
+     *
+     * @param resource The resource for which to get or generate an ID.
+     * @param currentPage The current request page.
+     * @param componentContext The current component context.
+     * @return The ID property value for the specified resource, or a generated ID if not set.
+     */
+    @NotNull
+    public static String getPropertyOrGeneratedId(@NotNull final Resource resource,
```
vladbailescu, line 70: Does it make sense to expose the other two methods? I would rather rename this to `getId` and keep the other two methods private.

I would move `Utils#generateId(prefix, path)` here and make it public though, in case people want to reuse it.

vladbailescu:
Regarding @kwin 's coment:

> I would move the datalayer package from com.adobe.cq.wcm.core.components.models.datalayer to com.adobe.cq.wcm.core.components.datalayer

I would actually put the builder in `com.adobe.cq.wcm.core.components.datalayer.builder` and keep `com.adobe.cq.wcm.core.components.models.datalayer` just for the serialization models

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/ComponentDataImpl.java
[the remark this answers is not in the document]
jckautzmann: `ComponentDataImpl` sounds like it represents the data of a generic component. I don't like the fact that `ComponentDataImpl` has methods that relate to more specific objects like page and image. What would it take to re-introduce the `PageDataImpl` and `ImageDataImpl` classes? Another option would be to remove `Component` from the class name and rename this class e.g. `DataImpl` or `DataLayerDataImpl`.
@ky940819 @vladbailescu Any other thoughts?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/ComponentDataImpl.java
```diff
@@ -15,101 +15,331 @@
...
+    }
+
+    /**
+     * Get the JSOn for component data.
+     *
+     * @param data The component data.
+     * @return The JSON.
+     */
+    @Nullable
+    static String getJson(@NotNull final ComponentData data) {
```
vladbailescu, line 325: Does it make sense to move this logic to a `static`? Were you planning on moving it to a public utils class?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/ComponentDataImpl.java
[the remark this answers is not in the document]
jckautzmann: I would find it easier to read/understand if each interface has a class that implements the methods of its interface.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/ComponentDataImpl.java
ky940819: This is probably an artifact of refactoring and is no longer needed. 
We could move it to a utility class if there is any purpose to that, otherwise the change can probably just be reverted.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/ComponentDataImpl.java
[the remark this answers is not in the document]
jckautzmann: yes, exactly!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/ComponentDataImpl.java
jckautzmann: we may want to move this method to the new `ComponentUtils` class.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
jckautzmann: It makes sense to move this method to the new `ComponentUtils` class.

jckautzmann:
It would be nice if you could add a few examples to the data layer integration page (https://github.com/adobe/aem-core-wcm-components/blob/master/DATA_LAYER_INTEGRATION.md) showing:
- how to use the new API
- when to use `DataLayerBuilder.forComponent()` as opposed to `DataLayerBuilder.extending(super.getComponentData()).asComponent()`
- how to extend a builder for a custom model getter like `getColor()`

jckautzmann:
I noticed that overriding the parent ID with `DataLayerBuilder.withParentId()` produces the expected JSON with `model.data.json` but does not override the value in the data layer client object as the value of the `parentId` property is populated in the client code and does not consider the `parentId` property of the `data-cmp-data-layer` attribute. We'll need to fix this.

vladbailescu (approved):

jckautzmann:
@ky940819 thx a lot for your great contribution, for the fruitful discussions and for your patience with us.
I will tackle the issue I found and provide a fix in a separate ticket.

PR #1120, opened 2020-07-24
Unclosed resource resolver in Utils.getComponentsResolver

### LINKED ISSUE #1118: Unclosed resource resolver in `Utils.getComponentsResolver`
## Bug Report

**Current Behavior**
The utility method `Utils.getComponentsResolver()` returns a new resource resolver for the `components-service` sub-service.
These resource resolvers are never closed.

**Expected behavior/code**
The resource resolvers should be closed.

**Environment**
- AEM 6.5.5.0
- Core Components 2.11.0
- JDK 11

**Possible Solution**
The resource resolvers should be closed once we are done with them.
This can be tricky to accomplish when the calling method may simply have access to a resource resolved using that resource resolver, which may be null - leaving no method of closing the resource resolver.

Preferably the utility methods that create a new resource resolver should simply be removed and the actual implementations that are requiring service-user based resource resolvers should be responsible for their creation. A simple solution is then to wrap the entire section in a try-with-resources block to ensure that the resource resolvers are always closed. 

Essentially, i would find every method that has a parameter of type `ResourceResolverFactory` and simply replace that with `ResourceResolver` instead. The caller that invokes the method is then responsible for supplying a resource resolver with the required permissions based on the sub-service requirements.

vladbailescu:
| Fixed Issues?            | Fixes #1118
| Tests Added + Pass?      | Existing

Updated implementation to close the resolver

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
@@ -245,19 +245,14 @@ public class Utils {
      * Gets the component resource for a given path
      *
      * @param path - the path
-     * @param request - the request
-     * @param resolverFactory - the resolver factory
+     * @param resolver - the resource resolver
      *
      * @return the corresponding resource
      */
     @Nullable
-    public static Resource getResource(String path, @NotNull SlingHttpServletRequest request, @NotNull ResourceResolverFactory resolverFactory) {
-        if (path == null) {
+    public static Resource getResource(String path, ResourceResolver resolver) {
+        if (path == null || resolver == null) {
             return null;
-        }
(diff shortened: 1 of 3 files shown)


vladbailescu:
@ky940819, WDYT?

ky940819:
> @ky940819, WDYT?

Hello, yes this should resolve the unclosed RRs.
I submitted a PR (#1121) because I already had the changes staged, I went a little further there in removing `Utils.getComponentsResolver` entirely so as to avoid any temptation to use it without closing it. But this PR should achieve the desired result

jckautzmann:
I tend to prefer the fix suggested by @ky940819 in the PR #1121 because, as I wrote in his PR, by completely removing the `Utils.getComponentsResolver` utility nobody will be tempted to use it and forget about closing the resource resolver.

PR #1121, opened 2020-07-24
Ensure service resource resolvers are closed

### LINKED ISSUE #1118: Unclosed resource resolver in `Utils.getComponentsResolver`

ky940819:
| Fixed Issues?            | Fixes #1118 
| Tests Added + Pass?      | none added

@vladbailescu Sorry I already had this staged when i saw you pushed #1120. I should have clarified that I would suggest a solution. I'll offer it up anyway if there is any interest in it.
 
- Removes utility for creating service resource resolvers with elevated permissions; requires caller to provide resource resolver
- Uses try-with-resource from caller (`ComponentFilesImpl` & `ClientLibrariesImpl`) to ensure that the resource resolver is closed, this also opens the potential to use different service users for different tasks in the future.
- Fixes a theoretical NPE in `Utils` due to page content resource potentially being null
- Reworks `ClientLibrariesImpl` a bit to have less deep nesting of conditions/loops (no functional change)

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
@@ -21,8 +21,6 @@ import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -223,13 +221,19 @@ public class Utils {
      * @return a set of the inherited resource types.
      */
     @NotNull
-    public static Set<String> getSuperTypes(@NotNull String resourceType, @NotNull final ResourceResolver resourceResolver) {
-        return Optional.ofNullable(resourceResolver.getResource(resourceType))
-            .map(Resource::getResourceSuperType)
-            .filter(StringUtils::isNotEmpty)
-            .map(superType -> Stream.concat(Stream.of(superType), getSuperTypes(superType, resourceResolver).stream()))
-            .orElseGet(Stream::empty)
-            .collect(Collectors.toSet());
+    public static Set<String> getSuperTypes(@NotNull final String resourceType, @NotNull final ResourceResolver resourceResolver) {
+        Set<String> superTypes = new HashSet<>();
+        String superType = resourceType;
+        while (superType != null) {
+            superType = Optional.ofNullable(resourceResolver.getResource(superType))
+                .map(Resource::getResourceSuperType)
+                .filter(StringUtils::isNotEmpty)
+                .orElse(null);
+            if (superType != null) {
+                superTypes.add(superType);
+            }
+        }
+        return superTypes;
     }
 
     /**
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
@@ -136,7 +136,8 @@ public class Utils {
      * @param page the {@link Page}
      * @param request the current request
      * @param modelFactory the {@link ModelFactory}
-     * @return
+     *
+     * @return The set of resource types for components used to render a page.
      */
     @NotNull
     public static Set<String> getPageResourceTypes(@NotNull Page page, @NotNull SlingHttpServletRequest request, @NotNull ModelFactory modelFactory) {
@@ -182,12 +183,9 @@ public class Utils {
         if (experienceFragment != null) {
             String fragmentPath = experienceFragment.getLocalizedFragmentVariationPath();
             if (StringUtils.isNotEmpty(fragmentPath)) {
-                ResourceResolver resolver = resource.getResourceResolver();
-                if (resolver != null) {
-                    Resource fragmentResource = resolver.getResource(fragmentPath);
-                    if (fragmentResource != null) {
-                        return getResourceTypes(fragmentResource, request, modelFactory);
-                    }
+                Resource fragmentResource = resource.getResourceResolver().getResource(fragmentPath);
+                if (fragmentResource != null) {
+                    return getResourceTypes(fragmentResource, request, modelFactory);
                 }
             }
         }
@@ -208,12 +206,9 @@ public class Utils {
         Template template = page.getTemplate();
         if (template != null) {
             String templatePath = template.getPath() + AllowedComponentList.STRUCTURE_JCR_CONTENT;
-            ResourceResolver resolver = page.getContentResource().getResourceResolver();
-            if (resolver != null) {
-                Resource templateResource = resolver.getResource(templatePath);
-                if (templateResource != null) {
-                    return getResourceTypes(templateResource, request, modelFactory);
-                }
+            Resource templateResource = request.getResourceResolver().getResource(templatePath);
+            if (templateResource != null) {
+                return getResourceTypes(templateResource, request, modelFactory);
             }
         }
         return Collections.emptySet();
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
@@ -19,30 +19,27 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
+import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sling.api.SlingHttpServletRequest;
-import org.apache.sling.api.resource.LoginException;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ResourceResolver;
-import org.apache.sling.api.resource.ResourceResolverFactory;
 import org.apache.sling.models.factory.ModelFactory;
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.adobe.cq.wcm.core.components.models.ExperienceFragment;
 import com.day.cq.wcm.api.Page;
 import com.day.cq.wcm.api.PageManager;
 import com.day.cq.wcm.api.Template;
 import com.day.cq.wcm.foundation.AllowedComponentList;
+import org.jetbrains.annotations.Nullable;
 
 public class Utils {
-
-    private static final Logger LOG = LoggerFactory.getLogger(Utils.class);
 
     /**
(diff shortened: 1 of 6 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
```diff
@@ -211,82 +206,30 @@ public static String generateId(String prefix, String path) {
...
-    private static void addSuperTypes(@NotNull String resourceType, @NotNull Set<String> superTypes, @NotNull SlingHttpServletRequest request, ResourceResolverFactory resolverFactory) {
-        Resource resource = getResource(resourceType, request, resolverFactory);
-        if (resource != null) {
-            String superType = resource.getResourceSuperType();
-            if (StringUtils.isNotEmpty(superType) && !superTypes.contains(superType)) {
-                superTypes.add(superType);
-                addSuperTypes(superType, superTypes, request, resolverFactory);
-            }
-        }
+    public static Set<String> getSuperTypes(@NotNull String resourceType, @NotNull final ResourceResolver resourceResolver) {
```
jckautzmann, line 226: I would prefer keeping the former implementation and not using the Java stream, which I find harder to debug and harder to follow with a recursion.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java
```diff
@@ -272,59 +282,62 @@ private String getInline(LibraryType libraryType) {
      *
      * @return {@link Set<String>} of clientlib categories
      */
+    @NotNull
     protected Set<String> getCategoriesFromComponents() {
-        Set<String> categories = new HashSet<>();
```
jckautzmann, line 276: Here also I would prefer keeping the former implementation and not using the Java stream, which I find harder to debug and harder to follow with a recursion.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java
```diff
@@ -272,59 +282,62 @@ private String getInline(LibraryType libraryType) {
...
-        if (StringUtils.equals(componentType, CQ_CLIENTLIBRARY_FOLDER)) {
-            ClientLibrary library = allLibraries.get(resource.getPath());
-            if (library != null) {
-                libraries.add(library);
-            }
-        }
-        for (Resource child : resource.getChildren()) {
-            addClientLibraries(child, libraries);
-        }
+    private static Stream<ClientLibrary> getClientLibraries(@org.jetbrains.annotations.Nullable final Resource resource,
```
jckautzmann, line 323: Here also I would prefer keeping the former implementation and not using the Java stream, which I find harder to debug and harder to follow with a recursion.

jckautzmann:
Thx @ky940819 for your contribution! I tend to prefer this fix to the one suggested by @vladbailescu in #1120 because by completely removing the `Utils.getComponentsResolver` utility nobody won't be tempted to use it and forget about closing the resource resolver.

As stated in the PR, I would prefer keeping the former implementation and not using the Java stream, which I find harder to debug and harder to follow, especially when used with a recursion.

ky940819:
@jckautzmann, I have updated the PR to remove the streams.
For future reference, do you not want the Java Stream API used at all in this project, or just recursive stream building?

Kyle

jckautzmann:
thx @ky940819 for updating the PR.
I'm not a big fan of the Java Stream API: I have to admit that it's easy to read and write but harder to debug and sometimes understand (like when used with recursion).

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java
```diff
@@ -272,59 +282,91 @@ private String getInline(LibraryType libraryType) {
...
-            for (String category : library.getCategories()) {
-                if (pattern != null) {
-                    if (pattern.matcher(category).matches()) {
-                        categories.add(category);
-                    }
-                } else {
-                    categories.add(category);
-                }
+    /**
+     * Get all resource types.
```
jckautzmann, line 323: I'd use the describing tense: Gets

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java
```diff
@@ -272,59 +282,91 @@ private String getInline(LibraryType libraryType) {
...
-     * Adds client libraries to the provided collection, starting from the given resource
+     * Gets a list of client libraries, starting from the given resource
      * and diving into its descendants.
      *
      * @param resource - the given resource, which will be checked to see if it's a client library
-     * @param libraries - the provided collection of libraries to add to
+     * @param allLibraries - Map of all client libraries.
+     * @return List of client libraries for the given resource.
      */
-    private void addClientLibraries(Resource resource, Collection<ClientLibrary> libraries) {
```
vladbailescu, line 314: Same as above, we used `add*` to avoid creating new collection/set on each recursion.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ComponentFilesImpl.java
```diff
@@ -27,24 +28,33 @@
...
 public class ComponentFilesImpl implements ComponentFiles {
+    private static final Logger LOG = LoggerFactory.getLogger(ComponentFilesImpl.class);
 
-    @Self
-    SlingHttpServletRequest request;
+    /**
+     * Name of the subservice used to authenticate as in order to be able to read details about components and
+     * client libraries.
+     */
+    public static final String COMPONENTS_SERVICE = "components-service";
```
vladbailescu, line 57: We could reuse previously defined constant.

ky940819: We could.
The duplication was mostly an artifact of the refactoring.
However, I decided to leave it because the `ComponentFilesImpl.COMPONENTS_SERVICE` subservice and the `ClientLibrariesImpl.COMPONENTS_SERVICE` could potentially be different subservices and i didn't want to encourage a single catch-all subservice to be used across the entire project.

The constant probably should have been renamed. The clearest way might be to actually change the subservice names to something like "componentfiles-service" and "clientlibraries-service", and then provide two mappings to indicate these are different services, that just happen to have the same permissions.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java
ky940819: How about if we remove the recursion entirely. 
This re-introduces streams, but could be done without them. It just feels natural to me to use a stream for filtering a collection:

```JAVA
    @NotNull
    private static List<ClientLibrary> getClientLibraries(@org.jetbrains.annotations.Nullable final Resource resource,
                                                          @NotNull final Map<String, ClientLibrary> allLibraries) {
        return Optional.ofNullable(resource)
            .map(Resource::getPath)
            .map(path -> allLibraries.entrySet().stream()
                .filter(entry -> entry.getKey().equals(path) || entry.getKey().startsWith(path + "/"))
                .map(Map.Entry::getValue)
                .collect(Collectors.toList()))
            .orElseGet(Collections::emptyList);
    }
```

Since we never add a client library unless it exists in the `allLibraries` collection; why not just filter that collection?
I think that filtering an existing collection will perform better than even a small tree-traversal followed by a collection lookup. This also safe-guards against super expensive traversals that would result if there is a huge tree under the provided resource.

ky940819: As a side note, this is an example of where I personally enjoy streams.
A different style of coding, but the way I would write this would be to return the stream directly, and just `flatMap` it in '#getAllClientLibraries()', which would avoid the collection until the final results.

ky940819: Also, another thought.
Do we want to filter out clientlibs that don't allow proxy?
Since we are searching /apps or /libs for client libraries and then rendering them inline, we are effectively proxying them.

If we want to do that then it's just adding a
```JAVA
    .filter(ClientLibrary::allowProxy)
```

ky940819:
@vladbailescu Do you have any thoughts on my responses? I'd love to get this cleared up. 
Thanks, kyle

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java
vladbailescu: Filtering `allLibraries` might be actually better, you're right. I would not worry about `allowProxy` as embedding a non-proxied clientlib in a proxied one will also make it available, similarly to what we are doing.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
```diff
@@ -211,82 +203,33 @@ public static String generateId(String prefix, String path) {
...
      */
     @NotNull
-    public static Set<String> getSuperTypes(@NotNull String resourceType, @NotNull SlingHttpServletRequest request, ResourceResolverFactory resolverFactory) {
+    public static Set<String> getSuperTypes(@NotNull String resourceType, @NotNull final ResourceResolver resourceResolver) {
         Set<String> superTypes = new HashSet<>();
-        addSuperTypes(resourceType, superTypes, request, resolverFactory);
-        return superTypes;
-    }
-
-    private static void addSuperTypes(@NotNull String resourceType, @NotNull Set<String> superTypes, @NotNull SlingHttpServletRequest request, ResourceResolverFactory resolverFactory) {
```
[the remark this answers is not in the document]
vladbailescu: Agree to just loop

vladbailescu:
Hi Kyle, thanks for following through on this. I tried to answer to the open points, @jckautzmann will probably help you further as I'll be away for a (short) while.

vladbailescu (approved):

PR #1125, opened 2020-08-07
[Testing] Convert all tests to use Junit5

ky940819:
| Fixed Issues?            | None at this time, ticket can be created and PR updated if the changes in this PR are desireable.
| Tests Added + Pass?      | Modified + Pass (no newly added tests)
| Any Dependency Changes?  | Test dependencies only for `core.wcm.components.core`; dependencies modified for `core.wcm.components.junit.core`

The purpose of this PR is to use a single unified testing framework (the latest).
Occasionally, developers reference previously existing tests as a blueprint for their tests - and in doing so, they may inadvertently regress to using Junit4 instead of Junit5.

By converting all tests to Junit5 (even old ones) and dropping the Junit4 dependencies, this can be avoided and all future tests will use Junit5.

This will also reduce the need to write helpers to deal with both Junit4 and Junit5 (e.g. `AemContext` has two versions).

- All unit tests using Junit4 are updated to use Junit5
- Junit4 specific test dependencies are removed from `com.wcm.components.core` project
- Junit4 dependencies are removed from the `core.wcm.components.junit.core` project. The tests now directly throw the assertion error instead of depending on a specific testing framework.
- Junit4 dependencies remain in parent dependency management because the `testing/it/http` project uses them

### DIFF
--- a/bundles/core/pom.xml
+++ b/bundles/core/pom.xml
@@ -499,23 +499,6 @@
             <scope>test</scope>
         </dependency>
 
-        <!-- Only required for junit 4 -->
-        <dependency>
-            <groupId>io.wcm</groupId>
-            <artifactId>io.wcm.testing.aem-mock.junit4</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.junit.vintage</groupId>
-            <artifactId>junit-vintage-engine</artifactId>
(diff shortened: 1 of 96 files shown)


bpauli:
@ky940819 Thank you very much for this effort! This changes will be part of the upcoming 2.12.0 release.

PR #1127, opened 2020-08-07
AmpModeForwardFilter breaks Adobe Campaign integration (fixes #1126)

### LINKED ISSUE #1126: AmpModeForwardFilter breaks Adobe Campaign integration
## Bug Report

**Current Behavior**
After installation of Core Components release 2.11.0 the standard integration with Adobe Campaign becomes non-functional.
The issue surfaces as HTTP 500 for the Adobe Campaign user when attempting to pull email page content from AEM, preventing usage of AEM content in Campaign. The root cause is as following:

- To retrieve the rendered HTML of an AEM email page, Adobe Campaign will send an HTTP POST like following to AEM:
http://_your-aem-author_/content/campaigns/_your-email-page_.**campaign**.content.json?delivery=_your-delivery-id_
- The selectors of this request are always "campaign.content". AmpModeForwardFilter erroneously manipulates this selector string and removes "amp" from it, so that the selector is changed to "caign.content".
- Once AmpModeForwardFilter passes the request along, it will not match the ContentServlet part of AEM's Adobe Campaign integration anymore because of the changed selector. Instead, it will match the generic SlingPostServlet, and trigger an HTTP 500.

Please see the attached screenshot for a Sling request processing sample demonstrating the issue.
It's worth noting that this HTTP request is the only way how AEM content can get loaded into Adobe Campaign out of the box, which means that this issue is a showstopper for all clients which use Adobe Campaign in combination with AEM and will prevent usage of this Core Components release for them.

The root cause lies here:
[AmpModeForwardFilter.java](https://github.com/adobe/aem-core-wcm-components/blob/core.wcm.components.reactor-2.11.0/extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java#L69)
```java
String selectors = slingRequest.getRequestPathInfo().getSelectorString();
// selectors is now "campaign.content"
boolean hasAmpSelector = selectors.contains(AMP_SELECTOR);
// hasAmpSelector is now true because it simply checks for "amp" anywhere in the String
String ampMode = AmpUtil.getAmpMode(slingRequest);
if (hasAmpSelector && (ampMode.equals(NO_AMP) || ampMode.isEmpty())) {
  // logic enters this condition despite AMP not being used
  String newSelectors;
  if (selectors.contains(DOT + AMP_SELECTOR)) {
    newSelectors = selectors.replace(DOT + AMP_SELECTOR, "");
  } else if (selectors.contains(AMP_SELECTOR + DOT)) {
    newSelectors = selectors.replace(AMP_SELECTOR + DOT, "");
  } else {
    // since "amp" is in the middle of a selector, it enters this else
    newSelectors = selectors.replace(AMP_SELECTOR, "");
    // newSelectors is now 'caign.content'
  }
  // [...]
}
```

**Expected behavior/code**
When there's no dedicated "amp" selector in RequestPathInfo, AmpModeForwardFilter shouldn't manipulate the selector String and passthrough the request unmodified.

**Environment**
- AEM 6.5 SP5
- Core Components version 2.11.0

**Possible Solution**
As shown above, the issue is that AmpModeForwardFilter simply checks for any presence of "amp" in the full selector string. Instead, it would be safe to loop through the requestPathInfo.getSelectors() array and compare individual selectors. It shouldn't use String.contains but String.equals or similar exact matching.
Additionally the same issue applies to the replacement logic: You'd also get this problem if you had a custom selector that starts or ends with "amp". The solution can be the same: removing a full selector only instead of a substring.

amp-filter

In case you need further info, please let me know.

Thanks!

Matthias

bpauli:
- add check for dedicated amp selector
- add unit test to test campaign selector

| Fixed Issues?            | Fixes #1126 
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Documentation Provided   | 
| Any Dependency Changes?  |

add dedicated check for amp selector and unit test to check result


extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
```diff
@@ -65,57 +66,55 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
-            // Move the amp selector to the front of the list of selectors.
-            RequestDispatcherOptions options = new RequestDispatcherOptions();
-            String newSelectors = selectors.replace(DOT + AMP_SELECTOR, "");
-            options.setReplaceSelectors(AMP_SELECTOR + DOT + newSelectors);
-
-            if (forward(slingRequest, response, options)) {
-                return;
+        if (hasAmpSelector || ampMode.equals(AMP_ONLY)) {
+
+            String selectors = StringUtils.defaultIfEmpty(slingRequest.getRequestPathInfo().getSelectorString(), StringUtils.EMPTY);
```
vladbailescu, line 75: I would rather we use https://sling.apache.org/apidocs/sling9/org/apache/sling/api/request/RequestPathInfo.html#getSelectors-- to manipulate selectors. This would simplify the logic and avoid future errors caused by string replacement. We already have these available as `selectorArray`

extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
```diff
@@ -46,7 +48,6 @@
         configurationPid = "com.adobe.cq.wcm.core.components.internal.services.amp.AmpModeForwardFilter",
         property = {
                 ServletResolverConstants.SLING_SERVLET_METHODS + "=" + HttpConstants.METHOD_GET,
-                ServletResolverConstants.SLING_SERVLET_EXTENSIONS + "=amp.html",
                 EngineConstants.SLING_FILTER_SCOPE + "=" + EngineConstants.FILTER_SCOPE_REQUEST,
                 EngineConstants.SLING_FILTER_PATTERN + "=/content/.*",
```
vladbailescu, line 52: I would restrict this to `/content/.*\.html`

extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
```diff
@@ -65,57 +66,55 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
-        } else if (selectors.contains(DOT + AMP_SELECTOR)) {
-
-            // Move the amp selector to the front of the list of selectors.
-            RequestDispatcherOptions options = new RequestDispatcherOptions();
-            String newSelectors = selectors.replace(DOT + AMP_SELECTOR, "");
-            options.setReplaceSelectors(AMP_SELECTOR + DOT + newSelectors);
-
-            if (forward(slingRequest, response, options)) {
-                return;
+        if (hasAmpSelector || ampMode.equals(AMP_ONLY)) {
```
vladbailescu, line 73: I think this extra condition makes the code difficult to read/understand

extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
```diff
@@ -64,61 +66,24 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
-                newSelectors = selectors.replace(AMP_SELECTOR + DOT, "");
-            } else {
-                newSelectors = selectors.replace(AMP_SELECTOR, "");
-            }
-
-            // Apply the updated selectors.
+        if (selectors.contains(AMP_SELECTOR) || ampMode == AMP_MODE.AMP_ONLY) {
             RequestDispatcherOptions options = new RequestDispatcherOptions();
-            options.setReplaceSelectors(newSelectors);
+            selectors = selectors.stream().filter(e -> !e.equals(AMP_SELECTOR))
```
vladbailescu, line 75: I would rather we just do `selectors.remove(AMP_SELECTOR)` here

extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
```diff
@@ -64,61 +66,24 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
 
         SlingHttpServletRequest slingRequest = (SlingHttpServletRequest) request;
+        AMP_MODE ampMode = getAmpMode(slingRequest);
 
-        // Read the full selector string from the request.
-        String selectors = slingRequest.getRequestPathInfo().getSelectorString();
-        if (selectors == null) {
-            selectors = "";
-        }
+        List<String> selectors = Arrays.asList(slingRequest.getRequestPathInfo().getSelectors());
```
vladbailescu, line 71: I would rather we create a modifiable list here: `List<String> selectors = new LinkedList(Arrays.asList(slingRequest.getRequestPathInfo().getSelectors()));`

bundles/core/src/test/resources/page/exporter-templated-page.json
```diff
@@ -15,9 +15,9 @@
             "xdm:template": "/conf/coretest/settings/wcm/templates/product-page",
             "repo:path": "/core/content/page/templated-page.html",
             "xdm:tags": [
-                "two",
                 "three",
-                "one"
+                "one",
```
vladbailescu, line 19: Why does the order change here?

extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
bpauli: `selectors.remove(AMP_SELECTOR)` will only remove a single element. What if the amp selector is provided more than once? We should make sure that all `amp` selectors get removed.

extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
bpauli: `Arrays.asList` will return a modifiable ArrayList. I don't see a benefit of a LinkedList here.

bundles/core/src/test/resources/page/exporter-templated-page.json
bpauli: The tags are saved in an unordered structure. The order has changed with the updated uber-jar and should remains stable after the update.

extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
vladbailescu: https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#asList(T...)
> Returns a fixed-size list backed by the specified array
That's read-only.

extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
vladbailescu: Ok, then we don't need the list to be writable anyway. I'll close https://github.com/adobe/aem-core-wcm-components/pull/1127#discussion_r468535568 as well.

PR #1138, opened 2020-08-14
[ExperienceFragment] Fix incorrect exported items order

### LINKED ISSUE #1135: [Experience Fragment] Method getExportedItemsOrder order is incorrect
## Bug Report

**Current Behavior**
The ExperienceFragment component was extended in PR #1117 (as part of #943).
This implementation uses a `HashMap` to store the child items.
`HashMap` does not guarantee any sort order, therefore relying on the keys of this `HashMap` to derive the exported items order leads to an ambiguous order.

The exporter test data demonstrates this issue, to pass the tests the expected exporter data has the items in the reverse order as they exist in the content.

As an additional aside that can be tackled with a little refactoring: the XF model initializer performs some heavy lifting when it constructs all the models for the children of the XF. The intent of #943 was to expand the content of the JSON serialization; however, the models are constructed even when not doing JSON serialization, which has a performance impact that can be avoided by refactoring that logic outside of the model initializer and directly into the appropriate getter methods.

**Expected behavior/code**
The expected result should be that the order reflects the JCR order.

**Environment**
- AEM 6.5.5.0
- Core Components version 2.11.1-SNAPSHOT (current development)
- JRE version JDK11

**Possible Solution**
The `ContentFragmentUtils.getComponentExporters(Iterator, ModelFactory, SlingHttpServletRequest)` uses a `LinkedHashMap`. The return value of this method should be changed from `Map` -> `LinkedHashMap` as an indication that the ordering is important. The XF model should then use a `LinkedHashMap` itself when storing the child items. By doing this the keyset can be used to determine the item ordering.

Will submit pull request!

ky940819:
| Fixed Issues?            | Fixes #1135 

- Fixes possibly incorrect exported items ordering by ensuring that the `children` models maintain the order that they are encountered in the repository.
- Corrects tests to match
- Replaces `Nonnull` annotations with `NotNull`
- Adds JavaDoc to previously undocumented members of `ExperienceFragmentImpl`, and correct several existing javadoc issues.
- Moves much of the model initializers to their respective getter methods to make it clearer how a getter return value is populated, remove the requirement that model initializer methods need to be called in specific orders for the getters to return reliable results, and to make sure that work is only done if required.
- Adds `Nullable` or `NotNull` annotations to private methods to reveal potential NPEs and correct.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/ContentFragmentUtils.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/ContentFragmentUtils.java
@@ -31,6 +31,7 @@ import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ResourceResolver;
 import org.apache.sling.api.resource.ValueMap;
 import org.apache.sling.models.factory.ModelFactory;
+import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -202,7 +203,7 @@ public class ContentFragmentUtils {
      * Returns an optional grid type configured via {@link ContentPolicy content policy} property
      * ({@value #PN_CFM_GRID_TYPE}) or {@value #DEFAULT_GRID_TYPE} as default.
      *
-     * @param fragmentResource         the content fragment resource resource to be checked
+     * @param fragmentResource the content fragment resource resource to be checked
      * @return the configured grid type of a default
      */
     public static String getGridResourceType(Resource fragmentResource) {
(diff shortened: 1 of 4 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/ContentFragmentUtils.java
```diff
@@ -235,13 +236,23 @@ public static String getGridResourceType(Resource fragmentResource) {
...
+     * @param modelFactory Model factory service.
+     * @param slingHttpServletRequest Current request.
+     * @return Ordered map of resource names to {@link ComponentExporter} models.
      */
-    public static Map<String, ComponentExporter> getComponentExporters(Iterator<Resource> resourceIterator,
-                                                                       ModelFactory modelFactory,
-                                                                       SlingHttpServletRequest slingHttpServletRequest) {
-        final Map<String, ComponentExporter> componentExporterMap = new LinkedHashMap<>();
+    @NotNull
+    public static LinkedHashMap<String, ComponentExporter> getComponentExporters(
```
jckautzmann, line 251: I'd keep the return type as it was before: `Map<String, ComponentExporter>`.
This way we keep the same pattern for the `ExperienceFragment` model:
- API: `Map<String, ? extends ComponentExporter> getExportedItems()`
- implementation: we return a `LinkedHashMap`

vladbailescu: Since this class is internal, I see no reason not to change to a better contract (`LinkedHashMap` ensures order)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ExperienceFragmentImpl.java
```diff
@@ -66,165 +60,207 @@
...
+        return this.children;
     }
 
-    @Nonnull
+    @NotNull
     @Override
     public String[] getExportedItemsOrder() {
-        return children.isEmpty() ?
-                new String[0] : children.keySet().toArray(new String[children.size()]);
+        return this.getExportedItems().keySet().toArray(new String[0]);
```
vladbailescu, line 241: `getExportedItems()` does not guarantee you get a `LinkedHashMap` that gives the desired order. I would rather we create a private/protected `getChildren` that returns a `LinkedHashMap` and takes care of populating/caching children resolution. `getExportedItems` and `getExportedItemsOrder` can then leverage that method.

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/ContentFragmentUtils.java
jckautzmann: sure we can the more precise return type `LinkedHashMap `. My suggestion was more to have consistent return types.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ExperienceFragmentImpl.java
jckautzmann: I don't see a need to introduce a `getChildren` as `children` is a `LinkedHashMap ` and the logic to populate it is based on the newly introduced `ContentFragmentUtils. getComponentExporters` which returns a `LinkedHashMap`.

Issue #916, opened 2020-02-13
[RFC] Email Components


gabrielwalt:
A set of core email components that render email-compliant HTML output to deliver the content via Adobe Campaign.

User Story: As an author, I want to create emails in AEM from an editable template and a set of core components so that I have the flexibility to assemble a layout that fits my needs, but also enough governance to not deviate from the brand guidelines. I then want to send out these emails via Adobe Campaign.

# Phase 1

Basic email editing capability.

## CSS Inliner Rewriter

Takes the HTML output of the page and inlines all CSS in as HTML attributes:
* Rewrite the output to move every CSS ruleset to the targeted elements as an HTML `style` attribute.
  - Properly handle quotes, like: `font-family: 'Timmana', "Gill Sans", sans-serif;`
  - Accept also browser-specific rules, like: `-ms-interpolation-mode: bicubic;`
  - Preserve duplicate properties, like: `display: inline-block; display: block;`
  - Preserve capitalization as-is, like: `Margin: 0px;`
* Keep the original `style` element in the HTML as-is in order to:
  - Preserve media queries, web fonts, and animations by not removing from the style element any “@” declaration.
  - Preserve unmatched CSS rules, like the Outlook `.ExternalClass` class name.
  - Preserve CSS pseudo-selectors, like `a:hover` or `p::before`.
* Leave client-specific comments untouched, like:<br/>`<!--[if mso]><style>.hide-mso { display: none; }</style><![endif]-->`

## Page component

Provides the following edit dialog options:
* Main tab
  - Email subject (`jcr:title`)
  - Pre-header (RTE with capability to use the Campaign personalization fields)
  - Description
  - Language + language root
  - Tags
  - HTML ID
* Advanced
  - Cloud configuration
  - Template settings
  - Authentication requirement
  - Export configuration
* Thumbnail

Provides the following page policy (aka design dialog) options:
* Main tab
  - Client libraries
* Styles tab

Requirements:
* Extends the base or core page component.
* Functions with Editable Templates.
* Inlines all CSS into the HTML output on a STYLE element that is then further inlined as STYLE attributes on the individual elements by the CSS inliner rewriter.
* Includes no JS from any of the included clientlibs.
* Pre-header adds the entered content at the top of the page body as follows:<br>`<span style="display: none; …">${page.preheader}</span>`

## Container component

Provides a simplified responsive grid behavior that is made out of 6 columns.

Provides the following edit dialog options:
* Main tab
[...]

PR #1143, opened 2020-08-21
PoC to inline Styles with CSS properties for E-Mail Templates *Do not Merge*

bpauli:
- add unit Test with possible logic to parse the inline styles and to add style attributes to matched elements


bpauli:
PoC for the CSS inlining which should be part of #916

vladbailescu:
This should probably be contributed to https://adobe-consulting-services.github.io/acs-aem-commons/features/e-mail/stylesheet-inliner/index.html as an optional way to inline CSS style to elements' `style`

Issue #1146, opened 2020-08-24
[Breadcrumb] Cannot edit breadcrumb in template editor


cjelger:
## Bug Report

**Current Behavior**
When it's added to a page template, the breadcrumb component cannot be edited in the template editor and actually "breaks" the rendering of the template in the Editor because it throws a NPE.

**Expected behavior/code**
The component should be editable in the template editor.

**Environment**
Latest AEM Cloud SDK.

**Possible Solution**
A brute-force solution is to simply ignore all pages in the template editor and add
```
if (!currentPage.getPath().startsWith("/content")) {
    return Collections.emptyList();
}
```
at the start of the `getItems()` method.

jckautzmann:
I was able to reproduce it and observed following error:
```
Caused by: java.lang.NullPointerException: null
    at com.adobe.cq.wcm.core.components.internal.models.v1.AbstractListItemImpl.<init>(AbstractListItemImpl.java:40) [com.adobe.cq.core.wcm.components.core:2.11.1.SNAPSHOT]
    at com.adobe.cq.wcm.core.components.internal.models.v1.PageListItemImpl.<init>(PageListItemImpl.java:53) [com.adobe.cq.core.wcm.components.core:2.11.1.SNAPSHOT]
    at com.adobe.cq.wcm.core.components.internal.models.v1.NavigationItemImpl.<init>(NavigationItemImpl.java:38) [com.adobe.cq.core.wcm.components.core:2.11.1.SNAPSHOT]
    at com.adobe.cq.wcm.core.components.internal.models.v1.BreadcrumbItemImpl.<init>(BreadcrumbItemImpl.java:33) [com.adobe.cq.core.wcm.components.core:2.11.1.SNAPSHOT]
    at com.adobe.cq.wcm.core.components.internal.models.v1.BreadcrumbImpl.createItems(BreadcrumbImpl.java:105) [com.adobe.cq.core.wcm.components.core:2.11.1.SNAPSHOT]
    at com.adobe.cq.wcm.core.components.internal.models.v1.BreadcrumbImpl.getItems(BreadcrumbImpl.java:83) [com.adobe.cq.core.wcm.components.core:2.11.1.SNAPSHOT]
```

jckautzmann:
I created a fix:
https://github.com/adobe/aem-core-wcm-components/pull/1147

jckautzmann:
The NPE is thrown when one of the page (breadcrumb item) does not have a `jcr:content` node.

PR #1147, opened 2020-08-24
GH-1146 - Breadcrumb component cannot be edited in page template editor

jckautzmann:
- add null check
- don't create a breadcrumb item when the corresponding page does not have a jcr:content node
- add tests


vladbailescu (approved):

vladbailescu:
Would be nice to also add an unit test for this

msagolj (approved):
Verified on local installation

jckautzmann:
@vladbailescu I made the breadcrumb item creation more robust and added a test in https://github.com/adobe/aem-core-wcm-components/pull/1147/commits/45f296d4ce8ca92b5943dd3640b8ed63b92587bf

vladbailescu (approved):

PR #1149, opened 2020-08-24
Fixes #1148 Make sure allowed components are displayed in template editor

### LINKED ISSUE #1148: [Container] No allowed components displayed for locked container if layout in policy is responsive grid
Since #1094 it's possible to set the container layout in content policy. If this feature is used and the container layout is set to "responsive grid", no allowed components are displayed in template editor.

Example of locked container component in template editor with layout in policy set to "Responsive Grid":
policy-responsive-grid

Example of locked container component in template editor with layout in policy set to "Simple" (or default):
policy-simple

stefanseifert:
Fixes #1148

This PR effectively solves the problem described in #1148, but in a unelegant way: it calls `com.day.cq.wcm.foundation.AllowedComponentList` twice, and applies its CSS to the wrapper element twice. The source of the problem that in responsive grid mode there are two wrapper elements around the container content, and the `aem-AllowedComponent--list` CSS is set on the inner one of those two.

A cleaner solution would be to no longer make a delegation to `wcm/foundation/components/responsivegrid/responsivegrid` but instead incorporate the logic of `/libs/wcm/foundation/components/responsivegrid/responsivegrid.html` and making sure there is only one wrapper element. But with this change updates to the AEM OOTB responsive grid HTL will no longer be reflected in the core component - not sure if this is a problem.

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/container/v1/container/responsiveGrid.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/container/v1/container/responsiveGrid.html
@@ -14,8 +14,9 @@
   ~ limitations under the License.
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
 <template data-sly-template.responsiveGrid="${ @ container}">
-    <div id="${container.id}"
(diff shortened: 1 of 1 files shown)


vladbailescu:
Please see https://github.com/adobe/aem-core-wcm-components/issues/1148#issuecomment-693399483

We plan to fix this at product level

PR #1161, opened 2020-09-03
[Image] AdaptiveImageServlet throws OOM when handling very large images

### LINKED ISSUE #1160: [Image] AdaptiveImageServlet throws OOM when handling very large images
The `AdaptiveImageServlet` will throw OOM when handling very large images. We should add some safeguards to make sure this not happens.

vladbailescu:
Fixes #1160

* Added configuration option for maximum processed width
* Added support for choosing a smaller rendition for some transformations
* Added safeguard to prevent processing images/renditions larger than the configured max width
* Added tests and test image


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -437,19 +440,77 @@ private void resizeAndStreamLayer(SlingHttpServletResponse response, Layer layer
...
+        renditions.addAll(asset.getRenditions());
+        EnhancedRendition bestRendition = null;
+        // Find first rendition that has a width larger or equal than wanted
         for (Rendition rendition : renditions) {
-            if (rendition.getName().startsWith(DamConstants.PREFIX_ASSET_WEB)) {
-                return rendition;
+            EnhancedRendition enhancedRendition = new EnhancedRendition(rendition);
+            Dimension dimension = enhancedRendition.getDimension();
+            if (dimension != null) {
+                if (dimension.getWidth() >= width) {
```
bpauli, line 473: Instead of throwing an `IOException` later in the filter call this method should already check if the requested with is already larger than the the maxInputWidth and should return the smales rendition which is larger than the requested width or maxInputWidth.
I would rather change this line to `dimension.getWidth() >= Math.min(width, maxInputWidth)

bpauli (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
vladbailescu: The filter is used for both _bestRendition_ and _original_ (when no transformation is needed). As it is now, `getBestRendtion` will always return something (falling back to _original_). If we do like you proposed and there's no rendition between `width` and `maxInputWidth` it would have to return `null` and we would need to add checks throughout the rest of the logic, which would make the code more complex and harder to read, IMO.

Also, I think throwing a `IOException` is good and meaningful as it will signal there's an error with reading that image/rendition (it's too large).

bpauli: Okay, I got it. LGTM!

bpauli (approved):

Issue #684, opened 2019-08-07
[Teaser] Configure title type on component dialog


bmalecki:
## Feature Request

**Is your feature request related to a problem? Please describe.**
Teaser component allows authors to select `Title Type` only on policy dialog.

**Describe the solution you'd like**
As an author I would like to have the possibility to change `Title Type` on component dialog. Please consider adding this field.

**Are there alternatives?**
- Adding styles in the policy (requires writing some specific css classes)
- Creating templates with different polices
- Creating proxy components form Core Components Teaser

nathanmckean:
Agree. This functionality should be universal across any components using headings (h1-6) not just teasers. The heading that I need is contextual to where it sits in the page. Placed under an H2, the titles should be H3. Placed under an H3, the titles should be H4, etc.

gabrielwalt:
+1, this would make sense! Your contribution is welcome if you have the possibility to create a PR for improving this.

nathanmckean:
Im an UX guy and a Github noob. I know what PR stands for, but that's about it sorry :(

cjelger:
Fixed by #1199

PR #1199, opened 2020-10-05
CIF-1660 - Configurable header tag for teaser component

cjelger:
- add "titleType" property in component dialog so it's possible to override the default tag

| Fixed Issues?            | CIF-1660, #684
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

This PR extends the component dialog of the Teaser component with the ability to select/override the tag for the title element. See the screenshot.

Screenshot 2020-10-05 at 10 46 17

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
@@ -26,7 +26,6 @@ import java.util.stream.StreamSupport;
 import javax.annotation.PostConstruct;
 import javax.inject.Inject;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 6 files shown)


jckautzmann:
Could you please add a link to the issue CIF-1660, or if it's an internal issue, create an issue in this repo with a description of the feature (or your use case) for future references?

cjelger:
Just found out there is already https://github.com/adobe/aem-core-wcm-components/issues/684

jckautzmann (approved):

msagolj:
I think the idea on why this title type can only be set in the policy was exactly to prevent the author from setting arbitrary H1-H6, risking to mess the layout and in some teaser use cases this value should not be changeable by the author at all.

But i also so the other side / other usecase as described in #684 that it makes sense to set the title type in context of where it is in the page. And @gabrielwalt already gave a +1 in the issue.

So i would suggest to make this field configurable in the page policy, so we get best of both worlds. And if thats not wanted at least change the description in the policy from 'Title Type' to 'Default title Type to make it more clear that its not a mandatory value anymore.

PR #1200, opened 2020-10-05
CIF-1660 - Configurable header tag for teaser component

cjelger:
- add option to hide/display title type field in component dialog

Extends #1199 with @msagolj proposal for the dialog.

Screenshot 2020-10-05 at 13 53 09

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
@@ -115,6 +115,11 @@ public class TeaserImpl extends AbstractImageDelegatingModel implements Teaser {
     private boolean titleHidden = false;
 
     /**
+     * Flag indicating if the title type should be hidden.
+     */
+    private boolean titleTypeHidden = false;
+
+    /**
      * Flag indicating if the description should be hidden.
      */
     private boolean descriptionHidden = false;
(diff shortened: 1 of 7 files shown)


msagolj:
there is an issue with switching the toggle on and off, example:
- default type is H3
- I uncheck 'Hide title type', so that author can set a different type
- author sets H6
- I check 'Hide title type' again

=> expectation is that title goes back to default set in policy , H3, but it stays on H6, with no way to change it.

cjelger:
@msagolj Good point, I was too fast and didn't realise how this worked for other values. I fixed it, please check again.

msagolj:
@cjelger thx, works now.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
```diff
@@ -114,6 +114,11 @@
      */
     private boolean titleHidden = false;
 
+    /**
+     * Flag indicating if the title type should be hidden.
+     */
+    private boolean titleTypeHidden = false;
```
msagolj, line 120: @cjelger to be backward compatible the toggle should be enabled by default, so shouldn't this be set to` true` ?

content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/_cq_design_dialog/.content.xml
```diff
@@ -62,6 +62,12 @@
                                         text="Hide title"
                                         name="./titleHidden"
                                         value="{Boolean}true"/>
+                                    <titleType
+                                        jcr:primaryType="nt:unstructured"
```
msagolj, line 66: @cjelger to be backward compatible the toggle should be enabled by default, so shouldn't this be set enabled by default with `checked =true` ?

msagolj (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
cjelger: Unfortunately this doesn't work, because when the field is not checked in the policy, the editor doesn't save the field as `false` in the JCR, but removes it completely so it ends up being `true` if we set `true` by default. Do you maybe know how to enforce that a `false` value is stored?

cjelger: Found it, there is an `uncheckedValue` property. I'll fix that.

content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/_cq_design_dialog/.content.xml
cjelger: @msagolj Please check again, as discussed on Slack, I had to "invert" the condition because it's not possible to set the checkbox to default to `true` ... :-(

msagolj (approved):

PR #1204, opened 2020-10-08
Post handler that forwards submitted data as JSON

### LINKED ISSUE #1154: [Form] Post handler that forwards submitted data as JSON
## Feature Request

Given that storing user-generated data on AEM isn't desired anymore, we need a form POST handler that forwards the submitted form data as JSON to a fully qualified HTTPS URL.

This also relates to #676.

bpauli:
- add new form action type to forward form data as JSON to remote service via http post

fixes #1154


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/form/FormPostServiceImpl.java
```diff
@@ -0,0 +1,191 @@
...
+            if (StringUtils.isNotEmpty(formEndpoint)) {
+                processFormApiSuccess = callFormApi(formEndpoint, jsonRequest);
+            }
+            sendRedirect(valueMap, request, response, processFormApiSuccess);
+        }
+        return processFormApiSuccess;
+    }
+
+    private void sendRedirect(ValueMap valueMap, SlingHttpServletRequest request, SlingHttpServletResponse response, boolean processFormApiSuccess) {
+        String redirect = valueMap.get("redirect", String.class);
```
kwin, line 107: You have to check the redirect parameter. It should not redirect to other servers.

msagolj:
This new action type should be documented in the README of the component e.g. how to set timeouts.

msagolj:
There is no input field to set the redirect/Thank you page for this action type :

image

Question: shouldn't there also be a different redirect in case the submit fails ?

msagolj:
Check that its a valid URL entered or fail gracefully, e.g. when adding a space in front you get:

image

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/FormRpcServlet.java
```diff
@@ -0,0 +1,53 @@
...
+
+@Component(
+        service = Servlet.class,
+        property = {
+                "sling.servlet.methods=" + HttpConstants.METHOD_POST,
+                "sling.servlet.resourceTypes=" + FormRpcServlet.RESOURCE_TYPE,
+                "sling.servlet.selectors=" + FormRpcServlet.SELECTOR
+        }
+)
+public class FormRpcServlet extends SlingAllMethodsServlet {
```
jckautzmann, line 40: I'd call it `FormActionRpcServlet` as it relates to the RCP form action.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/form/FormPostService.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.SlingHttpServletResponse;
+
+/**
+ * A service that posting form data to a remote service.
+ *
+ * @since com.adobe.cq.wcm.core.components.services.form 1.0.0
+ */
+public interface FormPostService {
```
jckautzmann, line 26: I'd use a more generic name, e.g. `FormHandler`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/form/FormPostService.java
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Send form data to a remote servcie.
+     *
+     * @param request the {@link SlingHttpServletRequest}
+     * @param response the {@link SlingHttpServletResponse}
+     * @return true if the remote request was successful, otherwise false
+     *
+     * @since com.adobe.cq.wcm.core.components.services.form 1.0.0
+     */
+    boolean sendFormData(SlingHttpServletRequest request, SlingHttpServletResponse response);
```
jckautzmann, line 37: I'd rename it `forwardFormData` as it is forwarding form data to an external service.

jckautzmann: When forwarding the form data to the external service, I would expect following arguments to also be present in the method signature:
- the form data, e.g. `JSONObject formData`. We could introduce a FormUtils class to retrieve the form data from the request
- the configuration of the external service, e.g. `String endPointUrl`. We could have a FormUtils method to retrieve it.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/form/FormPostServiceImpl.java
```diff
@@ -0,0 +1,191 @@
...
+        try {
+            response = client.execute(post);
+        } catch (IOException e) {
+            LOG.error(e.getMessage());
+        }
+        StatusLine responseStatusLine = response != null ? response.getStatusLine() : null;
+        return responseStatusLine != null && responseStatusLine.getStatusCode() == HttpStatus.SC_OK;
+    }
+
+    private JSONObject requestParamsToJSON(SlingHttpServletRequest req) throws JSONException {
```
jckautzmann, line 143: I'd rename it `getFormData`. It could be part of the FormUtils mentioned in the PR.

content/src/content/jcr_root/apps/core/wcm/components/form/actions/rpc/_cq_dialog/.content.xml
```diff
@@ -0,0 +1,20 @@
...
+    jcr:primaryType="nt:unstructured"
+    jcr:title="Form Container"
+    sling:resourceType="granite/ui/components/coral/foundation/container">
+    <items jcr:primaryType="nt:unstructured">
+        <formEndPointUrl
+            jcr:primaryType="nt:unstructured"
+            sling:resourceType="granite/ui/components/coral/foundation/form/textfield"
+            fieldLabel="Endpoint"
+            fieldDescription="Endpoint URL of the RESTful service"
+            name="./formEndPointUrl"
```
jckautzmann, line 13: I'd call it `externalServiceEndPointUrl` to make it clear that it is the end point of an external service.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/form/FormPostServiceImpl.java
```diff
@@ -0,0 +1,191 @@
...
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.metatype.annotations.AttributeDefinition;
+import org.osgi.service.metatype.annotations.Designate;
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.adobe.cq.wcm.core.components.services.form.FormPostService;
+import com.day.cq.wcm.foundation.forms.FormsHelper;
+import com.google.common.collect.ImmutableMap;
```
jckautzmann, line 51: this import is not used and can be removed.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/form/FormPostServiceImpl.java
```diff
@@ -0,0 +1,191 @@
...
+        try {
+            JSONObject jsonRequest = requestParamsToJSON(request);
+            return processFormApi(request, response, jsonRequest);
+        } catch (JSONException e) {
+            LOG.error(e.getMessage());
+        }
+        return false;
+    }
+
+    private boolean processFormApi(SlingHttpServletRequest request, SlingHttpServletResponse response, JSONObject jsonRequest) {
```
jckautzmann, line 92: I don't like to have `api` in the method name, I'd call it `processFormData` instead.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/form/FormPostServiceImpl.java
```diff
@@ -0,0 +1,191 @@
...
+                response.sendRedirect(request.getResourceResolver().map(request, redirect));
+            } else {
+                FormsHelper.redirectToReferrer(request, response, new HashMap<>());
+            }
+        } catch (IOException var13) {
+            LOG.error("Error redirecting to {}", redirect);
+        }
+    }
+
+    private boolean callFormApi(String formEndpointUrl, JSONObject jsonRequest) {
```
jckautzmann, line 122: Same as above, I wouldn't use `api` in the name, e.g. `forwardDataToEndPoint`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/FormRpcServlet.java
jckautzmann: Please add some javadocs to explain what it does.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/form/FormPostService.java
```diff
@@ -0,0 +1,38 @@
...
+ ~ See the License for the specific language governing permissions and
+ ~ limitations under the License.
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+package com.adobe.cq.wcm.core.components.services.form;
+
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.SlingHttpServletResponse;
+
+/**
+ * A service that posting form data to a remote service.
```
jckautzmann, line 22: should be: `...that posts...`

content/src/content/jcr_root/apps/core/wcm/components/form/actions/rpc/.content.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jcr:root xmlns:sling="http://sling.apache.org/jcr/sling/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
+    jcr:primaryType="sling:Folder"
+    jcr:title="Post form data"
+    sling:resourceType="foundation/components/form/action"
+    hint="Post form data as JSON to a REST endpoint"
```
jckautzmann, line 6: As it is a description, I'd use: `Posts`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/form/FormPostService.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @since com.adobe.cq.wcm.core.components.services.form 1.0.0
+ */
+public interface FormPostService {
+
+    /**
+     * Send form data to a remote servcie.
```
jckautzmann, line 29: I'd use the descriptive tense (sends)
servcie -> service

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/form/FormPostServiceImpl.java
bpauli: I don't think that this is required as in L119 the redirect path is mapped and will return null if the given path can't be mapped to an internal resource.

kwin: This only returns null if there is no mapping, but the default mapping will always return the same path. The javadoc explicitly says:
bq. Note, that technically the resourcePath need not refer to an existing resource. This method just applies the mappings and returns the resulting string. (https://github.com/apache/sling-org-apache-sling-resourceresolver/blob/master/src/main/java/org/apache/sling/resourceresolver/impl/ResourceResolverImpl.java#L420).

jckautzmann (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/form/FormPostServiceImpl.java
kwin: Compare also with https://issues.apache.org/jira/browse/SLING-9850!

bpauli: added test for internal resource in 88bc244100983582704ffa7992390d717b9da37c

vladbailescu (approved):

PR #1218, opened 2020-10-20
When publishing a page, its context-aware configuration should also b…

### LINKED ISSUE #1217: When publishing a page, its context-aware configuration should also be published
Now that we are able to include custom CSS, JS and metadata tags to a page (#1202) through a context-aware config, this config should also be published when publishing the page.

jckautzmann:
…e published

- use a reference provider to get the configuration node: AEM will automatically add this reference to the ones that will be published
- add tests

| Fixed Issues?            | Fixes #1217
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
@@ -28,6 +28,8 @@ import com.adobe.cq.wcm.core.components.config.HtmlPageItemsConfig;
 import com.adobe.cq.wcm.core.components.internal.DataLayerConfig;
 import com.adobe.cq.wcm.core.components.internal.services.pdfviewer.PdfViewerCaConfig;
 import com.day.cq.commons.jcr.JcrConstants;
+import com.day.cq.wcm.api.Page;
+import com.day.cq.wcm.api.PageManager;
 import com.day.cq.wcm.api.reference.Reference;
 import com.day.cq.wcm.api.reference.ReferenceProvider;
 
@@ -52,6 +54,15 @@ public class CaConfigReferenceProvider implements ReferenceProvider {
     @Override
     public List<Reference> findReferences(Resource resource) {
         List<Reference> references = new ArrayList<>();
+        // If the resource is not part of a page: stop the processing
+        PageManager pageManager = resource.getResourceResolver().adaptTo(PageManager.class);
+        if (pageManager == null) {
+            return references;
+        }
+        Page page = pageManager.getContainingPage(resource);
+        if (page == null) {
+            return references;
+        }
         addCaConfigReference(HtmlPageItemsConfig.class.getName(), resource, references);
         addCaConfigReference(DataLayerConfig.class.getName(), resource, references);
         addCaConfigReference(PdfViewerCaConfig.class.getName(), resource, references);
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
@@ -25,9 +25,13 @@ import org.apache.sling.caconfig.resource.ConfigurationResourceResolver;
 import org.osgi.service.component.annotations.Component;
 
 import com.adobe.cq.wcm.core.components.config.HtmlPageItemsConfig;
+import com.adobe.cq.wcm.core.components.internal.DataLayerConfig;
+import com.adobe.cq.wcm.core.components.internal.services.pdfviewer.PdfViewerCaConfig;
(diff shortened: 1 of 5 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * The @{@link ConfigurationResourceResolver} service.
+     */
+    @org.osgi.service.component.annotations.Reference
+    private ConfigurationResourceResolver configurationResourceResolver;
+
+    @Override
+    public List<Reference> findReferences(Resource resource) {
+        List<Reference> references = new ArrayList<>();
+        Resource configResource = configurationResourceResolver.getResource(resource, NN_SLING_CONFIGS, HtmlPageItemsConfig.class.getName());
```
bpauli, line 52: The CaConfig lookup should only be done for the containing page of the requested resource like it is done in the ContentPolicyReferenceProvider.

bpauli (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
jckautzmann: When publishing a page, the `resource` argument is already the page to be published.
And the lookup `configurationResourceResolver.getResource` will return the same config node, whether `resource` is a page or a component.

vladbailescu, line 52: We should expand this to cover all context-aware configurations shipped with Core Components: datalayer and PDF Viewer.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Provides Context Aware Configuration references for a given resource.
+ */
+@Component(
+        service = ReferenceProvider.class
+)
+public class CaConfigReferenceProvider implements ReferenceProvider {
+
+    private static final String NN_SLING_CONFIGS = "sling:configs";
```
vladbailescu, line 40: I would make this constant public in ComponentUtils so it can be reused throughout the project.

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
jckautzmann: addressed in commit https://github.com/adobe/aem-core-wcm-components/commit/2adf80cbc76e5d6213db21cbff1c3f45a9dddcaf

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
jckautzmann: addressed in commit https://github.com/adobe/aem-core-wcm-components/commit/2adf80cbc76e5d6213db21cbff1c3f45a9dddcaf

vladbailescu (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
bpauli: The `ActivationReferenceSearchServlet` is called not only when a page is published. For example it's also called when a DAM Asset is activated. In this case a CaConfig lookup should be avoided.

jckautzmann: addressed in https://github.com/adobe/aem-core-wcm-components/pull/1218/commits/3943368a45b77d03d0713180f3b531396b6568c8

bpauli (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/CaConfigReferenceProvider.java
bpauli: It would be nice to have a generic approach for this so this reference provider doesn't need an update each time we add a new CaConfig interface.

jckautzmann: good idea. I don't have time to work on it now and it's a change that can be implemented later. I'll merge as it is now and we'll tackle it later.

PR #1228, opened 2020-10-28
Add missing cq:isContainer property for example components

bpauli:
add missing `cq:isContainer` for example components as this property is not inherited in AEM 6.4


msagolj (approved):
Verified with 6.4.8.2

PR #1240, opened 2020-11-04
Enable ES6 for image client library

### LINKED ISSUE #1239: [Image] Clientlib generates error with default gcc setting
When precompiled using the default compiler settings, the compiler reports the following error during the minification of this library:

/libs/core/wcm/components/image/v2/image/clientlibs/site.min.js:179 - ERROR - This language feature is only supported for ECMASCRIPT6 mode or better: const declaration.

bpauli:
fixes #1239

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/.content.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/.content.xml
@@ -3,4 +3,4 @@
     jcr:primaryType="cq:ClientLibraryFolder"
     allowProxy="{Boolean}true"
     categories="[core.wcm.components.image.v2]"
-    jsProcessor="[default:none,min:gcc;languageIn=ECMASCRIPT_2015]"/>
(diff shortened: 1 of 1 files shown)


msagolj (approved):

content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/.content.xml
```diff
@@ -2,4 +2,5 @@
 <jcr:root xmlns:cq="http://www.day.com/jcr/cq/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
     jcr:primaryType="cq:ClientLibraryFolder"
     allowProxy="{Boolean}true"
-    categories="[core.wcm.components.image.v2]"/>
+    categories="[core.wcm.components.image.v2]"
+    jsProcessor="[default:none,min:gcc;languageIn=ECMASCRIPT_2015]"/>
```
kwin, line 6: With this GCC will generate unnecessary polyfills as long as the languageOut will not be modified as well (https://github.com/google/closure-compiler/wiki/ECMAScript6). You should probably set languageOut as well to ES6.

bpauli: good point - added in last commit

PR #1241, opened 2020-11-04
[Image] Adaptive image with quality + width selector does not maintain aspect ratio

### LINKED ISSUE #1237: [Image] Adaptive image with quality + width selector does not maintain aspect ratio
## Bug Report

**Current Behavior**
In version 2.12.0 the adaptive image servlet changes the aspect ratio of images to make them square when served with a quality+width selector. This causes the images to be severely distorted. 

**Expected behavior/code**
It should not do this. The image aspect ratio should always be maintained (assuming no crop, or other transformations are applied). 

**Environment**
- AEM version: 6.5.6.0
- Core Components version: 2.12.0
- JRE version: JDK 11

Steps to reproduce:

1. Go to /content/core-components-examples/library/page-authoring/image.html
2. Open the template editor for this page
3. On the template, edit the "Image" component policy
4. Add image quality (e.g. 82) and width selectors (e.g. 100, 1200).
5. Save the image policy component
6. Return to /content/core-components-examples/library/page-authoring/image.html
7. See that the images are now square and distorted.

See attached screenshot, where the left is the image prior to adding quality+width selectors, the right is the image after adding these policy settings.
Screenshot from 2020-11-03 13-22-32

vladbailescu:
- Fixed regression
- Added rectangular test image and tests using both original and web renditions

Fixes #1237


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -339,11 +339,8 @@ private void transformAndStreamAsset(SlingHttpServletResponse response, ValueMap
...
                 Dimension dimension = rendition.getDimension();
-                if (dimension != null) {
-                    originalWidth = dimension.width;
-                    originalHeight = dimension.height;
-                }
-                if (originalHeight > resizeWidth) {
+                if ((dimension != null && dimension.width > resizeWidth) ||
+                    (dimension == null && originalWidth > resizeWidth)) {
                     int resizeHeight = calculateResizeHeight(originalWidth, originalHeight, resizeWidth);
                     if (resizeHeight > 0 && resizeHeight != originalHeight) {
```
bpauli, line 345: The case where `resizeHeight == 0` is currently missing because of the added if statement

bpauli (changes requested):
-

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
vladbailescu: That's another issue than the aspect ration but well spoted!!!

msagolj (approved):
Tested with:
- square, portait and landscape
- jpg and png (gif and svg are rendered as is)
- with highest width set higher/lower then the original image
- with extremely big/small width values
works as expected

bpauli (approved):

PR #1248, opened 2020-11-06
Feature/expose wcmmode in datalayer

### LINKED ISSUE #1171: Expose WCMMode and run mode in dataLayer's page object
Currently only the data from https://github.com/adobe/aem-core-wcm-components/blob/master/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/datalayer/PageDataImpl.java is exposed on the dataLayer per page. Sometimes it is desirable to know whether a page has been loaded in edit/preview or disabled mode. For that an additional boolean attribute like `isAuthorMode` on the data layer would be really helpful. That should return `true` for any `WCMMode` but Disabled.

In addition an attribute exposing the run modes or a boolean attribute `isAuthor` would be good to conditionally do something only on the publish with Adobe Launch.

kwin:
| Fixed Issues?            | Fixes #1171

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageImpl.java
@@ -319,6 +319,7 @@ public class PageImpl extends AbstractComponentImpl implements Page {
     @Override
     @NotNull
     protected final PageData getComponentData() {
+        boolean isAuthor = slingSettings.getRunModes().contains("author");
         return DataLayerBuilder.extending(super.getComponentData()).asPage()
             .withTitle(this::getTitle)
             .withTags(() -> Arrays.copyOf(this.keywords, this.keywords.length))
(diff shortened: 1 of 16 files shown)


kwin:
I would like to gather some feedback whether the data layer should hide the underlying AEM details and expose these values in a more CMS agnostic way or you guys would be fine with having very AEM specific values in there as well.

vladbailescu:
To me this sounds overly specific to the current implementation of AEM editing functionality. Instead, I would rather we only expose something like `mode=[view|edit]` in the page data.

kwin:
`mode=[view|edit]` does not allow to distinguish between Author (with WCMMode=Disabled) and Publish. Are you fine with exposing the run modes next to it?

ky940819:
If these values are going to be included in the data layer, could optionss be added to the CAConfig to enable them.; For on-prem instances the runmodes might contain more information than just author/publish + environment. these values have never been exposed before, so it might cause problems if clients aren't aware that this information will suddenly become public without their explicit action.

kwin:
Maybe just `mode=[preview|edit|live]` would be sufficient, where the first two would only be ever active on Author and the latter only on Publish. WDYT? I agree that just exposing all run modes without filtering is probably not a good idea.

kwin:
@vladbailescu WDYT about this updated PR, I now only expose mode with the three values from above. The only thing unclear is basically https://github.com/adobe/aem-project-archetype/issues/528#issuecomment-725612202.

vladbailescu:
Needs:
- Resolving merge conflicts
- Verification

kwin:
@vladbailescu Do you want me to resolve the conflict, or are you gonna take it up internally?

vladbailescu:
@kwin any help is appreciated. We plan to go through all the outstanding PRs this week and merge them, then see what else we can include for 2.14.0.

PR #1253, opened 2020-11-11
[Search] Fix unclosed resource resolver in SearchServlet + other refactoring

### LINKED ISSUE #1249: Unclosed ResourceResolver in SearchResultServlet
## Bug Report

**Current Behavior**
When a search is performed using the SearchResultServlet a `ResourceResolver` is left open.
This is a known issue when using the `QueryBuilder` service. 

**Expected behavior/code**
There should be no unclosed `ResourceResolver` and no stacktrace should be shown in the logs.

**Environment**
- AEM version and patch level: 6.5.6.0
- Core Components version: 2.12.1
- JRE version: 11

**Possible Solution**
A best-attempt should be made to close the `ResourceResolver` once the query is complete.

Log:

> 25020-07.11.2020 16:35:48.765 *INFO* [Apache Sling Resource Resolver Finalizer Thread] org.apache.sling.resourceresolver.impl.CommonResourceResolverFactoryImpl Unclosed ResourceResolver was created here: 
25021-java.lang.Exception: Opening Stacktrace
25022-	at org.apache.sling.resourceresolver.impl.CommonResourceResolverFactoryImpl$ResolverReference.<init>(CommonResourceResolverFactoryImpl.java:540) [org.apache.sling.resourceresolver:1.6.8]
25023-	at org.apache.sling.resourceresolver.impl.CommonResourceResolverFactoryImpl.register(CommonResourceResolverFactoryImpl.java:235) [org.apache.sling.resourceresolver:1.6.8]
25024-	at org.apache.sling.resourceresolver.impl.ResourceResolverImpl.<init>(ResourceResolverImpl.java:104) [org.apache.sling.resourceresolver:1.6.8]
25025-	at org.apache.sling.resourceresolver.impl.ResourceResolverImpl.<init>(ResourceResolverImpl.java:97) [org.apache.sling.resourceresolver:1.6.8]
25026-	at org.apache.sling.resourceresolver.impl.CommonResourceResolverFactoryImpl.getResourceResolverInternal(CommonResourceResolverFactoryImpl.java:280) [org.apache.sling.resourceresolver:1.6.8]
25027-	at org.apache.sling.resourceresolver.impl.CommonResourceResolverFactoryImpl.getResourceResolver(CommonResourceResolverFactoryImpl.java:190) [org.apache.sling.resourceresolver:1.6.8]
25028-	at org.apache.sling.resourceresolver.impl.ResourceResolverFactoryImpl.getResourceResolver(ResourceResolverFactoryImpl.java:98) [org.apache.sling.resourceresolver:1.6.8]
25029-	at com.day.cq.search.impl.builder.QueryBuilderImpl.createResourceResolver(QueryBuilderImpl.java:210) [com.day.cq.cq-search:5.12.16]
25030-	at com.day.cq.search.impl.builder.QueryImpl.getResourceResolver(QueryImpl.java:240) [com.day.cq.cq-search:5.12.16]
25031-	at com.day.cq.search.impl.result.HitImpl.getResource(HitImpl.java:108) [com.day.cq.cq-search:5.12.16]
25032-	at com.adobe.cq.wcm.core.components.internal.servlets.SearchResultServlet.getResults(SearchResultServlet.java:203)
25033-	at com.adobe.cq.wcm.core.components.internal.servlets.SearchResultServlet.doGet(SearchResultServlet.java:101)

ky940819:
| Fixed Issues?            | Fixes #1249 

The primary purpose of this pull request is to address the unclosed resource resolver in the `SearchServlet` implementation, as documented in #1249. However, several other issues were discovered upon testing that have also been resolved.

In addition to the resource resolver not being closed, the following issues were discovered:

The testing was not complete, and did not test certain cases such as the query string being below the minimum allowed length, the IDs of the search results, the behavior of the servlet when no component is specified in the request suffix, Additionally, the test for a Search component used in a template was not actually testing that use case because the `searchRoot` property was missing from the component as used in the template, and therefore the test was falling back on the default search behavior - which just happened to be the expected search behavior. These tests have been added or corrected as needed.

The search result servlet would produce a NPE when the search resource does not exist (or is not specified in the request suffix) and the current page is not part of a language root. The search servlet assumed that the language manager would return a non-null value for the language root, resulting in a NPE. This issue was corrected; however, the logic for determining the local search root is identical to the logic used by the Navigation component. Therefore, a new private utility class `LocalizationUtils` was added to allow this logic to be reused. This utility was made in the internal package by intention, because I think that there are several (pre-existing) issues with it and that it can be made better (possibly including similar logic found in the Experience Fragment component), after which it may be advisable to expose it as a service or utility. That is out of the scope of this pull request.

The logic for determining the search configuration properties (such as the result size, and minimum search term length) was duplicated in both the search model implementation, and the search servlet. This tightly couples the search servlet to the structure and behavior of the current search component implementation, and does not allow for the underlying logic for how those properties are determined to be changed, even if that change still adheres to the Search model interface. To reconcile this the search servlet will now construct the search component model, and then use that model to provide the options through its getter methods that are documented in the Search model interface. 

The search root page path was determined only in the search servlet. Again, this ties the search servlet to the specific case of the Search component. The search page path property belongs to the Search component, and therefore it should be determined and provided by the model for that component. To accomplish this the method `Search#getSearchRootPagePath()` was added to the Search model interface; and the logic for calculating that value was moved out of the search servlet and into the model. 

The Search model implementation was made to extend `AbstractComponent`.
Although the Search model implements the `Component` interface, it did not extend `AbstractComponent` and did not provide all methods required by the `Component` interface; such as `getId()`.
This is important because the ID for the component will now come from the component model instead of generating the same ID repeatedly for every search result, using logic specific to the search servlet. This will ensure that if the ID of the component changes for some reason, that change can be captured entirely in the Search model, and will not require an update to the search servlet to ensure that they remain in-sync. 

It is now possible for a client to create a model that implements the Search component model interface, and have the properties provided by that model be honoured by the search servlet.

An additional bug was discovered where if the `offset` query parameter was not parsable to a `long` then the servlet would cause a 500 error, because the `parseLong(String)` was not checked for a `NumberFormatException`. The servlet will now send a `400 - Bad Request` in that situation.

----
On a side note for future discussion: A `DefaultSearch` class was provided, this class implements the `Search` interface, and is used by the `SearchResultServlet` when there is no search component located at the path referenced by the request suffix, or the request suffix is blank. This reveals that the behaviour of the search servlet is slightly odd.

1. It will return search results for any page, regardless of if that page contains a search component or not
2. If a user makes a request and removes the request suffix, they are able to escape the policies that govern the use of the search component (e.g. the minimum search term length, and maximum number of results)
3. Assuming that this search servlet is allowed to be accessed (via dispatcher filters), then there is no way to prevent searches from being performed wherever the user determines. IMHO the search servlet should only respond to requests for an actual search component, or at the very least a CA-Config to enable/disable searches outside the context of a search component
4. If a search is performed outside the scope of a specific search component then the search root will be the current language root, if a language root exists, otherwise it will be the current page path. This is confusing because if your site happens to have a language root (e.g. `/content/site/de`) then all of the global searches that can be performed from any page will have the exact same results (the search root is always the language root). However, if your site does not have a language root (e.g. `/content/site`) then every one of those searches (that can be performed from any page) will be different (the search root is the current requested page). This seems like a significant difference in the behaviour of this servlet simply by the presence, or lack thereof, of a single page in the tree.


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Search.java
```diff
@@ -79,6 +80,17 @@ default String getRelativePath() {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Gets the path of the search root.
```
jckautzmann, line 84: I would add that its the localized path: `Gets the path of the localized search root.`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/SearchResultServlet.java
```diff
@@ -16,282 +16,291 @@
...
+        Page currentPage = Optional.ofNullable(request.getResourceResolver().adaptTo(PageManager.class))
+            .map(pm -> pm.getContainingPage(request.getResource()))
+            .orElse(null);
         if (currentPage != null) {
-            Resource searchResource = getSearchContentResource(request, currentPage);
-            List<ListItem> results = getResults(request, searchResource, currentPage);
-            writeJson(results, response);
-        }
-    }
+            SlingBindings bindings = new SlingBindings();
```
jckautzmann, line 149: why are sling bindings introduced? Where are they used?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/SearchResultServlet.java
```diff
@@ -16,282 +16,291 @@
...
+        private final String searchRootPagePath;
+
+        /**
+         * Construct the default search.
+         *
+         * @param currentPage The current page.
+         * @param resourceResolver The resource resolver.
+         */
+        public DefaultSearch(@NotNull final Page currentPage, @NotNull final ResourceResolver resourceResolver) {
+            this.searchRootPagePath = Optional.ofNullable(currentPage.getContentResource())
```
jckautzmann, line 273: I'm not convinced that the search root path should be set to the language root when a search root is not defined. Instead I would set it to `null`.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v1/SearchImplTest.java
```diff
@@ -47,6 +54,21 @@ void testSearchProperties() {
         assertEquals(3, search.getSearchTermMinimumLength());
         assertEquals("/jcr:content/search", search.getRelativePath());
         assertEquals("core/wcm/components/search/v1/search", search.getExportedType());
+        assertEquals("/content/en/search", search.getSearchRootPagePath());
+        assertSame(search.getSearchRootPagePath(), search.getSearchRootPagePath());
```
jckautzmann, line 58: Why are you testing that both objects refer to the same object?

ky940819: It was just a test that it doesn't recompute the object.
Subsequent calls follow different branches, so this was mostly just crafting tests to make the branch coverage complete.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/SearchResultServlet.java
ky940819: I agree; however, please refer to the last section of the PR description. Specifically point #4.

This PR has been stale for a little but I believe the intention of this was to maintain the current behavior. of using the language root when the search root is not defined. It's a behavior that I don't like, but it already does this.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/SearchResultServlet.java
ky940819: I will double check, but I believe the sling bindings was related to fetching the Search component model.

Issue #1162, opened 2020-09-04
[Container] Provide UI test for the Container component


habansal:
## Bug Report

**Current Behavior**
There are ui-js tests for various Core components, but none exists for the Container component.
See https://github.com/adobe/aem-core-wcm-components/tree/master/testing/it/ui-js/src/content/jcr_root/apps/core/wcm/tests/test-cases

**Expected behavior/code**
There must be added a UI test for the Container component that tests that the container correctly renders it's children

**Environment**
- AEM version and patch level - NA
- Core Components version - latest version
- JRE version - NA

**Possible Solution**

**Additional context / Screenshots**
* Lack of an e2e test resulted in failure to detect a CSO for AEM-as-a-Cloud-Service. See (CSORCA-5288)
cc @gabrielwalt @pfauchere

PR #1255, opened 2020-11-12
Fixes #1187 (2nd try): try to get TemplatedResource from current resource, and fallback to current request's resource

### LINKED ISSUE #1187: [Container] Simple mode broken when using Sling Model Delegation Pattern
## Bug Report

**Current Behavior**
Scenario: A custom container component extends the container component by extending the Java logic using the Sling Model Delegation Pattern. It delegates the fetching of the container items (`getItems()`) to the original implementation. In Simple mode this logic is used to fetch the container children to render them.

This is broken in most recent AEM cloud service versions (both in the cloud and locally in AEM cloud SDK quickstart) with the result that no children are found and the paragraph is rendered empty (although there are children).

**Expected behavior/code**
The container should always render it's children, regardless of the selected render mode and AEM (cloud) version.

**Environment**
- AEM 2020.9.4245.20200921T202853Z-200827
- Core Components 2.11.1

**Possible Solution**
The reason is:
- the container component uses the [getEffectiveResource()](https://github.com/adobe/aem-core-wcm-components/blob/f8784d226e6b2fec53aafb5df712465519535688/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractContainerImpl.java#L264-L269) method to get the resource to fetch its children
- this works well if the original LayoutContainerImpl is used without "subclassing" it using the delegation pattern. in most recent AEM cloud service versions `this.resource` is already a `TemplatedResource` instance.
- however, when the delegation pattern is used the `this.resource` is not a `TemplatedResource` but some wrapper instance, and the logic adapts the *current request* to `TemplatedResource`.
- Adapting a request to `TemplatedResource` fetches it's resources and creates a templated resource from it.
- However in most recent AEM cloud versions request.getResource() already returns a templated resource, leading to create a templated resource from a templated resource. A check in the templated resource implementation detects that the source resource is already a synthetic resource (it extends it), and it skips fetching the children. This result is an empty children list and an empty container.

Probably this is a bug in AEM itself and should be fixed in the adapter factory for templated resources. However i think the code in core components is also not correct, instead of adapting the current request the current resource should be adapted to `TemplatedResource`. I will provide a PR for this.

stefanseifert:
Fixes #1187 

the first PR that was included in 2.12.0 was reverted in 2.12.2 (see [here](https://github.com/adobe/aem-core-wcm-components/issues/1187#issuecomment-724828689)) - this is a new PR which should over all use cases.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractContainerImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractContainerImpl.java
@@ -261,7 +261,9 @@ public abstract class AbstractContainerImpl extends AbstractComponentImpl implem
         if (this.resource instanceof TemplatedResource) {
             return this.resource;
         }
-        return Optional.ofNullable((Resource)this.request.adaptTo(TemplatedResource.class)).orElse(this.resource);
+        return Optional.ofNullable((Resource)this.resource.adaptTo(TemplatedResource.class))
+                .orElse(Optional.ofNullable((Resource)this.request.adaptTo(TemplatedResource.class))
+                .orElse(this.resource));
     }
 
     /*


vladbailescu:
We've been discussing this in the team and with our other colleagues. We would like to fix #1162 before we do this kind of changes, for better quality assurance.

stefanseifert:
+1

stefanseifert:
thanks for taking this into account for 2.14.0. this issue is still a big problem, and also breaks `.model.json` headless output as the items export also relies on the problematic `getEffectiveResource()` method - hopefully this is merged and released soon!

bpauli (approved):

PR #1260, opened 2020-11-13
[Form Text & Form Options] Fixes #1259: Streamline form text and form options dialog

### LINKED ISSUE #1259: [Form Text & Form Options] Change About & Element Name to be consistent
## Feature Request

No problem here, but the dialogs of the two components are structured a little bit differently.

 

For consistency reasons I think it would be better if the two dialogs were structured in the same way, i.e., _Form Options_ should have an _About_ tab (similar to _Form Text_) and _Form Text_'s _Element Name_ should simply be called _Name_ to be in line with _Form Options_.

**Are there alternatives?**

**Documentation**
Maybe a screenshot or design?

dhardtke:
| Fixed Issues?            | Fixes #1259 
| Major: Breaking Change?  | Possibly, if customers are overriding form/text/v2/cq:dialog and expected the "About" section to be located inside `<properties>`
| Documentation Provided   | No (code comments and or markdown)

- Changes "Form Text Field"'s "Element Name" label to "Name" to be in line with "Form Options Text"
- Moves "Form Option Field"'s "About" section to a dedicated tab and removes the headings "Main" and "About"

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/form/options/v2/options/_cq_dialog/.content.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/form/options/v2/options/_cq_dialog/.content.xml
@@ -47,11 +47,6 @@
                                         jcr:primaryType="nt:unstructured"
                                         sling:resourceType="granite/ui/components/coral/foundation/container">
                                         <items jcr:primaryType="nt:unstructured">
-                                            <mainHeading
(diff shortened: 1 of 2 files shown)


vladbailescu:
Needs:
- No new tab (it's not a backwards compatible change as it might clash with a similarly named tab in a customization)
- Verification

PR #1272, opened 2020-11-20
[HTML Page Items] Improve HtmlPageItemsConfig by exposing the whole config as API

### LINKED ISSUE #1234: Html Page Items context-aware configuration not editable
PR #1203 introduced support for including custom CSS, JS and metadata tags using a "html page item" concepts based on context-aware configuration.

however the annotation class `com.adobe.cq.wcm.core.components.config.HtmlPageItemsConfig` does not reflect what's editable - it contains only definition for the prefix path, but not for the items and properties that can be defined inside of it. with caconfig it's possible to describe these nested configurations as well, and only when this metadata is present it can be edited with configuration editors like https://wcm.io/caconfig/editor/

here is an example for describing a nested configuration in caconfig annotation classes:
https://github.com/wcm-io/wcm-io-caconfig/blob/develop/sample-app/bundles/core/src/main/java/io/wcm/caconfig/sample/config/ConfigSampleNested.java

please also not that the code in PageImpl is using the "low-level" ConfigrationResourceResolver which should not be done in this case as it should only be used for configuration not described via annotations and not stored in "sling:configs". when the configuration is fully described via it's annotation class the ConfigurationResolver can be used and all item data can be accessed in a type-safe way.

it would also be nice to have this feature documented and and example included in the component library. currently i do not fully get how this "path prefix" feature is supposed to work.

dhardtke:
Fixes #1234

| Fixed Issues?            | Fixes #1234
| Patch: Bug Fix?          | x
| Minor: New Feature?      | :+1: 
| Major: Breaking Change?  | :+1: 
| Tests Added + Pass?      | x
| Any Dependency Changes?  |

This PR adds two new Context-aware configurations:
- `com.adobe.cq.wcm.core.components.config.HtmlPageItemConfig`
- `com.adobe.cq.wcm.core.components.config.AttributeConfig`

They are used as child configurations in the already existing `com.adobe.cq.wcm.core.components.config.HtmlPageItemsConfig`.

This unfortunately implies a change in the config's structure as child configs are now stored under one node called `items`, compared to how they were stored as immediate children of the parent config. (See the Javadoc inside `HtmlPageItemsConfig` for the diff).

It also adds [wcm.io Editor](https://github.com/wcm-io/wcm-io-caconfig) specific properties to allow the selection of e.g. an item's element as dropdown.

TODOs:
- [ ] Fix `PageImplTest`
  - I'm not quite sure how to test this, i.e., how to provide the UUT with a (proper) instance of `HtmlPageItemsConfig` based off of `bundles/core/src/test/resources/page/v2/test-sling-configs.json`
- [ ] Clarify if the version changes of `com.adobe.cq.wcm.core.components.config` (1.0.0 -> 2.0.0) and `com.adobe.cq.wcm.core.components.models` (12.16.0 -> 13.0.0) make sense
- [ ] Clarify if the deprecations of `HtmlPageItem#PN_ELEMENT`, `HtmlPageItem#PN_LOCATION` and `HtmlPageItem#NN_ATTRIBUTES` make sense

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/config/AttributeConfig.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/config/AttributeConfig.java
@@ -13,12 +13,26 @@
  ~ See the License for the specific language governing permissions and
  ~ limitations under the License.
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
-/**
- * <p>
- *      This package defines the context-aware configurations used by the WCM Core Components
- * </p>
- */
-@Version("2.0.0")
 package com.adobe.cq.wcm.core.components.config;
 
-import org.osgi.annotation.versioning.Version;
+import org.apache.sling.caconfig.annotation.Property;
+
+public @interface AttributeConfig {
+    /**
+     * Returns the name of the attribute.
+     *
+     * @return The name of the attribute - if left empty the attribute won't be rendered out
+     * @since com.adobe.cq.wcm.core.components.config 2.0.0
+     */
+    @Property(label = "Name")
+    String name() default "";
+
+    /**
+     * Returns the value of the attribute.
+     *
+     * @return The value of the attribute
+     * @since com.adobe.cq.wcm.core.components.config 2.0.0
+     */
+    @Property(label = "Value")
+    String value() default "";
+}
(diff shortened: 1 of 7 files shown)


vladbailescu:
@dhardtke , thanks for taking the time to do this. It's probably what we should have done in the first place :(

Because new versions of Core Components are getting rolled out automatically to AEMaaCS customers, will need to move forward in a way that's backwards compatible (at least for a while) with the old way. That means:
* no major version bump for the packages
* falling back to the *legacy* structure if `item` nodes are located directly in the config path

As for the property and node names, I think we should expose them (as part of corresponding config interface) because they are effectively API (and some use cases might want to be able to reference them)

vladbailescu:
Needs:
- Fixing merge conflicts
- Adding backwards-compatible support for older structure
- Verification

jckautzmann (approved):

jckautzmann:
@dhardtke thx for your contribution. I created a new PR based on yours and integrating @vladbailescu 's feedback at:
https://github.com/adobe/aem-core-wcm-components/pull/1373

PR #1275, opened 2020-11-23
Unhide component group for cloud/AEMaaCS artifacts

### LINKED ISSUE #1274: Unhide component group for cloud/AEMaaCS artifacts
For AEM as a Cloud Service we are exploring scenarios where a simpler website could be created with no backend development. For that we would like to have the Core Components in visible groups: `Core Structure`, `Core Content` and `Core Form`.

Our guidance for developers using the Core Components in their custom projects will remain unchanged; use proxy components to:
- lock in the version of the Core Components
- customise the rendering or model
- avoid needing a content migration when bumping the version of a component

vladbailescu:
Fixes #1274

### DIFF
--- a/content/pom.xml
+++ b/content/pom.xml
@@ -273,8 +273,9 @@
                                         <replace file="target/jcr_root/libs/core/wcm/components/text/v2/text/.content.xml" token=".core-wcm" value="Core Content"/>
                                         <replace file="target/jcr_root/libs/core/wcm/components/title/v2/title/.content.xml" token=".core-wcm" value="Core Content"/>
 
+                                        <replace file="target/jcr_root/libs/core/wcm/components/form/container/v2/container/.content.xml" token=".core-wcm-form" value="Core Content"/>
+
                                         <replace file="target/jcr_root/libs/core/wcm/components/form/button/v2/button/.content.xml" token=".core-wcm-form" value="Core Form"/>
-                                        <replace file="target/jcr_root/libs/core/wcm/components/form/container/v2/container/.content.xml" token=".core-wcm-form" value="Core Form"/>
                                         <replace file="target/jcr_root/libs/core/wcm/components/form/hidden/v2/hidden/.content.xml" token=".core-wcm-form" value="Core Form"/>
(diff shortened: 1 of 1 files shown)


msagolj (approved):

content/pom.xml
```diff
@@ -241,6 +241,52 @@
...
                         </executions>
                     </plugin>
+                    <plugin>
+                        <artifactId>maven-antrun-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <phase>prepare-package</phase>
+                                <configuration>
+                                    <tasks>
+                                        <replace file="target/jcr_root/libs/core/wcm/components/breadcrumb/v2/breadcrumb/.content.xml" token=".core-wcm" value="Core Structure"/>
```
kwin, line 251: A bit late to the party but wouldn't it make sense to define Maven properties for the 3 group names? Also using Maven filtering instead of falling back to using Ant feels more natural in Maven, that would also allow to not needing to extend this list every time a new component is added. That way the default property value could be the On Prem/AMS component name, while for cloud you override the property values in a dedicated Maven Profile.
@vladbailescu WDYT?

wimsymons: I agree. The way it is implemented right now, feels like a nasty hack. Maven filtering is so much nicer.

wimsymons: @kwin @vladbailescu should I create a new ticket to track this?

vladbailescu: @wimsymons, @kwin , we already tried that approach in #1216 but we weren't able to get it work the way we want: we would like to be able to generate both regular (with hidden component groups) and `cloud` (with visible component groups) artifacts in the same build, if the `cloud` profile is active (it helps a lot in the release process). If you think of a better way to do it than the way I currently set it up, we'd be grateful to get a PR for your idea and test it out. You can reference #1274 in the PR.

kwin: @vladbailescu Actually that should be possible with the parameter `filterFiles` which allows you to reference two different java properties files for cloud and classic. That way replacement should work the way you intend.

PR #1305, opened 2020-12-08
Add runtime statistics for the AdaptiveImageServlet 

### LINKED ISSUE #1292: AdaptiveImageServlet should provide runtime statistics
## Feature Request

When using the AdaptiveImageServlet it is it's completely unclear what renditions are created, how often these operations are executed, and how much time it typically takes. As these operations can be expensive (in terms of CPU and heap) it should be possible to get an overview about this. It would be best if it can be provided as metrics, so it can be easily added to monitoring.

The AdaptiveImageServlet should create metrics for 

* the number of invocations for each dimension (the total number of different dimensions typically requested should be < 10)
* the rendering time of the individual requests
* as part of the overall rendering the fetching of the underlying binary from the blobstore should be measured separately.

**Are there alternatives?**
if these metrics are not available, the only chance is to extract the numbers from the access.log by applying regexpes to the path and searching for the characteristic patterns of the invocations of this servlet. And that's unlikely to work reliably :-(

**Documentation**
This feature is not directly visible to end-users but only to operators.

joerghoh:
| Fixed Issues?            | Fixes #1292
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Documentation Provided   | Inline
| Any Dependency Changes?  |

The AdaptiveImageServlet now maintains statistics as Sling Metrics, which you can check via /system/console/slingmetrics.

| Name | Description
|--------|--------
|sling:com.adobe.cq.wcm.core.components.internal.servlets.AdaptiveImageServlet:invocations| Number of total invocations of the AdaptiveImageSerlvet (Counter)
|sling:com.adobe.cq.wcm.core.components.internal.servlets.AdaptiveImageServlet:rendition-rendered | counts how often a rendition was rendered (in contrast to sending a 304 or 404)
|sling:com.adobe.cq.wcm.core.components.internal.servlets.AdaptiveImageServlet:original-rendition-used | counts how often the original rendition was used for rendering instead of any web rendition
| sling:com.adobe.cq.wcm.core.components.internal.servlets.AdaptiveImageServlet:base-rendition-rejected-because-of-size | counts how often the rendering was refused because the rendition to be used was larger than the configured maximum value
|sling:com.adobe.cq.wcm.core.components.internal.servlets.AdaptiveImageServlet:request-duration (timer)| a timer which collects the duration of the invocations of the AdaptiveImageServlet)

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -506,7 +506,7 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
         }
         // If no rendition was found, attempt to use original
         if (bestRendition == null) {
-            bestRendition = getOriginal(asset);
+            return getOriginal(asset);
         }
         return filter(bestRendition);
     }
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -46,6 +46,7 @@ import org.apache.sling.api.resource.ResourceUtil;
 import org.apache.sling.api.resource.ValueMap;
 import org.apache.sling.api.servlets.HttpConstants;
 import org.apache.sling.api.servlets.SlingSafeMethodsServlet;
+import org.apache.sling.commons.metrics.Timer;
 import org.apache.sling.commons.mime.MimeTypeService;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -102,6 +103,8 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
     private static final String SELECTOR_WIDTH_KEY = "width";
     private int defaultResizeWidth;
     private int maxInputWidth;
+    
+    private AdaptiveImageServletMetrics metrics;
 
     @SuppressFBWarnings(justification = "This field needs to be transient")
     private transient MimeTypeService mimeTypeService;
@@ -109,16 +112,20 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
     @SuppressFBWarnings(justification = "This field needs to be transient")
     private transient AssetStore assetStore;
 
-    public AdaptiveImageServlet(MimeTypeService mimeTypeService, AssetStore assetStore, int defaultResizeWidth, int maxInputWidth) {
+    public AdaptiveImageServlet(MimeTypeService mimeTypeService, AssetStore assetStore, AdaptiveImageServletMetrics metrics, 
+            int defaultResizeWidth, int maxInputWidth) {
         this.mimeTypeService = mimeTypeService;
         this.assetStore = assetStore;
+        this.metrics = metrics;
         this.defaultResizeWidth = defaultResizeWidth > 0 ? defaultResizeWidth : DEFAULT_RESIZE_WIDTH;
         this.maxInputWidth = maxInputWidth > 0 ? maxInputWidth : DEFAULT_MAX_SIZE;
     }
 
     @Override
     protected void doGet(@NotNull SlingHttpServletRequest request, @NotNull SlingHttpServletResponse response) throws IOException {
+        Timer.Context requestDuration = metrics.startDurationRecording();
         try {
+            metrics.markServletInvocation();
             RequestPathInfo requestPathInfo = request.getRequestPathInfo();
             List<String> selectorList = selectorToList(requestPathInfo.getSelectorString());
             String suffix = requestPathInfo.getSuffix();
@@ -231,10 +238,13 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
                     transformAndStreamAsset(response, componentProperties, resizeWidth, quality, asset, imageType,
                             imageName);
                 }
+                metrics.markRenditionRendered();
             }
         } catch (IllegalArgumentException e) {
             LOGGER.error("Invalid image request", e.getMessage());
             response.sendError(HttpServletResponse.SC_NOT_FOUND);
+        } finally {
+            requestDuration.stop();
         }
 
     }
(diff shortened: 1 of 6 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -496,7 +506,7 @@ private EnhancedRendition getBestRendition(@NotNull Asset asset, int width) thro
         }
         // If no rendition was found, attempt to use original
         if (bestRendition == null) {
-            bestRendition = new EnhancedRendition(asset.getOriginal());
+            bestRendition = getOriginal(asset);
```
vladbailescu, line 509: This will cause the `original` rendition to be filtered twice: once in `getOriginal` and another time down bellow.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServletMetricsTest.java
```diff
@@ -0,0 +1,72 @@
...
+import io.wcm.testing.mock.aem.junit5.AemContextExtension;
+
+@ExtendWith(AemContextExtension.class)
+public class AdaptiveImageServletMetricsTest {
+    
+    
+    
+    /**
+     * Oh really, you want me to write unit tests for this class, which 
+     * actually does not have ANY logic?
```
vladbailescu, line 40: Then don't create a class for this and inline all the counters and timers in AIS?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -231,10 +238,13 @@ protected void doGet(@NotNull SlingHttpServletRequest request, @NotNull SlingHtt
                     transformAndStreamAsset(response, componentProperties, resizeWidth, quality, asset, imageType,
                             imageName);
                 }
+                metrics.markRenditionRendered();
```
vladbailescu, line 241: This metric invocation is inconsistent with `markOriginalRenditionUsed`

vladbailescu (changes requested):
@joerghoh , thanks for submitting this, it's really something that we need do be adding to other places too.

However, I have a few things I'm not very happy with:
- we're inconsistent with the things we measure, eg usages of original vs other rendition
- there's no tracking of known errors (other refusals) or unknown ones

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
joerghoh: good finding, fixed it.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
joerghoh: How is it inconsistent?

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServletMetricsTest.java
joerghoh: AIS is not an OSGI service, so passing these arguments via the constructor is not what I wanted to do.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
vladbailescu: Perhaps I don't understand exactly what you mean by this. To me, the name and description suggest this will count how many times one of the asset's renditions are used.

What it does, in fact, is to count how many successful images are served by the AIS.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -511,7 +521,9 @@ private EnhancedRendition getBestRendition(@NotNull Asset asset, int width) thro
     @NotNull
     private EnhancedRendition getOriginal(@NotNull Asset asset) throws IOException {
         EnhancedRendition original = new EnhancedRendition(asset.getOriginal());
-        return filter(original);
+        EnhancedRendition filtered = filter(original);
+        metrics.markOriginalRenditionUsed();
```
vladbailescu, line 525: `getOriginal` is only triggered when the original renditions is requested explicitly. It could also be returned by `getBestRendition` implicitely (in the `for.. if` loop).

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServletMetricsTest.java
vladbailescu: It was implemented like that to allow mapping of multiple AIS instances to different configurations. You could still pass the `MetricsService` to the constructor and define the counters and timer as members of AIS. Instead of calling `markServletInvocations` you would call `invocations.increment()`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
joerghoh: This counter counts how often a rendition was actually created. Another outcome (instead of the various reasons to return a 404) is the case of returning a 304 "Not modified" [1]. The basic idea behind this counter is actually to determine how often a rendition was created vs how often a 304 was sent (which can be determined as difference between the RenditionRendered and the invocations counter).

[1] https://github.com/adobe/aem-core-wcm-components/blob/master/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L219

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServletMetricsTest.java
joerghoh: In that case I would have counters for each instance of AIS, right now I have consolidated counters for all of them. Which is fine for me right now.

vladbailescu:
Needs:
- Making sure all execution paths are treated/accounted for.
- Verification

PR #1311, opened 2020-12-10
[Image] AdaptiveImageServlet renders dark background when converting transparent PNG images to JPG

### LINKED ISSUE #1312: [Image] Render light background when converting transparent PNG images to JPG
Currently, when a transparent image is converted to a format that does not support transparency, the output has a dark background. It's more common in other tools to generate a light background.

vladbailescu:
Fixes #1312


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -347,6 +347,12 @@ private void transformAndStreamAsset(SlingHttpServletResponse response, ValueMap
                     int resizeHeight = calculateResizeHeight(originalWidth, originalHeight, resizeWidth);
                     if (resizeHeight > 0 && resizeHeight != originalHeight) {
                         layer = getLayer(rendition);
+                        if (layer.getBackground().getTransparency() != Transparency.OPAQUE &&
+                                ("jpg".equalsIgnoreCase(extension) || "jpeg".equalsIgnoreCase(extension))) {
+                            LOGGER.debug("Adding default (white) background to a transparent PNG: {}/{}", asset.getPath(),
+                                    rendition.getName());
+                            layer.setBackground(Color.white);
```
actinium15, line 354: I wonder if the background color should be "Dark Mode aware" and prefer black bg over white when dark mode is enabled.

vladbailescu: We considered that but it's not feasible as the images will get cached at dispatcher level. We could introduce a selector but that's already too much overhead and we'd rather invest more in something like #496. Then you could have a dark/light rendition pair.

For now, if you need to switch between dark and light themes, best bet is to use an image format that supports transparency.

jckautzmann (approved):

PR #1313, opened 2020-12-10
Update BUILDING.md

### LINKED ISSUE #1290: Clarify Building section in Readme
Currently the [readme](https://github.com/adobe/aem-core-wcm-components/blob/master/README.md) just states 

> To compile your own version of the Core Components, you can build and install everything on your running AEM instance by issuing the following command in the top level folder of the project:
`mvn clean install -PautoInstallSinglePackage`

This is not enough for local AEM instances based on a newer quickstart from https://experience.adobe.com/#/downloads/content/software-distribution/en/aemcloud.html.

The reason is twofold:
1. The AEM quickstart already comes with Core WCM Components preinstalled
1. The OSGi installer is configured in a way that it allows multiple version of the same BSN (https://sling.apache.org/documentation/bundles/osgi-installer.html#multi-version-support)

As the original Core WCM version cannot be uninstalled easily (as it is installed from location `launchpad:resources/install/20/core.wcm.components.core-2.11.1.jar`) there is no easy way to force the scripts to load the classes from the newer bundle (https://issues.apache.org/jira/browse/SLING-9210).

vladbailescu:
Fixes #1290

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -24,6 +24,10 @@ The hostname and port of the instance can be changed with the following user def
 * `aem.host` and `aem.port` for the author instance.
 * `aem.publish.host` and `aem.publish.port` for the publish instance.
 
+### AEM as a Cloud Service SDK
+
+When compiling and deploying to AEM as a Cloud Service SDK, you can use the `cloud` profile (in conjunction with previously documented profiles) to generate `cloud`-ready artifacts (with components located in `/libs` instead of `/apps`). To allow recompilation of the HTL scripts, you should disable `aem-precompiled-scripts` bundle.
+
 ## NPM modules
 
 Modules that export Javscript and CSS are usually also configured as NPM modules.
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -26,7 +26,12 @@ The hostname and port of the instance can be changed with the following user def
 
 ### AEM as a Cloud Service SDK
 
-When compiling and deploying to AEM as a Cloud Service SDK, you can use the `cloud` profile (in conjunction with previously documented profiles) to generate `cloud`-ready artifacts (with components located in `/libs` instead of `/apps`). To allow recompilation of the HTL scripts, you should disable `aem-precompiled-scripts` bundle.
+When compiling and deploying to AEM as a Cloud Service SDK, you can use the `cloud` profile (in conjunction with 
+previously documented profiles) to generate `cloud`-ready artifacts (with components located in `/libs` instead 
+of `/apps`). To allow recompilation of the HTL scripts, you should disable `aem-precompiled-scripts` bundle.
+
+Due to [FELIX-6365](https://issues.apache.org/jira/browse/FELIX-6365), please only use `autoInstallPackage` and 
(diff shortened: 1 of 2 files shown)


BUILDING.md
```diff
@@ -24,6 +24,10 @@ The hostname and port of the instance can be changed with the following user def
 * `aem.host` and `aem.port` for the author instance.
 * `aem.publish.host` and `aem.publish.port` for the publish instance.
 
+### AEM as a Cloud Service SDK
+
+When compiling and deploying to AEM as a Cloud Service SDK, you can use the `cloud` profile (in conjunction with previously documented profiles) to generate `cloud`-ready artifacts (with components located in `/libs` instead of `/apps`). To allow recompilation of the HTL scripts, you should disable `aem-precompiled-scripts` bundle.
+
```
kwin, line 30: Probably worth to mention that any other Core WCM bundle versions (the release ones) should be disabled due to https://issues.apache.org/jira/browse/FELIX-6365. Also the new profile `install-bundle` should be mentioned

parent/pom.xml
```diff
@@ -183,6 +183,7 @@
                     <version>2.1.0</version>
                     <executions>
                         <execution>
+                            <id>install-bundle</id>
```
kwin, line 186: shouldn't it rather be named `autoInstallBundle` to match the naming pattern of `autoInstallSinglePackage`?
Also it should be clarified how those two should be used together.

vladbailescu: That's not a profile name, it's just an execution id.

kwin: Sorry, makes sense then!

BUILDING.md
vladbailescu: I did not come across this issue when working with the latest snapshots:
image

kwin: I am talking about the same BSN in two versions. Your screenshot shows two BSNs in two versions!
How did you deploy the SNAPSHOT version bundle then?

vladbailescu:
```
[INFO] --- maven-sling-plugin:2.1.0:install (default-cli) @ core.wcm.components.core ---
[INFO] Installing Bundle com.adobe.cq.core.wcm.components.core(/Users/vlad/dev/aem/cq/modules/sample/aem-core-wcm-components/bundles/core/target/core.wcm.components.core-2.13.1-SNAPSHOT.jar) to http://localhost:4502/system/console/bundles via POST
```
That was deployed on top of `2.13.0`. The amp bundle was not deployed, it's there just as an example.

kwin: Ok, the order of deployment is relevant: 

1. If you deploy the first time via `autoInstallSinglePackage` the deployment of the SNAPSHOT bundle goes via the OSGi installer, which will deploy the bundle in parallel(!)
2. If you deploy the bundle again via `autoInstallPackage` (which goes via the Felix Web Console ReST API) you trigger the bug

vladbailescu: I see what you mean:

vladbailescu: Updated docs in e330129

kwin: Thanks, LGTM now

PR #1251, opened 2020-11-09
PWA support for page component

kbalasub:
This feature adds the ability to enhance any AEM site to a progressive web app (PWA) through configuration and not code. The page component now includes a model that checks for PWA properties from the sites project and references a manifest and service worker into the page making the site installable.

PR #1319, opened 2020-12-14
PWA support for page component (#1251)

vladbailescu:
PWA for sites.
Fixes issue: https://github.com/adobe/aem-core-wcm-components/issues/1251


kbalasub (approved):

ky940819:
Is there a reason for the hard-coded depth of the PWA root?
Would it make sense to have these configurations as part of a CA Config, or a policy on the template?

kbalasub:
The identification of a resource as a sites project has been an issue because it is a standard page without any additional attribute such as a sling resource type or other attribute to indicate a sites project. Also this solution has to work for already existing sites and not just newly created ones. In our discussions we observed no alternative to the current implementation. This may change in the future to support localized sites which is currently under discussion.

ky940819:
What about if you just climbed the tree until you find a page with PN_ENABLE_PWA = true, and then use that page as the root.

kbalasub:
That is a possiblility and duly noted and will be discussed. The challenge is that all sites projects do not need to have PWA to be enabled. Existing sites projects will not have that attribute and there is a potential performance impact if we need to read every resource on the way up the tree. We will however discuss this feedback as we evolve the solution.

jckautzmann:
@kbalasub Is there an issue in the Core Components repo explaining what we'd like to achieve with this PR? If not, could you please create one with all the details?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/PWA.java
```diff
@@ -0,0 +1,95 @@
...
+ * @since com.adobe.cq.wcm.core.components.models 12.18.0
+ */
+public interface PWA {
+
+    /**
+     * Name of the property that defines it PWA is enabled for the website.
+     *
+     * @since com.adobe.cq.wcm.core.components.models 12.18.0
+     */
+    String PN_ENABLE_PWA = "enablePWA";
```
jckautzmann, line 31: I'd rename this property and this field as: 
String PN_PWA_ENABLED = "pwaEnabled";

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/PWA.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Name of the property that defines the theme color used.
+     *
+     * @since com.adobe.cq.wcm.core.components.models 12.18.0
+     */
+    String PN_THEME_COLOR = "themecolor";
```
jckautzmann, line 38: Please use camelCase notation for the property name: `themeColor`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/PWA.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Name of the property that defines the PWA icon
+     *
+    @since com.adobe.cq.wcm.core.components.models 12.18.0
+    */
+    String PN_PWA_ICON = "pwaicon";
```
jckautzmann, line 45: Please use camelCase notation for the property name: `pwaIcon `

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/PWA.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Interface for reading progressive web apps configuration
+ *
+ * @since com.adobe.cq.wcm.core.components.models 12.18.0
+ */
+public interface PWA {
+
+    /**
+     * Name of the property that defines it PWA is enabled for the website.
```
jckautzmann, line 27: typo: it PWA -> if PWA

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/PWA.java
```diff
@@ -0,0 +1,95 @@
...
+    */
+    String PN_PWA_ICON = "pwaicon";
+
+    /**
+     * Returns {@code true} if PWA features are enabled, {@code false} otherwise
+     *
+     * @return whether PWA is enabled or not
+     * @since com.adobe.cq.wcm.core.components.models 12.18.0
+     */
+    default boolean isPWAEnabled() { throw new UnsupportedOperationException("Not Implemented"); };
```
jckautzmann, line 53: I'd rename it `isEnabled()` to be in sync with other methods naming

vladbailescu:
Needs:
- Requirements added to an issue
- All suggestions should be implemented

gabrielwalt:
> * Requirements added to an issue. TODO for @gabrielwalt
Done: https://github.com/adobe/aem-core-wcm-components/issues/1251

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PWAImpl.java
```diff
@@ -0,0 +1,151 @@
...
+
+import com.adobe.cq.wcm.core.components.models.PWA;
+import com.day.cq.commons.jcr.JcrConstants;
+
+@Model(adaptables = Resource.class,
+    adapters = {PWA.class})
+public class PWAImpl implements PWA {
+
+    static final String MANIFEST_NAME = "manifest.webmanifest";
+    static final int SITES_PROJECT_LEVEL = 3;
```
bpauli, line 35: I agree with @ky940819, we shouldn't hardcode the project level as this might be different depending on how the project is structured (language structure etc.) at customer level.

bpauli (approved):

examples/ui.content/src/content/jcr_root/conf/core-components-examples/_sling_configs/.content.xml
```diff
@@ -8,4 +8,7 @@
         jcr:primaryType="nt:unstructured"
         clientId="28b07c6be4d544f188dc2f36e33b196b"
         reportSuiteId=""/>
+    <com.adobe.cq.wcm.core.components.internal.services.page.PageCaConfig
```
jckautzmann, line 11: shouldn't we expose this config as public API?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PWAImpl.java
```diff
@@ -16,56 +16,79 @@
...
+            if (contentResource != null) {
+                ValueMap valueMap = contentResource.adaptTo(ValueMap.class);
+                if (valueMap != null) {
+                    isEnabled = valueMap.get(PN_ENABLE_PWA, isEnabled);
+                    if (isEnabled) {
+                        themeColor = colorToHex(valueMap.get(PN_THEME_COLOR, StringUtils.EMPTY));
+                        iconPath = valueMap.get(PN_PWA_ICON, StringUtils.EMPTY);
+                        projectName = pwaSiteRootPage.getName();
+                        manifestPath = pwaSiteRootPage.getPath() + "/" + MANIFEST_NAME;
+                        serviceWorkerPath = "/" + projectName + "sw.js";
```
jckautzmann, line 66: please test for `projectName != null`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/IsSiteProjectRenderConditionServlet.java
```diff
@@ -0,0 +1,76 @@
...
+
+@Component(
+        service = {Servlet.class},
+        property = {
+                "sling.servlet.resourceTypes=" + IsSiteProjectRenderConditionServlet.RESOURCE_TYPE,
+                "sling.servlet.methods=GET"
+        }
+)
+/*
+ * a condition to determine if the resource at the given path is a sites project
```
jckautzmann, line 47: what do you mean with sites project?
Is it checking whether the resource at the given path is part of a site? or is the site root?

PR #1322, opened 2020-12-15
[Teaser][Download] - the description area of the dialog is not editab…

### LINKED ISSUE #1295: [Teaser][Download] - the description area of the dialog is not editable on a Cloud instance
## Bug Report

**Current Behavior**
When opening the dialog of the Teaser and the Download components, the description area remains greyed out on various cloud environments.

This issue does not happen on local instances.

jckautzmann:
…le on a Cloud instance

- instantiate the fields only when both the dialog and the description RTE have loaded

| Fixed Issues?            | Fixes #1295 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/download/v1/download/clientlibs/editor/js/download.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/download/v1/download/clientlibs/editor/js/download.js
@@ -35,36 +35,50 @@
         var dialogContent = $dialogContent.length > 0 ? $dialogContent[0] : undefined;
 
         if (dialogContent) {
-            titleTuple = new CheckboxTextfieldTuple(dialogContent, titleCheckboxSelector, titleTextfieldSelector);
-            descriptionTuple = new CheckboxTextfieldTuple(dialogContent, descriptionCheckboxSelector, descriptionTextfieldSelector, true);
-            $cqFileUpload     = $dialog.find(".cq-FileUpload");
-            $cqFileUploadEdit = $dialog.find(".cq-FileUpload-edit");
-            if ($cqFileUpload) {
-                $cqFileUpload.on("assetselected", function(e) {
-                    fileReference = e.path;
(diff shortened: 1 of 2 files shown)


bartoszglow:
Does the provided solution assume that description textfield will not be initialized in case RTE was slow unless you start to edit it?

content/src/content/jcr_root/apps/core/wcm/components/download/v1/download/clientlibs/editor/js/download.js
```diff
@@ -35,37 +35,51 @@
...
-                    titleTuple.reset();
-                    descriptionTuple.reset();
+
+            var rteInstance = $(descriptionTextfieldSelector).data("rteinstance");
+            // wait for the description textfield rich text editor to signal start before initializing.
+            if (rteInstance && rteInstance.isActive) {
+                init($dialog, dialogContent);
+            } else {
+                $(descriptionTextfieldSelector).on("editing-start", function() {
+                    init($dialog, dialogContent);
```
vladbailescu, line 45: Won't this lead to multiple initialisations? If so, we should keep track if the initialisation already happened to avoid doing it multiple times.

jckautzmann: I tested it on localhost and cloud and can confirm that it only initializes once.

jckautzmann:
> Does the provided solution assume that description textfield will not be initialized in case RTE was slow unless you start to edit it?

@bartoszglow no, the fix makes sure that the fields are initialized once both the dialog and the description textfield RTE have loaded. the `editing-start` event is triggered when the RTE is ready, when editing can start.

bartoszglow (approved):

vladbailescu (approved):

PR #1331, opened 2020-12-24
Avoid swapping image back on scroll event

### LINKED ISSUE #1330: AEM Core Component Lazy Loading breaks replace asset in Target
Issue referenced in https://jira.corp.adobe.com/browse/CQ-4312621

**Current Behavior**
Images cannot be replaced in Target because of same AEM image components lazy load feature

**Expected behavior/code**
Customer can use lazy load while still being able to replace assets in their Target experiences.

**Environment**
- AEM 6.4 (AMS)+ Target
- Core Components version 
- JRE version (e.g. `Java(TM) SE Runtime Environment 

**Possible Solution**
Provide an alternate way to possibly lazy load image
Screenshots are included in the JIRA referenced above

andrewmkhoury:
Fixed image lazy loading loadImage so it doesn't replace dynamically swapped images on scroll event.  When custom javascript does an intentional swap on the img src attribute of a lazyloaded image, it swaps back to the original when the user scrolls the browser window.  Instead it should be compared to the placeholder.

| Fixed Issues?            | Fixes #1330
| Minor: New Feature?      |
| Major: Breaking Change?  | 
| Tests Added + Pass?      | 
| Any Dependency Changes?  |

Instead of comparing to URL of src, compare against EMPTY_PIXEL placeholder:
Before:
```
            if (that._elements.image.getAttribute("src") !== url) {
```
After:
```
            if (that._elements.image.getAttribute("src") == EMPTY_PIXEL) {
```

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
@@ -232,7 +232,7 @@
             }
             var url = that._properties.src.replace(SRC_URI_TEMPLATE_WIDTH_VAR, replacement);
 
-            if (that._elements.image.getAttribute("src") !== url) {
+            if (that._elements.image.getAttribute("src") == EMPTY_PIXEL) {
                 that._elements.image.setAttribute("src", url);
                 if (!hasWidths) {
                     window.removeEventListener("scroll", that.update);
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
@@ -232,7 +232,7 @@
             }
             var url = that._properties.src.replace(SRC_URI_TEMPLATE_WIDTH_VAR, replacement);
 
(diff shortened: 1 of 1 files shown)


content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/clientlibs/site/js/image.js
```diff
@@ -232,7 +232,7 @@
             }
             var url = that._properties.src.replace(SRC_URI_TEMPLATE_WIDTH_VAR, replacement);
 
-            if (that._elements.image.getAttribute("src") !== url) {
+            if (that._elements.image.getAttribute("src") === EMPTY_PIXEL) {
```
habansal, line 235: We should also add a null check here, i guess, for the first load.
If you just load any AEM Page, containing an image component, in Authoring mode, you will notice that the image is not loaded without the null check.

```suggestion
            var imgSrcAttribute = that._elements.image.getAttribute("src");
            if (imgSrcAttribute === null || imgSrcAttribute === EMPTY_PIXEL) {
```

habansal (commented):
Added a suggestion-comment on the PR.

Change, in general, looks fine to me.
However, the Core components owners will be the right people to provide a thorough and final review.

bpauli:
Changes LGTM

bpauli (approved):

PR #722, opened 2019-09-05
[Search] Results are not limited to specified directory #835

coorghari:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            |  #835
| Patch: Bug Fix?          | 👍 
| Minor: New Feature?      | 👍 
| Major: Breaking Change?  | 👍 
| Tests Added + Pass?      | 
| Documentation Provided   | Yes (code comments and or markdown)
| Any Dependency Changes?  | No
| License                  | Apache License, Version 2.0

Extended the core Search component and authored the Search component in XF (Experience Fragment) configured the search component.
ex : 
Search Root * :  "/content/project/us/en/abc" 
Search component searches from "/Content/project/*"

Experience Fragment with Search component included in the page, Search results are different from configured "Search Root *".

Changes I have made in [0] finding the "fragmentPath" and passing the "relative content resource 
ex: [1]" to get the Search resource.

[0] : src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/SearchResultServlet.java
 [1] : jcr:content/root/search

Attached Images:

search-in-XF
Search-in-PAGE

PR #1348, opened 2021-01-19
[Search] Results are not limited to specified directory #835 (#722)

### LINKED ISSUE #835: [Search] Results are not limited to specified directory
When including a search component through an Experience Fragment, the results are not limited to the directory specified in the configuration.

vladbailescu:
* Finding Resource from fragments.
* Fixing the alignment issues.

Fixes #835


vladbailescu:
@ky940819 , @jckautzmann , I believe the test failures on this PR (which is a redo of #722) expose an issue with the refactored localization logic. The search component in the XF is configured with `"searchRoot" : "/content/en/search/xf-page/searched-tree"` and this is properly resolved up until https://github.com/adobe/aem-core-wcm-components/blob/issue/835/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/SearchImpl.java#L150, where the returned value is the current page (`/content/en/search/xf-page`), although the search root exists and is in the right (`en`) language tree.

This leads to:
```
Error:  Failures: 
Error:    SearchResultServletTest.testXFSearch:252->validateResponse:287 expected: <null/content/en/search/xf-page/searched-tree/found-01.html> but was: <null/content/en/search/xf-page/searched-tree.html>
Error:    SearchResultServletTest.testXFTemplateSearch:269->validateResponse:281 expected: <2> but was: <0>
```

PR #1349, opened 2021-01-19
[Form Text & Form Options] Fixes #1259: Streamline form text

### LINKED ISSUE #1259: [Form Text & Form Options] Change About & Element Name to be consistent

bartoszglow:
Follow up for #1260 

Changes applied to original PR:
* removed about section
* added main and about headings back

All changes that updated PR provides:
* Changes "Form Text Field"'s "Element Name" label to "Name" to be in line with "Form Options Text"


bartoszglow:
Ticket verified. Ready to be merged into development.

msagolj (approved):

PR #1350, opened 2021-01-19
Feature/expose wcmmode in datalayer (fixes #1171)

### LINKED ISSUE #1171: Expose WCMMode and run mode in dataLayer's page object

bartoszglow:
Follow up to: #1248 

Resolved merge conflicts.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageImpl.java
@@ -40,9 +40,11 @@ import org.apache.sling.api.resource.ValueMap;
 import org.apache.sling.models.annotations.Exporter;
 import org.apache.sling.models.annotations.Model;
 import org.apache.sling.models.annotations.injectorspecific.InjectionStrategy;
+import org.apache.sling.models.annotations.injectorspecific.OSGiService;
 import org.apache.sling.models.annotations.injectorspecific.ScriptVariable;
 import org.apache.sling.models.annotations.injectorspecific.Self;
 import org.apache.sling.models.factory.ModelFactory;
+import org.apache.sling.settings.SlingSettingsService;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -55,6 +57,7 @@ import com.adobe.cq.wcm.core.components.models.Page;
 import com.day.cq.tagging.Tag;
 import com.day.cq.wcm.api.NameConstants;
 import com.day.cq.wcm.api.Template;
+import com.day.cq.wcm.api.WCMMode;
 import com.day.cq.wcm.api.designer.Design;
 import com.day.cq.wcm.api.designer.Designer;
 import com.day.cq.wcm.api.designer.Style;
@@ -91,6 +94,9 @@ public class PageImpl extends AbstractComponentImpl implements Page {
 
     @Inject
     private SlingModelFilter slingModelFilter;
+
+    @OSGiService
+    private SlingSettingsService slingSettings;
 
     @Self
     private SlingHttpServletRequest request;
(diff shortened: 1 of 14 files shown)


bartoszglow:
Verification in progress.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageImpl.java
```diff
@@ -313,14 +319,27 @@ protected void addPolicyClientLibs(List<String> categories) {
     @Override
     @NotNull
     protected final PageData getComponentData() {
+        boolean isAuthor = slingSettings.getRunModes().contains("author");
```
kwin, line 322: How do we deal with this being deprecated (https://github.com/adobe/aem-project-archetype/issues/528#issuecomment-725617232)?

vladbailescu: The guidance at the moment is to use environment/runmode-specific OSGi configuration values (https://experienceleague.adobe.com/docs/experience-manager-cloud-service/implementing/deploying/configuring-osgi.html?lang=en#cloud-manager-api-format-for-setting-properties) to distinguish between different node and environment types.

So I guess we will need to adjust requirements so that in DataLayer we expose:
- node type: author, publish, (other might be available at some point)
- environment type: dev, stage, prod

kwin: The original requirement was about WCM Mode (#1171). Exposing node type and environment type in addition is nice, but does not replace the proposed data layer field `mode`!

kwin: @vladbailescu Any thing I can do to speed this up?

kwin: My use case was really about the edit mode/preview mode. With wcmmode=disabled it should behave the same on author and publish!

kwin: I see the issue has been moved to icebox which feels a bit rude as I provided this PR quite some time ago. What exactly is the concern here? Using the run mode API or anything else?

vladbailescu: @kwin , I'm sorry if that came across as rude, I'm pretty sure that wasn't @gabrielwalt 's intention. Both him and I highly appreciate your contributions to this project, and the AEM Community in general.

There's been a lot of discussion internally as well and the general consensus between our senior developers/ architects is that server-side rendering should be the same regardless of run mode and editing/WCM mode. Hence the SlingSettingService is deprecated in Cloud Service (https://javadoc.io/doc/com.adobe.aem/aem-sdk-api/latest/org/apache/sling/settings/SlingSettingsService.html).

Our guidance at the moment is: projects that NEED to have this information should use a custom OSGi configuration that allows them to differentiate between the node and env types (author, publish & preview).

I know we have some bad examples regarding WCM mode in the Core Components themselves, and we should probably clean them up, but we definitely don't want to add more of these.

I will go ahead and close the PR un-merged, hope you understand it's not out-of-spite or to offend you (some more) but we just don't want to open this potential can of worms in the project.

kwin: This request is about wcmmode, not run mode! That API hasn’t been deprecated AFAIK. I am also fine with filling the datalayer on client side with that information. And for sure the evaluation happens on client side as well.

vladbailescu: @kwin , I would definitely not oppose an **optional** client-side script to detect WCM mode (perhaps even changes, on the author) and update the data layer, if you think that would be valuable.

PR #1355, opened 2021-01-20
Add runtime statistics for the AdaptiveImageServlet (#1305)

### LINKED ISSUE #1292: AdaptiveImageServlet should provide runtime statistics

bpauli:
| Fixed Issues?            | Fixes #1292
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Documentation Provided   | Inline
| Any Dependency Changes?  |

The AdaptiveImageServlet now maintains statistics as Sling Metrics, which you can check via /system/console/slingmetrics.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -136,11 +136,13 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
                 if (StringUtils.isNotEmpty(suffixExtension)) {
                     if (!suffixExtension.equals(requestPathInfo.getExtension())) {
                         LOGGER.error("The suffix part defines a different extension than the request: {}.", suffix);
+                        metrics.markImageError();
                         response.sendError(HttpServletResponse.SC_NOT_FOUND);
                         return;
                     }
                 } else {
                     LOGGER.error("Invalid suffix: {}.", suffix);
+                    metrics.markImageError();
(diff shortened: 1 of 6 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -231,10 +238,13 @@ protected void doGet(@NotNull SlingHttpServletRequest request, @NotNull SlingHtt
                     transformAndStreamAsset(response, componentProperties, resizeWidth, quality, asset, imageType,
                             imageName);
                 }
+                metrics.markRenditionRendered();
```
vladbailescu, line 241: This is ambiguous to me, I'd rename it `markImageStreamed`

vladbailescu (changes requested):
I would also measure all errors (whenever we have a response.sendError)

vladbailescu (approved):

PR #1373, opened 2021-01-28
[Page] expose the caconfig HtmlPageItemsConfig as an API

jckautzmann:
Adds to PR #1272 
Fixes ticket #1234


bpauli (approved):
added ca config mocks in aacdf248487c387169be666c3e7bba81ac380142
I had to update the example content to the new expected API

LGTM

vladbailescu:
I tried out the wcm.io ca-config editor (using the sample app) and ended up with the following structure when creating a new `HtmlPageItemsConfig`:

image

That's different from the documented:

```
 * <pre>
 *      com.adobe.cq.wcm.core.components.config.HtmlPageItemsConfig
 *          - prefixPath="/some/path"
 *          + items
 *              + item01
 *                  - element=["link"|"script"|"meta"]
 *                  - location=["header"|"footer"]
 *                  + attributes
 *                      - attributeName01="attributeValue01"
 *                      - attributeName02="attributeValue02"
 *                      ...
 *              + item02
 *                  ...
 *              ...
 *  </pre>
 ```

@stefanseifert , is that because of the sample app itself?

Also, I noticed than after each save the `prefixPath` is populated but, on edit, its value is not being picked up (shows empty), same as the inheritance check:

image
shows as:
image

Its value is also ignored when in page rendering:
image

**Comment by vladbailescu** [comment]
The documentation of the attributes seems wrong, IIUC the code it should be:

```
 * <pre>
 *      com.adobe.cq.wcm.core.components.config.HtmlPageItemsConfig
 *          - prefixPath="/some/path"
 *          + items
 *              + item01
 *                  - element=["link"|"script"|"meta"]
 *                  - location=["header"|"footer"]
 *                  + attributes
 *                      + attribute01
 *                          - name="attribute01Name"
 *                          - value="attribute01Value"
 *                      + attribute02
 *                          - name="attribute02Name"
 *                          - value="attribute02Value"
 *                      ...
 *              + item02
 *                  ...
 *              ...
 *  </pre>
```

jckautzmann:
@vladbailescu I addressed the documentation issue with https://github.com/adobe/aem-core-wcm-components/pull/1373/commits/30f6e033226e9d033c1c9c8384bbf203a46cacf7

jckautzmann:
@dhardtke @stefanseifert Could you please:
1) validate that the current implementation in this PR meets your expectations for the wcm.io ca-config editor?
2) adapt the unit test `com.adobe.cq.wcm.core.components.internal.models.v2.PageImplTest#testPage()` so that it passes with the expected structure of the ca-config? It currently passes although there is no `jcr:content` node between the config and the `items` node [0]. It seems that the sling ca-config mocks miss some of the expected logic.

[0] https://github.com/adobe/aem-core-wcm-components/blob/issue/1234-HtmlPageItemsConfig-API-2/bundles/core/src/test/resources/page/v2/test-sling-configs.json

vladbailescu:
Moving to 2.15.0 milestone to allow us all more time to refine this.

dhardtke:
Hi @jckautzmann,

to answer at least question 1: The implementation itself seems fine, thanks.

Regarding question 2: I assume you meant to refer to `com.adobe.cq.wcm.core.components.internal.models.v2.PageImplTest#testHtmlPageItemsConfig` as `testPage` is merely a helper method preparing the tests and w.r.t. CA configs only loads the right one.

I tried changing the `test-sling-configs.json` to include the additional `jcr:content` node which led the test to fail. After debugging a bit I could not figure out why. It is also quite peculiar why the wcm.io editor creates the jcr:content node in the first place, when the "AEM Page" Persistence Strategy hasn't been installed & configured [1]
When configuring the `DataLayerConfig` this doesn't happen btw. So it must be related to the config somehow.

Maybe @stefanseifert knows why the editor creates the `jcr:content` node without the aforementioned Persistence Strategy.

[1] https://wcm.io/caconfig/extensions/persistence-strategies.html#Persistence_Strategy_AEM_Page

bpauli:
@dhardtke, @jckautzmann,

the addional `jcr:content` node for the items is required because of the `com.adobe.granite.conf.impl.PersistenceStrategy` implementation which comes OOTB with a higher service ranking in AEM. This CAConfig PersistanceStrategy is picked up before the default implmentation of Sling which is used by the caconfig-mock. Unless no other PersistanceStrategy is configured (e.g. the PagePersistenceStrategy which is provided by the wcm.io caconfig extension) the documentation in the Javadoc of the `HtmlPageItemsConfig` is correct. I will add a hint to the Javadoc and we can then merge this PR.

@stefanseifert the wcm.io editor seems not to work correctly with nested configurations and this persistace strategy as the `prefixPath` property on the root level disapears after child items are added. Any ideas why this happens? Thanks!

PR #1382, opened 2021-02-06
perf: Use and re-use ObjectWriter for data serialization (fixes #1381)

### LINKED ISSUE #1381: Improve Component#getData performance by re-using ObjectMapper
## Feature Request

No problem. Just an observation.

Right now `com.adobe.cq.wcm.core.components.internal.models.v1.datalayer.ComponentDataImpl#getJson` constructs a new `ObjectMapper` for each invocation of the method. 

Additionally, the configuration of the serializer in `com.adobe.cq.wcm.core.components.models.Component#getData` should be moved inside of `com.adobe.cq.wcm.core.components.internal.models.v1.datalayer.ComponentDataImpl#getJson` because right now a build-warning is generated since `ComponentData` is part of the public API (inside Export-Package) but `com.adobe.cq.wcm.core.components.internal.jackson.ComponentDataModelSerializer` is inside `internal`.

**Are there alternatives?**
The construction per invocation of `getJson`is unnecessary as ObjectMapper is thread-safe (as long as no configuration changes occur after constructing the object).

Alternatively or even better we can switch to `ObjectWriter` since we only use Jackson to serialize the Data for the JavaScript to consume as `ObjectWriter` is guaranteed to be immutable.

**Documentation**

dhardtke:
Also moves the configuration of ComponentData's serializer inside ComponentDataImpl

| Fixed Issues?            | Fixes https://github.com/adobe/aem-core-wcm-components/issues/1381
| Minor: New Feature?      | Maybe
| Tests Added + Pass?      | No new tests necessary
| Any Dependency Changes?  |


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Component.java
```diff
@@ -55,12 +52,10 @@ default String getId() {
      * Returns the data layer information associated with the component
      *
      * @return {@link ComponentData} object associated with the component
-     *
      * @since com.adobe.cq.wcm.core.components.models 12.12.0
      */
     @Nullable
     @JsonProperty("dataLayer")
-    @JsonSerialize(using = ComponentDataModelSerializer.class)
```
ky940819, line 63: The removal of this line breaks the serialization of the data model. 
The serialized format should be along the lines of:
```JSON
{
    "<ComponentID>": {
        "....serialized ComponentData..."
    }
}
```

The results of this change cause 
```JSON
{
    "....serialized ComponentData..."
}
```

ky940819 (changes requested):
This PR should probably be made against `adobe:development` and not `adobe:master`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Component.java
ky940819: Here's a diff of actual test output (expected on left, actual on right).
Please make sure to run the tests.

Screenshot from 2021-02-06 03-27-23

dhardtke: > This PR should probably be made against `adobe:development` and not `adobe:master`.

Thanks, I changed the base of this PR.

> The removal of this line breaks the serialization of the data model.
Yes, it did. The change was intentional, though. As I mentioned in the PR's description it's probably not a good idea to reference the serializer (which is located inside "internal") from within the public API.

As far as I can tell the serializer is only necessary to be used inside Unit tests, correct? Are consumers of the API expecting to be able to serialize instances of `Component` directly which in turn should use the Serializer? Or are consumers only supposed to call `ComponentData#getJson`?

(This might be a concern especially when using `ComponentExporter`)

vladbailescu: `Component/ContainerExporter` serialization is important for SPA developers (which consume `.model.json` output instead of rendered HTML attributes to feed the DataLayer).

PR #1385, opened 2021-02-07
feat: Init teaser dialog even if description field is not present (#1…

### LINKED ISSUE #1384: [Teaser] Enable "Get title from linked page" even when description field is not present
## Feature Request

We have a bunch of components extending the core teaser component. Some of them have the whole `descriptionGroup` field hidden in the `cq:dialog` node using `sling:hideChildren`.
Still, we would like to use the "Get title from linked page" feature in the dialog. https://github.com/adobe/aem-core-wcm-components/commit/7c1b6f958b85ff6ed8de2ba8bbc85b003eb752f5 made it so that the feature only works if the description field is present.

The JS should check if a description field is present. If it is not, it should just initialize the dialog anyway.

**Are there alternatives?**
We could provide a custom clientlib in our project which would essentially be a copy of the clientlib provided with the CC. But that would lead to a lot of code duplication.

**Documentation**

dhardtke:
| Fixed Issues?            | Fixes #1384
| Patch: Bug Fix?          |
| Minor: New Feature?      | Yes, depending on whether or not the use case should have been supported previously.
| Any Dependency Changes?  |

We initialize the teaser's dialog logic regardless of whether `.cq-RichText-editable[name="./jcr:description"]` is available or not.

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/clientlibs/editor/js/teaser.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/teaser/v1/teaser/clientlibs/editor/js/teaser.js
@@ -37,16 +37,21 @@
         var dialogContent = $dialogContent.length > 0 ? $dialogContent[0] : undefined;
 
         if (dialogContent) {
-
-            var rteInstance = $(descriptionTextfieldSelector).data("rteinstance");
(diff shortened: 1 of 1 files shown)


vladbailescu (approved):

richardhand:
Thanks @dhardtke for the contribution!

PR #1392, opened 2021-02-10
Fix/1390 panelselect should ignore wrapper divs

### LINKED ISSUE #1390: [Tabs] [Accordion] [Carousel] Panelselect doesn't work if proxy component adds new wrapper div
## Bug Report

**Current Behavior**
Suppose I have the following template inside my proxy component `tabs.html`:
```html
<div data-sly-include="/apps/core/wcm/components/tabs/v1/tabs/tabs.html" class="aem-tabs"></div>
```

This leads to the panelselect button to be missing inside the editor because of checks [here](https://github.com/adobe/aem-core-wcm-components/blob/8e9280e9241e63face5a81a5b4744888ef24a0e1/content/src/content/jcr_root/apps/core/wcm/components/commons/editor/clientlibs/panelcontainer/v1/js/panelContainerUtils.js#L61) and [here](https://github.com/adobe/aem-core-wcm-components/blob/8e9280e9241e63face5a81a5b4744888ef24a0e1/content/src/content/jcr_root/apps/core/wcm/components/commons/editor/clientlibs/panelcontainer/v1/js/panelContainerUtils.js#L109).

**Expected behavior/code**
The panelselect should still work like when the wrapper div is not present.

**Environment**
- AEM version and patch level: AEM 6.5.7
- Core Components version: 2.14.0
- JRE version: openjdk 11.0.10 2021-01-19

**Possible Solution**
I would suggest ditching the whole "is one of the children of the editable a panel container type" check and rather search recursively, e.g.
```javascript
getPanelContainerHTMLElement: function(editable) {
    var container = getPanelContainerType(editable);
    var element;

    if (container) {
        return $(editable.dom[0]).find(container.selector)[0];
    }

    return element;
},
```

(and since editable.dom is already a jQuery object we could call find directly on editable.dom, e.g. `editable.dom.find(container.selector)[0]`)

It's about this button:
image

dhardtke:
| Fixed Issues?            | Fixes #1390
| Tests Added + Pass?      | We don't have FE tests for editor clientlibs
| Any Dependency Changes?  |

Looking for the panel container recursively (by using jQuery.find()) allows us to ignore arbitrary wrapper elements and makes us much more flexible in terms of how the proxy component is structured.

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/accordion.html
@@ -14,6 +14,7 @@
   ~ limitations under the License.
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
 <div data-sly-use.accordion="com.adobe.cq.wcm.core.components.models.Accordion"
+     data-panelcontainer="${wcmmode.edit && 'accordion'}"
      id="${accordion.id}"
      class="cmp-accordion"
      data-cmp-is="accordion"
--- a/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/editorhook/js/panelcontainer.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/editorhook/js/panelcontainer.js
@@ -22,6 +22,7 @@
             window.CQ.CoreComponents.panelcontainer.v1.registry.register({
                 name: "cmp-accordion",
                 selector: ".cmp-accordion",
+                wrapperSelector: '[data-panelcontainer="accordion"]',
                 itemSelector: "[data-cmp-hook-accordion='panel']",
                 itemActiveSelector: "[data-cmp-hook-accordion='item'][data-cmp-expanded] [data-cmp-hook-accordion='panel']"
             });
--- a/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
@@ -15,6 +15,7 @@
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
 <div data-sly-use.carousel="com.adobe.cq.wcm.core.components.models.Carousel"
      data-sly-use.templates="core/wcm/components/commons/v1/templates.html"
+     data-panelcontainer="${wcmmode.edit && 'carousel'}"
      id="${carousel.id}"
      class="cmp-carousel"
      role="group"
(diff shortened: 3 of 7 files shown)


bpauli (changes requested):
@dhardtke Thank you for this enhancement! While testing your changes I found the following issue: Panel selector/ Component Toolbar is not working for tabs with "Default active item"
Steps to reproduce:
1. Install the provided patch for the panelContainerUtils.js
2. Create proxy component as mentioned in #1390 
3. Go to the component library at /content/core-components-examples/library/container/tabs.html
4. Select the panel container of the tab in the section "Default active item"
observed behaviour: The component toolbar is not showing up -> no panel selector available
If I rollback the change in the panelContainerUtils.js the component toolbar is showing up without the panel selector (which is expected because of #1390 )

Can you please try to fix this? Thanks again for your contribution!

dhardtke:
Hi @bpauli ,

thanks for your review - appreciate it :)

Unfortunately, the issue is a bit more complicated than I thought.
The reason the "Default active item" case does no longer work is the following:
1. When the page is loaded the following snippet ensures overlays of inactive / hidden panel items are disabled https://github.com/adobe/aem-core-wcm-components/blob/cffaeb54509d0b8b52c896f94f25ec2057c6212d/content/src/content/jcr_root/apps/core/wcm/components/commons/editor/clientlibs/panelcontainer/v1/js/contentTree.js#L56-L58
2. Because of the changes in the PR parent containers of e.g. the Tab component are also having their overlays (partially? for some reason only the "Default active item" case was broken) disabled

Hence, we need to find a reliable way to identify if an editable *is* e.g. a Tab component. I would propose changing the `name` property e.g. in https://github.com/adobe/aem-core-wcm-components/blob/cffaeb54509d0b8b52c896f94f25ec2057c6212d/content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/clientlibs/editorhook/js/panelcontainer.js#L23 to `name` (instead of the current `cmp-tabs`) and patching https://github.com/dhardtke/aem-core-wcm-components/blob/38f8e671b79d4e59a78478a9531021476f1c10a2/content/src/content/jcr_root/apps/core/wcm/components/commons/editor/clientlibs/panelcontainer/v1/js/panelContainerUtils.js#L99 to check using `if (editable.dom.hasClass(container.name)) {`

This is not perfect, as it relies on the container components writing all component names onto the wrapper divs' classList. But it works.

Do you have any better idea? I tried finding a method of accessing arbitrary properties set on the component node (e.g. I added a `isPanelContainer` to `/apps/core/wcm/components/tabs/v1/tabs`) from within the Editor's JS but was unsuccessful as `editable.config` only includes a subset of properties.

We could also think about adding a data attribute to the first wrapper div and keep the current implementation, but switch to this data attribute to identify whether an editable is a component with a particular panelcontainer.

dhardtke:
I just pushed another commit that now (hopefully) fixes all edge cases.

How it works is this:
- There is a new property `wrapperSelector` present to be supplied when registering "panel containers"
- `CQ.CoreComponents.panelcontainer.v1.utils.isPanelContainer` checks if the given editable is a panel container by preferring `wrapperSelector` or falling back to `selector` (previous behavior) if not present
- `CQ.CoreComponents.panelcontainer.v1.utils.getPanelContainerHTMLElement` always uses `selector` to find the right DOM element. It searches recursively for the panel container

You can try it out yourself by creating `/apps/core-components-examples/components/tabs/tabs.html` with the contents

```html
<div data-sly-include="/apps/core/wcm/components/tabs/v1/tabs/tabs.html" class="aem-tabs"></div>
```

bpauli (approved):

bpauli:
@dhardtke LGTM. Thank you for your contribution!

PR #1397, opened 2021-02-15
Issue/1367

### LINKED ISSUE #1367: Release Core Components JS as npm modules
For an improved integration of core components JS in an application that uses the core components, it would be very useful to have versioned NPM Modules of the Core Components' JS.

The @adobe namespace in npm could be used. The version of the components should be according to the core components release version. That would result, for example, in component names like:

@adobe/core/accordion/v1@2.13.4
@adobe/core/text/v1@2.13.4
@adobe/core/text/v2@2.13.4

This would be an easy step that helps a lot in integrating the components in modern FE build stacks.

The prerequisites for this are given because the js modules have a clean capsuled component scope. All the components would need is 
- an .npmrc per js file 
- an export default at the end of the js or a plugin that converts plain js into es6 modules
- an integration of this module release in the build job and finally in the build pipelines

jkoebner:
| Fixed Issues?            | Fixes #1367 
| Tests Added + Pass?      | Not required
| Documentation Provided   | No (code comments and or markdown)

Adding exports for all js modules to ease use of core components in different FE Setups. 
Adds a min.js to all core components that have a sites js client library. Requires the js files in their order and exports them. Also adds main js file for all components so the different components can be selected individually.

Usage: 
add core components dependency to your package.json, e.g.
`
	"dependencies": {
		"com.adobe.cq.core.wcm.components.content": "^2.13.5-SNAPSHOT"
	}
`
import the needed module in your js

`
import { accordion } from 'com.adobe.cq.core.wcm.components.content'; // eslint-disable-line
`

execute in constructor

`
accordion()
`


jkoebner:
i did update this now, get the code up to date with development and export all the additional js files that were added. We'd really appreciate support here and would love to be able to add this contribution to the code because it is very important for our component architecture that we have more flexibility in using the JS inside our application code. maybe we can get some support and feedback here @gabrielwalt  @vladbailescu  ?

PR #1408, opened 2021-02-18
[Container] Missing @Exporter annotation - JSON output skipped in model.json Export

### LINKED ISSUE #1407: [Container] Missing @Exporter annotation - JSON output skipped in model.json Export
The content of every page can be exported by appending model.json. Container components are exported as well, including their children.

However, if you inspect closely the output of a component resource referencing the proxy component with sling:resourceSuperType=`core/wcm/components/container/v1/container`, you will see JSON output like this:

```
{
	"columnClassNames": {
		"responsiveimage_933266850": "aem-GridColumn aem-GridColumn--default--12",
		"text_1649938976": "aem-GridColumn aem-GridColumn--default--12",
		"text": "aem-GridColumn aem-GridColumn--default--12",
		"title": "aem-GridColumn aem-GridColumn--default--12"
	},
	"columnCount": 12,
	"gridClassNames": "aem-Grid aem-Grid--12 aem-Grid--default--12",
	"allowedComponents": {
		"applicable": false,
		"components": [ ... ]
	},
	":items": { ... },
	":itemsOrder": [ ... ],
	":type": "myapp/components/container"
}
```

this JSON output contains property which are not defined in the model of the container component. instead, it is the output of the model associated with the old WCM foundation component `wcm/foundation/components/responsivegrid`.

the reason is a missing `@Exporter` annotation for the class `com.adobe.cq.wcm.core.components.internal.models.v1.LayoutContainerImpl`.

with this in place the output should change to something like this:

```
{
	"id": "container-09eae90b3e",
	"layout": "SIMPLE",
	":items": { ... },
	":itemsOrder": [ ... ],
	":type": "myapp/components/container"
}
```

stefanseifert:
Fixes #1407

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/LayoutContainerImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/LayoutContainerImpl.java
@@ -23,16 +23,20 @@ import javax.annotation.PostConstruct;
 
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.resource.Resource;
+import org.apache.sling.models.annotations.Exporter;
 import org.apache.sling.models.annotations.Model;
 import org.apache.sling.models.annotations.injectorspecific.ScriptVariable;
 import org.jetbrains.annotations.NotNull;
 
+import com.adobe.cq.export.json.ExporterConstants;
 import com.adobe.cq.wcm.core.components.models.LayoutContainer;
 
 /**
  * Layout container model implementation.
  */
 @Model(adaptables = SlingHttpServletRequest.class, adapters = LayoutContainer.class, resourceType = LayoutContainerImpl.RESOURCE_TYPE_V1)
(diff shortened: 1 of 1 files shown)


jckautzmann:
It would be nice to add a unit test to verify that the JSON output of the container component is available.

stefanseifert:
unit tests are already in place. the absence of this annotation does not change the result in unit tests.
it would only be possible to test it in integration tests from outside testing the model.json output.

vladbailescu:
I can see this breaking different use cases. Right now, with the missing annotations, the `ResponsiveGrid` exporter is used (through inheritance chain). We will probably need to ensure that when `responsiveGrid` layout is used we include the export from the `ResponsiveGrid` as well.

stefanseifert:
yes, this can definitely break applications relying on the "wrong" output of the ResponsiveGrid model. i do not see a benefit to switch models for simple/responsive layout - the output of the old ResponsiveGrid  is really not well suited for JSON export (outputting a lot of authoring-related stuff like allowedComponents which should not be contained in the JSON output).

a pity this annotation was not in place when the container was introduced. probably now this change has to be postponed to version 3.0.0 to no break things, this would be a clean cut.

PR #1410, opened 2021-02-23
Sample setup for the wcm.io caconfig editor

vladbailescu:
* Since #1234 is updating HtmlPageItemsConfig to be editable by wcm.io Context-Aware Configuration Editor, here's a sample setup.

image
image


examples/ui.apps/pom.xml
```diff
@@ -57,7 +57,18 @@
                 <artifactId>filevault-package-maven-plugin</artifactId>
                 <extensions>true</extensions>
                 <configuration>
-                    <packageType>application</packageType>
+                    <packageType>mixed</packageType>
```
stefanseifert, line 60: probably "container" is a better option than "mixed" which is discouraged?

jckautzmann:
Please add some docs (e.g. in the README) how to use this caconfig editor.

stefanseifert (commented):
i assume you also need a OSGi configuration to enable `https://wcm.io/caconfig/extensions/persistence-strategies.html` to make sure the complex config structres of HtmlPageItemsConfig is persisted and read correctly?
see https://wcm.io/caconfig/extensions/persistence-strategies.html

examples/ui.apps/pom.xml
vladbailescu: `container` does not allow content. We would need to create a `container` just for the bundles, which is not that helpful.

vladbailescu:
> Please add some docs (e.g. in the README) how to use this caconfig editor.

I added a file describing our configurations in https://github.com/adobe/aem-core-wcm-components/commit/b79b415c123604ff05e5f83650d75cc7dd1d0254

vladbailescu:
@stefanseifert :
> i assume you also need a OSGi configuration to enable `https://wcm.io/caconfig/extensions/persistence-strategies.html` to make sure the complex config structres of HtmlPageItemsConfig is persisted and read correctly?
> see https://wcm.io/caconfig/extensions/persistence-strategies.html

I added page persistence strategy configuration in 1c7ef06

jckautzmann (approved):

jckautzmann:
It would be nice to link the CONFIGS from the main README.

stefanseifert:
@vladbailescu 
> I added page persistence strategy configuration in 1c7ef06
ok - please note that `io.wcm.caconfig.extensions.persistence.impl.PagePersistenceStrategy` is not a factory but a singleton configuration - to probably the suffix `-components-library` should removed.

Issue #426, opened 2019-01-18
Couple of issues with the core Search component


tim-merrillcorp:
## Bug Report

**Current Behavior**
A clear and concise description of the behavior and your setup; if applicable, proxy component pattern and component version.

**Expected behavior/code**

**Environment**
- AEM version and patch level (e.g. AEM 6.3 SP1 CFP2)
- Core Components version (e.g. 2.0.4)
- JRE version (e.g. `Java(TM) SE Runtime Environment (build 1.8.0_152-b16)`)

**Possible Solution**

**Additional context / Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

richardhand:
@tim-merrillcorp, thanks for your report, I moved the second issue to #427.

As the results are provided on the fly, it seems like they're not run through the default link checker and therefore require handling to ensure they are mapped correctly.

gabrielwalt:
This should be fixed by #713.

kwin:
As #713 might take some time, why not simply calling `RR.map(...)` on the JSON value in the mean time? Would you accept a PR for that?

kwin:
@vladbailescu WDYT about a fix which would just call RR.map  on https://github.com/adobe/aem-core-wcm-components/blob/345377e9701a6724abc8ddf5a1fc5c3ade85f406/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java#L61?

vladbailescu:
@kwin, it would require some testcases to make sure it does not break stuff, but could be a solution. We would appreciate some help here, especially with the testcases.

kwin:
@vladbailescu Isn't this solved in 2.16.0 (https://github.com/adobe/aem-core-wcm-components/blob/2e055f9d4a1777155229dae22d948ce7ec095a42/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/SearchResultServlet.java#L305)?

olenaslyshkova:
As I can see the issue is still reproducible on the core component version 2.17.10.
Can be also seen in the https://wknd.site/us/en.html site (search for "magazine") which uses the  2.16.4 core components version.

cpiosecny:
The results in _SearchResultServlet_ are getting wrapped in _PageListItemImpl_ which is calling _LinkManager_ to build a Link.
[...]

Issue #62, opened 2017-09-11
Consistent and customizable site root detection


stefanseifert:
the core components currently implement different (and insufficient) site root detection strategies:
* Breadcrumb and Search (Sandbox) component use a "start level" parameter addressing an absolute parent page
* Navigation (Sandbox) and Language Navigation (Sandbox) provide a root path picker

only providing a "start level" parameter is limiting, because when set in the editable template for a structure element it requires all sites starting at the same level. this is not necessarily the case for more complex site setups.

defining the root level as path is even more inflexible - if defined in the template for the structure component this template can only use for exactly one site. if the navigation component is defined as editable in page, the path has to be entered for every page of the site again.

what is required is a strategy to detect the site root which is configured once, and used for all components the same way that requires a site root (and this strategy should be usable for custom components as well).

ACS AEM Commons implements such a feature, but it is insufficient as well because it is not possible to configure different strategies for different tenants or contant areas, and not possible to customize the site root lookup beyond using a regex: https://github.com/Adobe-Consulting-Services/acs-aem-commons/issues/1037

If the Site Root detection would be implemented as an OSGi service it would be possible to provide own strategies (should support multiple strategies to be tenant-aware). Another option is to use Sling Context-Aware Configuration to either lookup the configuration for a site, or directly use the "inner-most" Context-Aware Configuration context path as site root (this is the default strategy used by [wcm.io URL Handler](http://wcm.io/handler/url/).

joerghoh:
I would generalize the concept and go beyond of "site root". You could introduce the concept of a "Site", where the site root is an essential part of it. I could also contain much more site-specific properties (like domain name, other settings, etc).

In general there should be much more abstraction; at the moment the typical AEM application just consists of a bunch of pages, as the product does not ship with an abstraction which goes beyond the concept of "page".
Having such an abstraction (or levels of abstraction) could help a lot in the way how applications are structured and navigation is built.

stefanseifert:
this is also related to #63

henrykuijpers:
Very big +1 with @joerghoh 's comment!

I think the (relatively) new CAConfig concept can help out with the solution for these issues! The content policies are a good start, but if they can be backed (or replaced) with CAConfigs... :)

gabrielwalt:
(CQ-4247365)

PR #1411, opened 2021-02-25
[Link Handling][Stage 1] Make link references consistent and flexible #713 - Link and utilities

vladbailescu:
* Added new Link interface and LinkHandler utilities


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Externalized link URL.
+     * @return Link URL or null if link is invalid.
+     */
+    @Nullable
+    String getURL();
+
+    /**
+     * Map with Attributes for HTML Anchor tag for this links.
```
jckautzmann, line 46: typo: ... for this link.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
```diff
@@ -0,0 +1,61 @@
...
+     * This usually also contains the Link URL as <code>href</code> attribute,
+     * but may contain additional attributes like <code>target</code> and others.
+     * @return Map with HTML-specific anchor attributes, or an empty map if link is invalid
+     */
+    @NotNull
+    Map<String, String> getHtmlAttributes();
+
+    /**
+     * Returns the references target page if the links points to an internal page.
```
jckautzmann, line 55: typo: 
references -> referenced
links -> link

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -0,0 +1,167 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import com.adobe.cq.wcm.core.components.commons.link.Link;
+import com.day.cq.wcm.api.Page;
+import com.day.cq.wcm.api.PageManager;
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * Simple implementation for resolving and validating links from model's resources.
+ * This is a Sling model that can injected into other models using the <code>@Self</code> annotation.
```
jckautzmann, line 40: typo: ... that can **be** injected ...

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -0,0 +1,167 @@
...
+        return getLink(resource, PN_LINK_URL);
+    }
+
+    /**
+     * Resolves a link from the properties of the given resource.
+     * @param resource Resource
+     * @param linkURLPropertyName Property name to read link URL from.
+     * @return Link may be invalid, but is never null
+     */
+    public @NotNull Link getLink(@NotNull Resource resource, String linkURLPropertyName) {
```
jckautzmann, line 73: isn't `PN_LINK_URL` the only property name allowed to store a link? If so this method is not needed.
What's the use case for having other properties which store a link URL?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -0,0 +1,167 @@
...
+        Page targetPage = pageManager.getPage(linkURL);
+
+        return new LinkImpl(resolvedLinkURL, resolvedLinkTarget, targetPage);
+    }
+
+    /**
+     * Returns an invalid link.
+     * @return Invalid link, never null
+     */
+    public @NotNull Link getInvalid() {
```
jckautzmann, line 112: I'd rename it `getInvalidLink` to be consistent with the other `getLink` methods.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -0,0 +1,167 @@
...
+        if (!StringUtils.isEmpty(linkURL)) {
+            return getLinkURL(linkURL);
+        }
+        else {
+            return null;
+        }
+        
+    }
+
+    private String validateAndResolverLinkTarget(String linkTarget) {
```
jckautzmann, line 126: typo: validateAndResolverLinkTarget -> validateAndResolveLinkTarget

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -0,0 +1,167 @@
...
+     * If the provided {@code path} identifies a {@link Page}, this method will generate the correct URL for the page. Otherwise the
+     * original {@code String} is returned.
+     * @param path the page path
+     * @return the URL of the page identified by the provided {@code path}, or the original {@code path} if this doesn't identify a {@link Page}
+     */
+    @NotNull
+    private String getLinkURL(@NotNull String path) {
+        Page page = pageManager.getPage(path);
+        if (page != null) {
+            return getPageLinkURL( page);
```
jckautzmann, line 145: uneeded space before `page`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
```diff
@@ -0,0 +1,100 @@
...
+    public LinkImpl(String url, String target) {
+        this(url, target, null);
+    }
+
+    /**
+     * @param url Link URL
+     * @param target Target
+     * @param targetPage Target page
+     */
+    LinkImpl(String url, String target, Page targetPage) {
```
jckautzmann, line 61: I find this constructor confusing. We could have a `url` pointing to `page1` and `targetPage` pointing to `page2`. Where would the link point to in this case?
I'd rather introduce following constructors:
```
LinkImpl(Page targetPage)
LinkImpl(Page targetPage, String target)
```
which would mirror the other 2 constructors.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
```diff
@@ -0,0 +1,100 @@
...
+        this.url = url;
+        this.htmlAttributes = buildHtmlAttributes(url, target);
+        this.targetPage = targetPage;
+    }
+
+    @Override
+    public boolean isValid() {
+        return url != null;
```
jckautzmann, line 69: Shouldn't we also check for following conditions?
a) target page is null:
- url is not blank
- url is validated by the AEM link checker
b) else:
- target page exists

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    public @Nullable String getURL() {
+        return url;
```
jckautzmann, line 74: shouldn't we return the canonical URL?

jckautzmann: we could introduce a method that returns the resolved URL, based on the `url` string or on the target page:
`public @Nullable String getResolvedURL()`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
jckautzmann: I just realised that we currently have different property names (link, linkURL) in the core components. Having above method would help keeping backwards compatibility. So let's keep it.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return true if component has a valid link defined
+     */
+    boolean isValid();
+
+    /**
+     * Externalized link URL.
+     * @return Link URL or null if link is invalid.
+     */
+    @Nullable
+    String getURL();
```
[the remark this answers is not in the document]
jckautzmann: I agree with @kwin . I suggested earlier to keep this method and to introduce a method that returns the resolved URL, based on the url string or on the target page:
```
public @Nullable String getResolvedURL()
```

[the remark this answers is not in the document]
stefanseifert: you're right: the javadoc is misleading here.

but keep in mind that this interface should represent *any* type of link, not only links to target pages. so it could also return a link to an external website, or to an asset (e.g. PDF download).
the goal of the method is to have a ready-to-use URL for all contexts. so it should be as mappend and as externalized as possible, and also usable in JSON output (where is no LinkTransformer running over it afterwards).
on the other hand we are not that far with the other parts for the implementation to really externalize URLs (no concept for site root #62, not means for configuring URLs to externalize for site context etc.). not possible to introduce all this in this PR we have to start simple.
what indeed could make sense is to introduce returning a fully mapped method here, we already have support for vanity paths.
i'm not a fan of using Sling Rewriter to rewrite/map URLs - my goal would be the core components are fully functional without any post processing transformation step. mapped URLs should work fine even a Sling Rewriter is running over it afterwards.

i would not like to introduce another URL like getResolvedURL, that get's very confusing for the developers using the interface - and will mostly use to misuse as no one understands the difference (think of ResourceResolver.getResource vs. ResourceResolver.resolve).

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
stefanseifert: (sorry for all my typos, thanks for taking care of...)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
stefanseifert: yes - alternative 1:
use method names like `get(...)` and `getInvalid()`
alternative 2:
use method names like `getLink(...)` and `getInvalidLink()`
for consistency. i slightly tend to alternative 1.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
stefanseifert: the point is: link objects can represent any type of links - link to target pages, link to assets, link to external URLs.
the target page is just an optional parameter given as extra metadata together with an internal link.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
stefanseifert: validating the target page is already done in the LinkHandler, not need to do it here again. "LinkImpl" is only a POJO.
support for validating external inks using AEM link checker could be added later in the LinkHandler, but let's start simple.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
stefanseifert: see discussion above - there is no "canocical URL" e.g. for external links.

stefanseifert:
@vladbailescu  thanks for picking this up!

stefanseifert (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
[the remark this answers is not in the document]
kwin: The problem is that calling ResourceResolver.map() on an already mapped url leads to double URL encoding (https://issues.apache.org/jira/browse/SLING-1899). Therefore mapped urls which are processed by the Link Transformer might get invalid.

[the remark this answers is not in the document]
stefanseifert: i know, this flaw is in the Sling API for years - we have to de-encode it in the LinkHandler like [here](https://github.com/wcm-io/wcm-io-handler/blob/5a620b0fbbc1c99341f4bbd5978f5d6edcf80250/url/src/main/java/io/wcm/handler/url/impl/Externalizer.java#L54-L100). it's ugly, but works well; we're doing it for years.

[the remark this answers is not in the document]
stefanseifert: to correct myself: the code mentioned is not decoding the URL after calling the mapping methods. it should be kept URL-encoded if it contains special chars that are invalid unencoded in URLs. this way they should be valid in all use cases - HTL or JSON. i've not had a problem with sling rewriter. if you have a custom link rewriter in place you should decode the URL before processing it like [here](https://github.com/wcm-io/wcm-io-handler/blob/5a620b0fbbc1c99341f4bbd5978f5d6edcf80250/url/src/main/java/io/wcm/handler/url/rewriter/impl/UrlExternalizerTransformer.java#L101).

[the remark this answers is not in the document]
stefanseifert: but let keep us this PR small - it is a first step introducing a dedicated interfaces representing link, with not much more capabilites than the existing URL generation. let's add mapping support and all later. so maybe, just now only fix the misleading javadoc.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    @NotNull
+    Map<String, String> getHtmlAttributes();
+
+    /**
+     * Returns the references target page if the links points to an internal page.
+     * @return Target page or null
+     */
+    @Nullable
+    Page getTargetPage();
```
ky940819, line 59: My personal opinion: I wouldn't even allow getting the target page for a link. 
If models want to provide pages, they can provide pages. I feel that including a mechanism to go backwards from a link to a page is going to lead people to abuse the Link interface as a vehicle for passing Page object references. 

This would be fine, except for the fact that the concept of a Link should be (IMHO) target-agnostic (e.g. page, asset, external, whatever); thus, it's a weak interface that can't be trusted to provide the target page of a link and should not be relied upon to provide that Page object. It requires me to know how the implementation that provided the Link constructed it.

It should be just as valid to add a `getTargetAsset()`, or half a dozen other things that it could point to. 

If we really wanted a reliable mechanism for doing this then I think we would need to have a `PageLink` and `AssetLink` extensions of `Link`, that promise to provide the underlying [not null] object.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
jckautzmann: I much prefer well describing method names (alternative 2)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
stefanseifert: i think it's important to have access to target page (and yes, also target asset) - think of providing additional metadata for client data layer for example. but we can make it a bit less obvious to make clear it should not be used just to build a link.
we could make the Link interface adaptable, and allow to adapt to a CQ Page or asset, if the link point to any of those.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
vladbailescu: Done in acf1edb

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
vladbailescu: Done in acf1edb

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
vladbailescu: Done in acf1edb

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
vladbailescu: I think `LinkHandler#getLink` and `LinkHandler#getInvalidLink` make for more self-explanatory names. Done in acf1edb

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
vladbailescu: Done in acf1edb

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
vladbailescu: Done in acf1edb

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
vladbailescu: Since `LinkImpl` is only a POJO and the logic is in `LinkHandler` it makes sense to have this.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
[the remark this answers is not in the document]
vladbailescu: For now I updated the Javadoc in (acf1edb) to mention that the URL has support for context-path and vanity paths. I'm not sure having something like `getOriginalURL` would be that helpful.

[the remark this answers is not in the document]
kwin: IMHO it definitely needs to be maped (see e.g. https://github.com/adobe/aem-core-wcm-components/issues/426) for responses not handled by the link transformer, like JSON

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
vladbailescu: I agree with @ky940819 on this one, having something like `PageLink` and `AssetLink` that extend `Link` make for a better contract.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkConstants.java
```diff
@@ -0,0 +1,41 @@
...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+package com.adobe.cq.wcm.core.components.commons.link;
+
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * Constants for link handling.
+ */
+@ProviderType
+public final class LinkConstants {
```
vladbailescu, line 24: I'm not a fan of having the constants in a different class, I'd move them to `Link`.

stefanseifert: sonarcube usually complaints about putting constants in interface, and i agree it's better to put them in separate classes.
think of NameConstants in the WCM API.

jckautzmann: we may want to introduce a central class for constants, e.g. `CoreCmpConstants`

vladbailescu: To me, constants such as property names are actually part of API so they should also be part of the interface that defines the API.

vladbailescu: I think many people view this as an anti-pattern because Bloch wrote:

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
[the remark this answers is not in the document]
vladbailescu: @kwin Would mapping as https://github.com/adobe/aem-core-wcm-components/pull/1411/commits/d7ed7a29c4c2ab96a86aa5f18802d61b2cbd4978#diff-3f692edefc2435524e017da0411675032dc8d070acf852ff46f291b73c146b39L182 be enough?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
vladbailescu: Having thought a bit about this, I think it's a good use-case for generics, as proposed in d7ed7a2.

Then, `Teaser` could define `Link<Page> getLink()` while `Download` could define `Link<Asset> getLink()` and `Link#getReference` will return the correct type. It will also be easy to add new types of references without needing to add extra interfaces. WDYT @stefanseifert & @ky940819 ?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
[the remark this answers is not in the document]
kwin: Please always use the two argument version of map taking also the request into account. Some mapping rules may depend on the HOST header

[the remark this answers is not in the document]
vladbailescu: @kwin updated to two arguments version in 980b71e

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
stefanseifert: generics are eliminated at run time, and if you get a link from a model/link handler it's difficult to detect if it's link to a page or asset - you would have to get the object returned be getTarget and test it's type. 
it makes usage of the Link interface it a bit inconvenient as in most case you just deal with the link and do not care about the targeted type, so you always have to use Link<?> instead of just Link (or get probably warnings in IDE that you do not specify the generics type).

vladbailescu: Indeed, generics are removed at run time but Java is statically-typed so all checks happen at initial compilation, if all the handlers and models respect the generics annotations, there will be no need for checking of type at runtime.

So yes, the helpers and models would need to specify the generics in the return type of their `getLink`. But I believe they would always know what that link should point to.

A handler that can work with multiple target types will probably expose different methods: `Link<Page> getPageLink(...)` and `Link<Asset> getAssetLink(...)`. It would need to have different logic for them anyway.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkConstants.java
vladbailescu: Moved constants in ccbfe54

jckautzmann:
Some remarks:
1. IMHO the `LinkHandler#getLink` methods should return null if `Link.getURL` is null or empty.
2. I would also get rid of `LinkHandler#getInvalidLink`: a link is invalid if Link.isValid returns `true`.
3. IMHO when a link is not defined in the JCR there should be no JSON output. With the current implementation the JSON output of a title that has no link is:
```
"link": { "valid": false }
```

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
vladbailescu: @stefanseifert , unless there's a clear -1 from you, I would like to merge these changes to the `development` branch, along with [Stage 2](#1413). We will allow for a couple more weeks until including the changes in a release (we need at least to prepare all feature branches for [Stage 3]), that will give us some time to revisit and adjust the interfaces.

vladbailescu:
@jckautzmann , the `LinkHandler` is an internal class (for now, we might want to make this a public utility later on) so we can refactor/adjust down the road. I get your point but this can be solved at Sling Model level:

```
private Link link; // for lazy loading/caching

public Link getLink() {
    if (link == null) {
       link = linkHandler.getLink(..);
    }
    if (link.isValid()) {
        return link;
    }
    return null;
}
```

Whereas if you allow `LinkHandler` to return `null`, implementations that need to get the link would need to be refactored to something like:
```
public String getURL() {
    return Optional.ofNullable(linkHandler.getLink(resource))
            .map(Link::getURL)
            .orElse(null);
}
```
In this case, I would rather the `LinkHandler` returned the `Optional` itself.

jckautzmann:
@vladbailescu That would work but we would end up with a lot of duplicated code for all the models that have links. I would prefer to have this code in a central place.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
stefanseifert: agreed - this PR gets a bit confusing as it mixes up the fundamentals of API discussion for link handling (which we already did on the first PR in 2019 or so) with the normal code review. it's fine to have new views and opinions on the best API design, but it get's cluttered up in this PR. probably it would have been better to put up a PR with only the interfaces first and dicusss only the public APIs - it's the only thing we cannot change easily after the first release.

vladbailescu: PR for the interface should have been [Stage 0], my bad 🤦 ! I'll merge this as is and continue the refactoring of `LinkHandler` implementation in [Stage 2] PR (#1413).

vladbailescu:
@jckautzmann Merged #1411 and #1413 as we were straying too far with all the rework. Created #1431 for not returning invalid links.

PR #1413, opened 2021-02-25
[Link Handling][Stage 2] Make link references consistent and flexible #713 - Updating existing components

vladbailescu:
* Merged changes to existing Sling Models


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -59,6 +60,7 @@
      * Reference to {@link PageManager}
      */
     @ScriptVariable
+    @Optional
```
kwin, line 63: Wouldn't it make more sense to rely on the `PageManagerFactory` OSGi service (and make that mandatory?)

vladbailescu: For most cases the `PageManager` is readily available in the script bindings so it does not make sense to go back to the factory. For those few cases where it's unavailable (mostly servlet contexts) adapting the `ResourceResolver` should be equivalent to getting it from the factory.

PR #1419, opened 2021-03-01
[Title][LinkHanlder][Stage3] Create Title v3 and integrate the new link handler feature

jckautzmann:
Pick und udpate changes from PR #862.
Fixes #713

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/TitleImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/TitleImpl.java
@@ -1,0 +1,51 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2021 Adobe
+ ~
+ ~ Licensed under the Apache License, Version 2.0 (the "License");
+ ~ you may not use this file except in compliance with the License.
+ ~ You may obtain a copy of the License at
+ ~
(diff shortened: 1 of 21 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/TitleImpl.java
```diff
@@ -0,0 +1,54 @@
...
+import com.adobe.cq.wcm.core.components.commons.link.Link;
+import com.adobe.cq.wcm.core.components.models.Title;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+@Model(adaptables = SlingHttpServletRequest.class,
+        adapters = {Title.class, ComponentExporter.class},
+        resourceType = TitleImpl.RESOURCE_TYPE)
+@Exporter(name = ExporterConstants.SLING_MODEL_EXPORTER_NAME,
+        extensions = ExporterConstants.SLING_MODEL_EXTENSION)
+public class TitleImpl extends com.adobe.cq.wcm.core.components.internal.models.v1.TitleImpl implements Title {
```
abidaJustice, line 35: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names shouldn’t shadow simple name of superclass (NM_SAME_SIMPLE_NAME_AS_SUPERCLASS)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-shouldn-t-shadow-simple-name-of-superclass-nm-same-simple-name-as-superclass).

vladbailescu (approved):

PR #1420, opened 2021-03-01
[Button][LinkHandler][Stage3] Create Button v2 and integrate the new link handler feature

jckautzmann:
- create a new version Button v2 that integrates the new link handler feature
- Pick und udpate changes from PR #862.
- Fixes #713


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ButtonImpl.java
```diff
@@ -0,0 +1,54 @@
...
+@Model(
+        adaptables = SlingHttpServletRequest.class,
+        adapters = {Button.class, ComponentExporter.class},
+        resourceType = ButtonImpl.RESOURCE_TYPE
+)
+@Exporter(
+        name = ExporterConstants.SLING_MODEL_EXPORTER_NAME,
+        extensions = ExporterConstants.SLING_MODEL_EXTENSION
+)
+public class ButtonImpl extends com.adobe.cq.wcm.core.components.internal.models.v1.ButtonImpl {
```
abidaJustice, line 38: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names shouldn’t shadow simple name of superclass (NM_SAME_SIMPLE_NAME_AS_SUPERCLASS)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-shouldn-t-shadow-simple-name-of-superclass-nm-same-simple-name-as-superclass).

content/pom.xml
```diff
@@ -274,7 +274,7 @@
                                         <replace file="target/jcr_root/libs/core/wcm/components/search/v1/search/.content.xml" token=".core-wcm" value="Core Structure" />
 
                                         <replace file="target/jcr_root/libs/core/wcm/components/accordion/v1/accordion/.content.xml" token=".core-wcm" value="Core Content" />
-                                        <replace file="target/jcr_root/libs/core/wcm/components/button/v1/button/.content.xml" token=".core-wcm" value="Core Content" />
+                                        <replace file="target/jcr_root/libs/core/wcm/components/button/v2/button/.content.xml" token=".core-wcm" value="Core Content" />
```
vladbailescu, line 277: We should not expose this in a public group until it's ready for production

vladbailescu (approved):

content/pom.xml
jckautzmann: Addressed in https://github.com/adobe/aem-core-wcm-components/pull/1420/commits/8276d5f09fd31d01840c8e2160e97aa9653f1b3a

Issue #1374, opened 2021-01-29
[SPA/StyleSystem] Support for inclusion of Stylesystem props in Sling Models


friendlymahi:
## Feature Request

**Is your feature request related to a problem? Please describe.**
As of date, leveraging SPAEditor integration comes with limitations like not being able to use Style System capability of AEM, which is very powerful for managing the styles library one may load into the policies. And Core Components on the other hand are a great feature no one wants to rebuild as new components. So to start with we need ability for Core Components to be able to export additional styles to apply, via Sling Models, so it is up to the consumer to leverage that integration. And its for this reason we see a need for this enhancement in this repo.

**Describe the solution you'd like**
Component interface that we have today in Core Components provides methods like getData to support data layer. Similarly we can provide getStyles or getStylesToAttach for each component, and include the logic to bind component policy and read the styles applied using dialog. 

**Are there alternatives?**
We may be forced to re-do every sling model using the delegation pattern which is not ideal especially when we know we can resolve this issue from one place which shall eventually serve all. We have sample code that is used in this alternative that can go just in one place so there is no need for customizing sling models or write brand new unit tests when using the delegation pattern for things that are inevitable

**Documentation**
As with normal pages (non SPA), we use the paint brush to select the styles. And now its for the React/Angular components obtain this metadata for styles and wrap their components accordingly. Here is a sample snippet. We exposed sling model to return style props as **additionalStyles**, and in React we use **className** attribute to map the value to.
```
import React from 'react';
import { Heading } from 'my-ui-controls';
import './style.scss'

export default (props) => {
    const renamedProps = (({ type: tag = 'h1', additionalStyles: className = "headerdefault", ...remainingProps }) => ({ tag, className, ...remainingProps }))(props);
    return <div className="dbp-cmp-title" >
        <Heading {...renamedProps} />
    </div>
}
```

friendlymahi:
@joerghoh @godanny86 -Any thoughts on this? I am tagging you as I saw your inputs very helpful in the past. Wondering if I can contribute to this. Thank you

pfauchere:
Hi,

Thank you very much for the input.

We captured the need and intend to explore the topic soon.

That said, and if we are able to decorate Sling Models when they are being exported, I am not certain that the contribution will apply to the current repository. I cannot share or confirm any architecture decision as the work hasn't started yet.

Internal Ref: CQ-4292931

friendlymahi:
[...]

PR #1421, opened 2021-03-02
[StyleSystem/SPAEditor] Allow BaseComponent Sling Model to export CSS classes selected by the content author

friendlymahi:
| Fixed Issues?            | #1374 

Refer to issue #1374 for complete details. StyleSystem is a powerful feature that's missing for SPA Editor implementation. This shall result in creation of too many proxy components. 

However keeping the SPA Editor aside, Sling Models when exporting CSS class names as part of the Json shall enable consuming applications to associate equivalent styles in the respective apps. 

Fix Details: As child item models don't retrieve componentContext specific to the resource for which model json is being generated, a utility method to extract the component policy and eventually determine the CSS classes for the mapped cq:StyleIds when an author selects a style on the dialog, is added and invoked from the root Component.java/AbstractComponent.java classes. Required test data and unit tests are included. Now you shall see a new json prop addlCssClassNames included in all non page items of the model.json

Note that this is not a fix for SPA Editor to support Style System. However this is essential for SPA Editor or any custom implementation to leverage StyleSystem support.


friendlymahi:
@vladbailescu - Good Morning to you !. Please review my PR and advise me for any changes. I made the code changes and tested the same against AEM v6.5.2 and AEM v6.5.6

friendlymahi:
@vladbailescu - Hello sorry for any trouble. For some reason the GPG check fails. Can you advise on this? Thank you 1

vladbailescu:
@friendlymahi , thank you for your contribution! Before we begin the review, could you please make sure to merge the latest changes from `development` branch, so we have a clean changeset?

friendlymahi:
> @friendlymahi , thank you for your contribution! Before we begin the review, could you please make sure to merge the latest changes from `development` branch, so we have a clean changeset?

@vladbailescu - Completed merging Dev to my branch so the PR looks clean. Ran the unit tests and verified that the new changes I did work as expected. Thank you !

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Component.java
```diff
@@ -64,6 +64,19 @@ default String getId() {
...
+    /**
+     * Returns the style system information associated with the component
+     *
+     * @return CSS classes selected by the content author delimited using a SPACE character
+     *
+     * @since com.adobe.cq.wcm.core.components.models 12.20.0
+     */
+    @Nullable
+    @JsonProperty("addlCssClassNames")
+    default String getStyleSystemClasses() {
```
vladbailescu, line 77: I would name this just `getAppliedStyleClasses`. Also, I would rather we return a list here instead of a string, it would make checking if a certain class is set so much easier.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
```diff
@@ -49,6 +59,38 @@
...
+
+	 /**
+     * Equivalent CSS classes for styleIds
+     */
+	public static final String CQ_STYLECLASSES = "cq:styleClasses";
+	
+	 /**
+     * Default CSS class to provide in absence of any selection
+     */
+	public static final String CQ_STYLECLASSES_DEFAULT = "cq:styleDefaultClasses";
```
vladbailescu, line 93: We should use the constants already defined in `com.adobe.cq.wcm.style.StyleConstants` instead of redefining them.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
```diff
@@ -170,4 +212,88 @@ private static String generateId(@NotNull final Resource resource,
...
+					    String styleid = entry.getKey();
+					    String styleClass = entry.getValue();
+					    if (ArrayUtils.contains(styleids, styleid)) {
+					    	styleClasses.add(styleClass);
+						}
+					  }
+			}
+		
+		return styleClasses;
+	}
```
vladbailescu, line 298: This should leverage the implementation of `ComponentStyleInfo#getAppliedStyles` instead of reimplementing everything.

.gitignore
```diff
@@ -20,3 +20,6 @@ atlassian-ide-plugin.xml
 node
 node_modules
 core.wcm.components.commons.datalayer.*
+*/**/META-INF/MANIFEST.MF
+*/**/package-lock.json
+jacoco.exec
```
vladbailescu, line 25: These should not be necessary

vladbailescu (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
friendlymahi: Can you please share the GitHub URL for this class? I wasn't able to locate. If its part of a different bundle or is-a different service, please advise if I can do the required integration (Sorry I didn't ever come across this class). Thank you !

.gitignore
friendlymahi: Please let me know if you want me to undo the changes? Thank you !

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
friendlymahi: Wasn't able to locate the StyleConstants class using project dependencies. If you can please share the reference URL to the code you are referring to, I shall make the required changes. Thank you !

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Component.java
friendlymahi: Done. Please review the updated code and mark this as Resolved if you are satisfied. Thank you !

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
vladbailescu: Both `com.adobe.cq.wcm.style.StyleConstants` and `ComponentStyleInfo` are exposed by the `com.adobe.cq.com.adobe.cq.wcm.style` bundle as internal AEM APIs. Perhaps it makes sense to make those public so they could be leveraged here. I'll start a conversation with the rest of the development team.

friendlymahi: @vladbailescu - Does this mean, the changes shall appear in next version of uber jar? Also any timeline on this? Just wanted to set expectation to my team on leveraging these changes so they don't end up creating new proxy components and sling models to support the capabilities

vladbailescu: @friendlymahi, the package should now be available in the AEM SDK: https://javadoc.io/doc/com.adobe.aem/aem-sdk-api/latest/com/adobe/cq/wcm/style/package-summary.html (since `2021.3.5026.20210309T210727Z-210225`)

vladbailescu: It should also be available in the 6.5.8 uber-jar (though https://javadoc.io/doc/com.adobe.aem/uber-jar/latest/index.html is not updated yet)

vladbailescu: The usage pattern is `resource.adaptTo(ComponentStyleInfo.class)`

friendlymahi: @vladbailescu - Made the changes per your recommendation. However I did come across some issues and I had to take few extra steps that may require your review.

- Updating uber jar version broke some existing classes as the corresponding interfaces seem to be having new methods that require impl in child classes. So added placeholder methods in EnhancedRendition.java and MockHtmlLibraryManager.java
- SlingBindings didn't work in Junit. So updated Sling API version in POM file (was already targeting only scope=test) with v2.22.0 (matching with what we have in AEM v6.5.8)
- After these changes, for some reason OSGI services injection fail within my Junit. Unsure if the same can happen in pipeline aswell.

Please advise on the next steps. Thank you !

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Component.java
friendlymahi: @vladbailescu - Changed the method signature to match with existing ComponentStyleInfo.class behavior. List is replaced with String. Reason - ComponentStyleInfo.class#getAppliedStyleClasses() doesn't return list of strings, but returns list of StyleInfo beans. So parsing them again will duplicate the logic. So used the existing string method as-is. If you feel otherwise please let me know. Thank you !

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
```diff
@@ -135,7 +137,28 @@ public ComponentData getData() {
...
+		ComponentStyleInfo componentStyleInfo = this.resource.adaptTo(ComponentStyleInfo.class);
+		String appliedCssClassNames = null;
+		
+		if(componentStyleInfo != null)
+		{
+			appliedCssClassNames = componentStyleInfo.getAppliedCssClasses();
+		}
+
+		// Returning null so sling model exporters don't return anything for this property if not configured
+		return StringUtils.defaultIfBlank(appliedCssClassNames, null);
```
ky940819, line 159: This can be simplified, if not caching the `appliedCssClassNames` then it could just be 

```JAVA
public String getAppliedCssClasses() {
    return Optional.ofNullable(this.resource.adaptTo(ComponentStyleInfo.class))
        .map(ComponentStyleInfo::getAppliedCssClasses)
        .filter(StringUtils::isNotBlank)
        .orElse(null);
}
```

Regardless of if this change is made, can the formatting style remain consistent with the rest of the project, just to make it easier to read?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
```diff
@@ -35,139 +36,167 @@
...
+								&& StringUtils.equals(parentContextPage.getPath(), currentPage.getPath()));
+						inTemplate = (template != null
+								&& parentContextResource.getPath().startsWith(template.getPath()));
+						if (inCurrentPage || inTemplate) {
+							path = parentContextResource.getPath().concat(resource.getPath());
+							break;
+						}
+					}
+					parentContext = parentContext.getParent();
+				}
```
ky940819, line 155: It looks like your IDE auto applied line length restrictions. 
I'm not sure that anything in this file actually needs to be changed, but I could be missing something since the diff covers most of the file.

ky940819: Ah, i see what's happened, you have changed the files to tab indentation instead of space indentation. 
Does this project have an actual style guide?
I think everything else is always space indentation.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
friendlymahi: @ky940819 - Taken care of !

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
friendlymahi: @ky940819 - Not that I am aware of. But if you want me to change it to original format, I can do that. Please let me know. Thanks !

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractComponentImpl.java
friendlymahi: @ky940819 - This is now taken care of !

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v1/ComponentImplTest.java
```diff
@@ -61,13 +79,32 @@ public void testReferencedComponentId() {
...
-
+    
+    
+    @Test
+    public void testStyleSystemClasses()
+    {
+    	final String WE_RETAIL_TITLE = TEST_PAGE_EN+"/jcr:content/root/title_core";
+    	Object[] mappings = null;
+    	String[] styleClasses = getComponentUnderTest(WE_RETAIL_TITLE,mappings).getAppliedCssClasses().split(StringUtils.SPACE);
+    	assertTrue(styleClasses!=null && styleClasses.length == 2);
```
ky940819, line 90: Potential NPE, `styleClasses` can't possibly be null because `String#split()` doesn't return null. 
However, `Component#getAppliedCssClasses()` does potentially return null.
The null check should be done with `Assertions#assertNotNull()`.

Check of the array value should be done with `Assertions.assertArrayEquals()` instead of just checking the length;; or, since `getAppliedCssClasses()` returns a string instead of an array, just check the string value against the expected value. 

Additionally, should check the other situations, i.e. when the string is empty and/or not set.

friendlymahi: @ky940819 - Please take a look at the updated code. Since we are just calling another re-usable function, I was testing for just 1 usecase, so we dont repeat re-testing all the use cases that were originally tested in ComponentStyleInfo.class. However, if you feel it would be nice to do the due deligence. Please let me know. I originally had those test cases before we brought this ComponentStyleInfo reference. Thanks in advance !

ky940819: I was just thinking that you have logic that returns null (i.e. when string is blank, or is null). 
It would be nice to have this covered such that if It is ever changed to return something else in that case, it would be caught by the tests. 

I'm not suggesting re-testing all of ComponentStyleInfo, I'm just thinking of the situations:

- `Mockito.doReturn("   ").when(componentStyleInfoMock).getAppliedCssClasses();` -> null
- `Mockito.doReturn(null).when(componentStyleInfoMock).getAppliedCssClasses();` -> null

I'd also consider changing
```JAVA
String[] styleClasses = getComponentUnderTest(WE_RETAIL_TITLE,mappings).getAppliedCssClasses().split(StringUtils.SPACE);
assertTrue(styleClasses!=null && styleClasses.length == 2);
```

to
```JAVA
String[] styleClasses = getComponentUnderTest(WE_RETAIL_TITLE,mappings).getAppliedCssClasses();
Assertions.assertNotNull(styleClasses);
Assertions.assertEquals("class1 class2", styleClasses);
```

friendlymahi: @ky940819 - This is now taken care of !

.gitignore
friendlymahi: Taken care of !

friendlymahi:
@ky940819 @vladbailescu - Pretty much all feedback is incorporated. Please take one final look and advise me for next steps. I didn't update POM version or CHANGELOG yet, awaiting for your instructions. Thank you !

vladbailescu:
@friendlymahi , we had a feature review with our Product Manager (@gabrielwalt) and identified that, in the future, we will need to be able to categorise applied class names based on their source:
* style system
* responsive grid/layout
We're still discussing about how these should be structured but we know it would be nice to be able to do something like:

friendlymahi:
@vladbailescu @gabrielwalt - Thanks for the update. Here is my understanding. As of date classes provided by responsive grid are determined or included in the sling model of layout container. Even though the column size and offset are stored on individual nodes, they are interpreted either while rendering the resource or while generating exporter for parent grid component. However with style system, policy or style ids are determined at individual component and has no dependency on parent. Similarly panel container components have panel title stored at component level but is available via exporter or while rendering only when viewed via panel container perspective, so invoking sling model for the panel item doesn't even provide the title value. So in way this makes sense to me, that containers take over control of certain props, considering the fact that they only have meaning with in that context. Please let me know your thoughts. Thanks

friendlymahi:
@vladbailescu - any updates on this? Thanks

vladbailescu:
@friendlymahi:

friendlymahi:
@vladbailescu - Understood. So just for clarity, does this mean we can't merge these changes and have to wait until the new design is in place? Please advise as this helps us decide whether we should upgrade to 6.5 sp8 and wait for new core components or create proxy components for the ones that needs style system and use current version of aem and core components that we already have. Appreciate your support in advance. Thanks

vladbailescu:
Since we do not have a clear timeline for this, you should probably add access to styles to your own Sling Models for now. I still think it would be good if you leveraged the newly exposed `com.adobe.cq.com.adobe.cq.wcm.style` APIs.

friendlymahi:
@vladbailescu - Hope you're doing well. We just completed upgrading our AEM instances to v6.5.8 and in the process of validating the same. any update or roadmap on integrating the styles? Thank you !

friendlymahi:
@vladbailescu - Good afternoon/evening to you ! As per your suggestion I incorporated the Style APIs provided by you via uber-6.5.8. To my suprise it worked fine when using the APIs inside Core Components. But upon using the APIs in our app code, the bundle doesn't start. It was prevented from resolving WCM Style API bundle. After some research I found that bundles other than com.adobe or com.aem are not allowed for importing/sharing. So I went ahead and updated the resolver hook configuration to let my bundle import the same. Is this the right way to do it or do you suggest a different way to adress this? Thanks in advance !

vladbailescu:
@friendlymahi , I double-checked and I think I missed to update something in the API controller, making the package visible in uber-jar and not resolved on the instance. My apologies, it will be fixed in 6.5.10.0. Meanwhile, let's try to get this included in Core Components 3.0.0, could you please resolve the conflicts so we can review it one more time?

friendlymahi:
Thanks for confirming the same @vladbailescu. Yes I will be able to resolve the conflicts and send you a note when it's ready for review. Btw anytime line on when 3.0.0 will be out so I can hold on doing custom components for some more time.

friendlymahi:
@vladbailescu Thanks for merging the PR. Just so you are aware after pulling the latest Dev changes into my branch there were some compilation errors due to some plug-ins and eventually some sonar errors for code duplication, etc, both of them are a result of code already existing on Dev branch. In the event you need me to make any code change, please do let me know.

vladbailescu:
@friendlymahi , I fixed a couple of small issues introduced by merging that PR in ab58a3a.
As for a release date, we currently target beginning of June for 3.0.0. However, since this change introduces a dependency to 6.5.8 uber-jar, we'll need a decision from our product management to see if we can drop support for 6.4 or we will need to backport the style system changes there as well.

Issue #2, opened 2016-05-16
cq-wcm-sightly-extension dependency not available 


etherel:
<dependency>
            <groupId>com.adobe.cq.sightly</groupId>
            <artifactId>cq-wcm-sightly-extension</artifactId>
            <version>1.3.4</version>
            <scope>provided</scope>
        </dependency>

In commons-testing/pom.xml can not be resolved

https://repo.adobe.com/nexus/content/repositories/public/com/adobe/cq/

etherel:
Other problems exist with the dependencies meant to be in the repo, the versions of:
- cq-search
- cq-tagging 
are not present.

justinedelson:
@gabrielwalt @raducotescu shouldn't this project depend upon the uberjar?

etherel:
@justinedelson thanks for forwarding this on.  The uber.jar is a dependency in many places e.g. see commons-testing:

     ```
   <dependency>
            <groupId>com.adobe.cq.sightly</groupId>
            <artifactId>cq-wcm-sightly-extension</artifactId>
            <version>1.3.4</version>
            <scope>provided</scope>
        </dependency>
        <dependency>
            <groupId>com.adobe.aem</groupId>
            <artifactId>uber-jar</artifactId>
            <classifier>obfuscated-apis</classifier>
            <scope>provided</scope>
        </dependency>
```

raducotescu:
Those artifacts, which unfortunately don't seem to be deployed to the Adobe public repo, are needed for correctly running the unit tests for the Sightly use objects. Due to how the uber jar was built, API classes seem to have a default constructor generated that does nothing, with the same signature as the original constructor.

justinedelson:
Right, I meant the unobsfucated version.

etherel:
@raducotescu can't you deploy the missing artefacts to the public repo?

Issue #1, opened 2016-04-19
Empty placeholder should display component's title


gabrielwalt:
The empty template of components shouldn't hard-code the text to display, but use the component's jcr:title property instead.

Otherwise, if multiple components extend that component, they all display the same placeholder text:
![ab25fde0-eaab-11e5-8de0-05dfb8c85e2c](https://cloud.githubusercontent.com/assets/130083/14636964/ce305b2a-062e-11e6-8bbb-68000a4455ac.png)

For e.g. this is where the "List" placeholder text is hardcoded:
https://github.com/Adobe-Marketing-Cloud/aem-core-wcm-components/blob/master/list/src/content/jcr_root/apps/core/wcm/components/list/v1/list/templates/emptylist.html#L17

PR #1424, opened 2021-03-03
Experience Fragment enhancement to allow remote offer from Adobe Target

epsilon-asutosh:
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/experiencefragment.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/experiencefragment.html
@@ -20,6 +20,23 @@
      data-sly-set.selector="content.${request.requestPathInfo.selectorString}"
      data-sly-resource="${@path=fragment.localizedFragmentVariationPath, selectors=selector, wcmmode='disabled'}"
      id="${component.id}"
-     class="cmp-experiencefragment cmp-experiencefragment--${fragment.name}">
+     class="cmp-experiencefragment cmp-experiencefragment--${fragment.name} ${properties.enableRemoteOffer ? ' mboxDefault' : ''}">
 </div>
 <sly data-sly-call="${template.placeholder @ isEmpty=!fragment.configured, classAppend='cmp-dd-experiencefragment'}"></sly>
+
+<script data-sly-test="${properties.enableRemoteOffer && properties.id}" data-sly-set.mboxId="${properties.id}">
+    adobe.target.getOffer({
+        mbox: "${mboxId @ context='scriptString'}",
+        success: function (offer) {
+            adobe.target.applyOffer({
+                mbox: "${mboxId @ context='scriptString'}",
+                selector: "#${mboxId @ context='scriptString'}",
+                offer: offer
+            });
+        },
+        error: function (error) {
+            var el = document.getElementById("${mboxId @ context='scriptString'}");
+            el.style.visibility = "visible";
+        }
+    });
+</script>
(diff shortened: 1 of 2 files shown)


content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/experiencefragment.html
```diff
@@ -20,6 +20,23 @@
...
      data-sly-resource="${@path=fragment.localizedFragmentVariationPath, selectors=selector, wcmmode='disabled'}"
      id="${component.id}"
-     class="cmp-experiencefragment cmp-experiencefragment--${fragment.name}">
+     class="cmp-experiencefragment cmp-experiencefragment--${fragment.name} ${properties.enableRemoteOffer ? ' mboxDefault' : ''}">
 </div>
-<sly data-sly-call="${template.placeholder @ isEmpty=!fragment.configured, classAppend='cmp-dd-experiencefragment'}"></sly>
\ No newline at end of file
+<sly data-sly-call="${template.placeholder @ isEmpty=!fragment.configured, classAppend='cmp-dd-experiencefragment'}"></sly>
+
+<script data-sly-test="${properties.enableRemoteOffer && properties.id}" data-sly-set.mboxId="${properties.id}">
```
vladbailescu, line 27: I would prefer if we moved the script to a clientlib instead of inlining it. It could activate on page load and process all elements that have a certain data attribute (for example: `data-enable-remote-offer`).

Also, shouldn't the `mboxId` leverage `component.id` instead o `properties.id`. It's a subtle difference but sometime they could be different.

content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/experiencefragment.html
```diff
@@ -20,6 +20,23 @@
...
+    adobe.target.getOffer({
```
vladbailescu, line 28: I would make sure the global `adobe.target` is available before calling `getOffer`.

vladbailescu (changes requested):

content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/experiencefragment.html
epsilon-asutosh: @vladbailescu I have moved the inline scripts into clientlib and added the data-attibutes.

Explicetely we are reading the mboxId from properties.id because that field is optional and if someone does not want to use the experience targeting offer, they will not configure the field in which case the backend sling model will generate a random id and which we do not want to use for target. So if any customer wants to use the target offer they will need to provide an id explicitly and the same will be used by both component.id and properties.id

content/src/content/jcr_root/apps/core/wcm/components/experiencefragment/v1/experiencefragment/experiencefragment.html
epsilon-asutosh: @vladbailescu added the check to make sure adobe.target is available before it's called.

epsilon-asutosh:
@vladbailescu Can you please review and let me know if all looks good?

epsilon-asutosh:
@vladbailescu Can you please review and let me know if all looks good?

epsilon-asutosh:
@richardhand @bpauli @jckautzmann 
Can you please review this and let me know if all looks good?

epsilon-asutosh:
@vladbailescu Were you able to review this and merge it?

epsilon-asutosh:
@vladbailescu do you mind reviewing the PR?

PR #1425, opened 2021-03-03
nullcheck to avoid 500 if asset does not exist #1414

### LINKED ISSUE #1414: Image Core component throws NPE if fileReference points to a non existing asset
https://github.com/adobe/aem-core-wcm-components/blob/core.wcm.components.reactor-2.15.0/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L875

## Bug Report

**Current Behavior**
Requesting any core/wcm/components/image/v2/image component that has a fileReference which points to a nonexisting image throws a 500. This happens due to the fact that https://github.com/adobe/aem-core-wcm-components/blob/core.wcm.components.reactor-2.15.0/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L875 can be null 

**Expected behavior/code**
Requesting a core/wcm/components/image/v2/image should never throw an NPE internally due to an invalid fileReference. https://github.com/adobe/aem-core-wcm-components/blob/core.wcm.components.reactor-2.15.0/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L875 take invalid fileReference into account

**Environment**
- AEM 6.5 SP7 )
- Core Components version: 2.15.0

Noel-Voelin:
As described in the issue, requesting a core image component that has a file reference which points to a non existinga asset triggers several exceptions and results the request to be responded as a 500 instead of a 404.

I fixed this by simply null checking the imageResource before setting the source to Source.ASSET. The source will remain to be set to Source.NONEXISTING and the therefore correctly be treated as nonexisting (404)

There is a description on how to reproduce this as well as an example har file attached to the ticket. 

I added an unit-test to ensure the HttpServletResponse code to be correct.

Fixes #1414

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -873,7 +873,7 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
             String fileReference = component.getValueMap().get(DownloadResource.PN_REFERENCE, String.class);
             if (StringUtils.isNotEmpty(fileReference)) {
                 imageResource = component.getResourceResolver().getResource(fileReference);
-                if(imageResource != null) {
+                if (imageResource != null) {
                     source = Source.ASSET;
                 }
             } else {
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -873,7 +873,9 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
             String fileReference = component.getValueMap().get(DownloadResource.PN_REFERENCE, String.class);
             if (StringUtils.isNotEmpty(fileReference)) {
                 imageResource = component.getResourceResolver().getResource(fileReference);
-                source = Source.ASSET;
+                if(imageResource != null) {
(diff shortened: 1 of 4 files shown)


Noel-Voelin:
[Adobe CLA_encrypted_.pdf](https://github.com/adobe/aem-core-wcm-components/files/6078631/Adobe.CLA_encrypted_.pdf)

msagolj (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -873,7 +873,9 @@ private long getRequestLastModifiedSuffix(@Nullable String suffix) {
             String fileReference = component.getValueMap().get(DownloadResource.PN_REFERENCE, String.class);
             if (StringUtils.isNotEmpty(fileReference)) {
                 imageResource = component.getResourceResolver().getResource(fileReference);
-                source = Source.ASSET;
+                if(imageResource != null) {
```
vladbailescu, line 876: Please add a space between `if` and parentheses!

vladbailescu (approved):
Thanks for the fix! LGTM except for a small code style issue.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
Noel-Voelin:
```suggestion
                if (imageResource != null) {
```

PR #1431, opened 2021-03-05
[Link Handling][Stage 2] Make link references consistent and flexible #713 - Do not return Link objects for invalid links

vladbailescu:
* Switched LinkHandler to return `Optional<Link>` instead of default invalid link object


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -69,7 +69,7 @@
      *
      * @return Link may be invalid, but is never null
      */
-    public @NotNull Link getLink(@NotNull Resource resource) {
+    public Optional<Link<Page>> getLink(@NotNull Resource resource) {
```
jckautzmann, line 72: it could be that the reference is not a Page: shouldn't <Page> be dropped?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -80,7 +80,7 @@
      *
      * @return Link may be invalid, but is never null
      */
-    public @NotNull Link getLink(@NotNull Resource resource, String linkURLPropertyName) {
+    public Optional<Link<Page>> getLink(@NotNull Resource resource, String linkURLPropertyName) {
```
jckautzmann, line 83: it could be that the reference is not a Page: shouldn't <Page> be dropped?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
vladbailescu: At the moment all these methods defer to `new LinkImpl(url, target, page)` so they can only target a page.

vladbailescu: Removed in dd1fd25

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
vladbailescu: Removed in dd1fd25

jckautzmann (approved):

PR #1435, opened 2021-03-09
DataLayer AssetData backward compatibility #1415

### LINKED ISSUE #1415: DataLayer AssetData backward compatibility
## Bug Report

**Current Behavior**
In Core CIF Components, we implemented the `dataLayer` functionality on top of Core WCM.
CIF Components relies on version `2.9.0` of Core WCM Components.

Lately, we extended some the `dataLayer` for some components to include asset data. And for that we [implemented](https://github.com/adobe/aem-core-cif-components/blob/core-cif-components-reactor-1.8.0/bundles/core/src/main/java/com/adobe/cq/commerce/core/components/internal/datalayer/AssetDataImpl.java) the `AssetData` interface

The problem is that when more recent version on Core WCM Components bundles are present (2.15.0 for example), the serialization of the `dataLayer` for those models fails. Version 2.15.0 is included in the latest AEMaaCS build.

The culprit is the new `getSmartTags` method of `AssetData` that has a `default` behavior of throwing `UnsupportedOperationException`.

**Expected behavior/code**
Default implementation of new methods added on interfaces should not break existing functionality.

**Environment**
- AEM SDK v2021.2.4944.20210221T230729Z-210128
- Core Components version 2.15.0
- JDK11

**Possible Solution**

```
26.02.2021 13:34:28.564 *ERROR* [[0:0:0:0:0:0:0:1] [1614339266359] GET /content/venia/us/en/products/product-page.honora-wide-leg-pants.html HTTP/1.1] com.adobe.cq.commerce.core.components.internal.datalayer.ComponentDataImpl Unable to generate dataLayer JSON string
com.fasterxml.jackson.databind.JsonMappingException: (was java.lang.UnsupportedOperationException) (through reference chain: com.adobe.cq.commerce.core.components.internal.datalayer.ProductDataImpl["xdm:assets"]->com.adobe.cq.commerce.core.components.internal.datalayer.AssetDataImpl[0]->com.adobe.cq.commerce.core.components.internal.datalayer.AssetDataImpl["xdm:smartTags"])
        at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:397) [com.fasterxml.jackson.core.jackson-databind:2.11.3
```

vladbailescu:
* Switched all default methods to no longer throw UnsupportedOperationException
* Removed default methods that could be inherited from super-interfaces
* Methods will return null where the return type is not primitive or annotated as NotNull
* Updated tests
* Cleaned up imports

Fixes #1415


bpauli (commented):
@vladbailescu Maybe we should trace a message in case we return a default value by the default method? Something like `log.trace("return default value " + value)` Customers can than easily  detect new methods in the interface which they might want to implement in their own Sling Model.

cjelger:
@bpauli I'm wondering if this useful at all: I think usually customers would find out about new methods because they have a need to use them, they are probably not interested in anything they don't need for a good reason. Also the default Sling model implementation would usually implement all methods, so this is only useful for `default` methods that are not going to be implemented in the WCM component models.

bpauli:
@cjelger In some cases the default methods might be used by FE devs which might not be aware that there is no meaningful implementation. As the suggested message would only be logged when the method is used it might be a useful way to get some insights.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Component.java
```diff
@@ -72,6 +72,6 @@ default ComponentData getData() {
     @NotNull
     @Override
     default String getExportedType() {
```
msagolj, line 74: can't this method not also be removed because its an override ?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Accordion.java
```diff
@@ -50,7 +50,7 @@ default boolean isSingleExpansion() {
      * @since com.adobe.cq.wcm.core.components.models 12.8.0
      */
     default String[] getExpandedItems() {
-        throw new UnsupportedOperationException();
+        return null;
```
msagolj, line 53: Shouldn't all the default methods that now return `null` instead of an exception have a `@Nullable` Annotation to mark the method as potentially returning a `null` value ?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Component.java
vladbailescu: It overrides the method in `ComponentExporter` which does not provide a `default`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Accordion.java
vladbailescu: I believe most static analysis tools will treat no annotation as `@Nullable` for method return types. I see no point in being explicit here. The contract is more important for `@NotNull`.

msagolj: ok

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/LanguageNavigation.java
```diff
@@ -51,16 +47,7 @@
      * @since com.adobe.cq.wcm.core.components.models 12.2.0
      */
     default List<NavigationItem> getItems() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 50: Shouldn't we return an empty list as a general pattern (as done above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/List.java
```diff
@@ -149,7 +149,7 @@
      */
     @Deprecated
     default Collection<Page> getItems() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 152: shouldn't we return an empty list (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Navigation.java
```diff
@@ -79,17 +75,7 @@
...
-
-    /**
-     * @see ComponentExporter#getExportedType()
-     * @since com.adobe.cq.wcm.core.components.models 12.2.0
-     */
-    @NotNull
-    @Override
-    default String getExportedType() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 78: shouldn't we return an empty list (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/NavigationItem.java
```diff
@@ -60,7 +60,7 @@ default boolean isActive() {
      * @since com.adobe.cq.wcm.core.components.models 12.2.0
      */
     default List<NavigationItem> getChildren() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 63: shouldn't we return an empty list (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -168,7 +169,7 @@ default Calendar getLastModifiedDate() {
      */
     @JsonIgnore
     default String[] getKeywords() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 172: shouldn't we return an empty array (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -248,7 +249,7 @@ default String getBrandSlug() {
      */
     @JsonIgnore
     default String[] getClientLibCategories() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 252: shouldn't we return an empty list (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -261,7 +262,7 @@ default String getBrandSlug() {
      */
     @JsonIgnore
     default String[] getClientLibCategoriesJsBody() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 265: shouldn't we return an empty list (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -274,7 +275,7 @@ default String getBrandSlug() {
      */
     @JsonIgnore
     default String[] getClientLibCategoriesJsHead() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 278: shouldn't we return an empty list (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/SocialMediaHelper.java
```diff
@@ -138,16 +135,7 @@ default boolean hasPinterestSharing() {
      * @since com.adobe.cq.wcm.core.components.models 11.0.0; marked <code>default</code> in 12.1.0
      */
     default Map<String, String> getMetadata() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 138: shouldn't we return an empty map (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Teaser.java
```diff
@@ -143,7 +142,7 @@ default boolean isActionsEnabled() {
      * @since com.adobe.cq.wcm.core.components.models 12.4.0
      */
     default List<ListItem> getActions() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 145: shouldn't we return an empty list (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/contentfragment/DAMContentFragment.java
```diff
@@ -249,7 +250,7 @@ default String getType() {
     @Nullable
     @JsonIgnore
     default List<DAMContentElement> getElements() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 253: shouldn't we return an empty list (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/contentfragment/DAMContentFragment.java
```diff
@@ -290,7 +291,7 @@ default String getType() {
     @Nullable
     @JsonIgnore
     default List<Resource> getAssociatedContent() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 294: shouldn't we return an empty list (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/AssetData.java
```diff
@@ -75,7 +75,7 @@ default String getFormat() {
      */
     @JsonProperty("xdm:tags")
     default String[] getTags() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 78: shouldn't we return an empty array (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/AssetData.java
```diff
@@ -100,7 +100,7 @@ default Date getLastModifiedDate() {
      */
     @JsonProperty("xdm:smartTags")
     default Map<String, Object> getSmartTags() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 103: shouldn't we return an empty list (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/ContainerData.java
```diff
@@ -30,6 +30,6 @@
      * @since com.adobe.cq.wcm.core.components.models.datalayer 1.0.0
      */
     default String[] getShownItems() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 33: shouldn't we return an empty array (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/ContentFragmentData.java
```diff
@@ -28,7 +28,7 @@
      */
     @JsonProperty("elements")
     default ElementData[] getElementsData() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 31: shouldn't we return an empty array (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/datalayer/PageData.java
```diff
@@ -33,7 +33,7 @@
      */
     @JsonProperty("xdm:tags")
     default String[] getTags() {
-        throw new UnsupportedOperationException();
+        return null;
```
jckautzmann, line 36: shouldn't we return an empty array (see comment above)?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/LanguageNavigation.java
vladbailescu: I opted to return `null` for all methods (expect primitives and those annotated with `@NotNull`) as these properties will then be omitted during JSON serialization and I find that nicer than: `items:[]`

bpauli (approved):

jckautzmann (approved):

PR #1444, opened 2021-03-11
[Navigation][LinkHandler][Stage3][ID] Create Navigation v2 and integrate the new link handler feature

jckautzmann:
* Creates a new version Navigation v2 that integrates the new link handler feature
* Picks und updates changes from PR #862.
* Fixes #713


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/NavigationImpl.java
```diff
@@ -0,0 +1,45 @@
...
+import com.adobe.cq.wcm.core.components.models.Navigation;
+import com.adobe.cq.wcm.core.components.models.NavigationItem;
+import com.day.cq.wcm.api.Page;
+import com.day.cq.wcm.api.components.Component;
+
+@Model(adaptables = SlingHttpServletRequest.class,
+        adapters = {Navigation.class, ComponentExporter.class},
+        resourceType = {NavigationImpl.RESOURCE_TYPE})
+@Exporter(name = ExporterConstants.SLING_MODEL_EXPORTER_NAME , extensions = ExporterConstants.SLING_MODEL_EXTENSION)
+public class NavigationImpl extends com.adobe.cq.wcm.core.components.internal.models.v1.NavigationImpl {
```
abidaJustice, line 37: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names shouldn’t shadow simple name of superclass (NM_SAME_SIMPLE_NAME_AS_SUPERCLASS)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-shouldn-t-shadow-simple-name-of-superclass-nm-same-simple-name-as-superclass).

vladbailescu (approved):

PR #1462, opened 2021-03-25
CQ-4289582 [Test] Convert Hobbes Tests to JUnit5/Selenium

absarasw:
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


vladbailescu (changes requested):
I generally like the way things are setup but I do see some inconsistencies that should be addressed:

1. For each component we should have:
    * A model class that defines the output of the component and provides getters and/or asserts for output sections/fields (ex. `components/button/Button.java`). For components that have big differences between versions we could have version-specific models (ex. `components/form/button/FormButtonV1.java` and `FormButtonV2.java`)
    * A model class for the edit/configuration dialog that provides getters and setters for configuring the component (ex. `components/button/ButtonConfigurationDialog.java`) - this is inconsistent across components at the moment, we should settle on a common pattern. Could be versioned if there are big differences between versions.
    * An IT class for each component version that we test, which provides the testing logic: drag component on the page and get hold of the component model and configuration dialog, set field values (and possibly assert expected behaviour in the dialog), assert expected elements/behaviour for the rendered component. (ex. `tests/button/ButtonV1IT.java`)
2. Where possible, a higher version of a model/test class should inherit from the lower version. ITs should run compatible tests from the lower version instead of redefining (copy-pasting) the same code.
3. It's best if we extract all constants/strings and name them in all-caps, following the same convention across classes: CSS selectors (`SELECTOR_BUTTON_TITLE`), field  names (`FIELD_BUTTON_TITLE`) etc. It's also best if we don't redefine common constants across classes.
4. We should strive to not use Granite UI or Coral internal constants (such as `coral-tab-label` and so on) as these might be subject to change when an UI redesign/shift takes place. I know the Hobbes tests are using them but now it's a good moment to drop them.

absarasw:
@vladbailescu ,

> 1. For each component we should have:
>    
>    * A model class that defines the output of the component and provides getters and/or asserts for output sections/fields (ex. `components/button/Button.java`). For components that have big differences between versions we could have version-specific models (ex. `components/form/button/FormButtonV1.java` and `FormButtonV2.java`)

Yes, we are following it and I have updated the PR accordingly.
>    * A model class for the edit/configuration dialog that provides getters and setters for configuring the component (ex. `components/button/ButtonConfigurationDialog.java`) - this is inconsistent across components at the moment, we should settle on a common pattern. Could be versioned if there are big differences between versions.

It's difficult to combine for all the components. But we can do this for the components which are similar like Button and FormButton. Like I have used the same configuration dialog for the mentioned components.
>    * An IT class for each component version that we test, which provides the testing logic: drag component on the page and get hold of the component model and configuration dialog, set field values (and possibly assert expected behaviour in the dialog), assert expected elements/behaviour for the rendered component. (ex. `tests/button/ButtonV1IT.java`)

I'm refining the test classes to only provide the testing logics.
> 2. Where possible, a higher version of a model/test class should inherit from the lower version. ITs should run compatible tests from the lower version instead of redefining (copy-pasting) the same code.

Updated the PR with the same.
> 3. It's best if we extract all constants/strings and name them in all-caps, following the same convention across classes: CSS selectors (`SELECTOR_BUTTON_TITLE`), field  names (`FIELD_BUTTON_TITLE`) etc. It's also best if we don't redefine common constants across classes.

Should we add component specific constants as well in a common class? Like if some selectors are specific to Accordion component but not relevant to other components.
> 4. We should strive to not use Granite UI or Coral internal constants (such as `coral-tab-label` and so on) as these might be subject to change when an UI redesign/shift takes place. I know the Hobbes tests are using them but now it's a good moment to drop them.

Yes, Updated the PR for the same.

vladbailescu:
@absarasw 
>> A model class for the edit/configuration dialog that provides getters and setters for configuring the component (ex. components/button/ButtonConfigurationDialog.java) - this is inconsistent across components at the moment, we should settle on a common pattern. Could be versioned if there are big differences between versions.
>
> It's difficult to combine for all the components. But we can do this for the components which are similar like Button and FormButton. Like I have used the same configuration dialog for the mentioned components.

I did not mean to combine these across classes but rather to make sure we have a common pattern. For example, `Button` has a `ButtonConfigurationDialog` as a standalone class file while `Accordion` has an inline `EditDialog` class.

vladbailescu:
@absarasw 

>> * It's best if we extract all constants/strings and name them in all-caps, following the same convention across classes: CSS selectors (SELECTOR_BUTTON_TITLE), field names (FIELD_BUTTON_TITLE) etc. It's also best if we don't redefine common constants across classes.
>
> Should we add component specific constants as well in a common class? Like if some selectors are specific to Accordion component but not relevant to other components.

Component-specific constants should sit within the component IT model (or the configuration dialog model). Common stuff should be extracted to avoid duplication. We will still need to use the same naming patterns.

PR #1478, opened 2021-03-30
[Accessibility][Image v2] Always provide the ALT attribute, even when empty (#1467)

### LINKED ISSUE #1467: [Accessibility][Image] Always provide the ALT attribute, even when empty
From an accessibility perspective, the IMG element should always have an ALT attribute. Typically, even when the author configures the image to be decorative, the ALT attribute should remain present, even if it's empty.

To learn more about it, see for example: https://davidwalsh.name/accessibility-tip-empty-alt-attributes

The current behavior is that if the ALT attribute is empty, it will be entirely removed.

It would be expected that the ALT attribute never gets removed, even when it's kept empty.

Possible solution: on [line 42 of the image.html template](https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/image.html#L42), write the ALT attribute as follows: `alt="${image.alt || true}"`

As this bug fix shouldn't affect compatibility, it could safely be applied to all versions of the image component: v2, and v3 if it exists.

comanV:
- always provide ALT attribute for image, in Image Component v2
Fixes #1467

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/image.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/image.html
@@ -23,7 +23,7 @@
      data-cmp-src="${image.srcUriTemplate ? image.srcUriTemplate : image.src}"
      data-cmp-widths="${image.widths}"
      data-cmp-dmimage="${image.dmImage}"
-     data-cmp-smartcroprendition="${image.smartCropRendition}"     
+     data-cmp-smartcroprendition="${image.smartCropRendition}"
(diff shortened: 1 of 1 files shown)


vladbailescu (approved):

vladbailescu:
* always provide ALT attribute for image, in Image Component v2

Fixes #1467

vladbailescu:
We should carry over this fix to v3 as well in https://github.com/adobe/aem-core-wcm-components/pull/1423

PR #1479, opened 2021-03-30
[Container] Background image with whitespace do not work

### LINKED ISSUE #1468: [Container] Background image with whitespace do not work
### Expected Behaviour
Image with whitespace in it's path should work
### Actual Behaviour
Image with whitestaces in it's path do not work
#### Steps to Reproduce
1. Create a container
2. Set background image using file that have 'space' in it's name
3. Verify that no background image is displayed
#### Platform and Version
Core components 2.13.2

msagolj:
| Fixed Issues?            | Fixes #1468
| Documentation Provided   | no

Images with whitespace in the name should work. Can be reproduced by uploading an asset with spaces as an background image. According to [W3C Standard](https://www.w3.org/TR/2011/REC-CSS2-20110607/syndata.html#uri) the URI values should either be in single or double quotes or special characters like whitespace should be escaped.

Can't use quotes, since this gets escaped in the output by the `@context` setting:

`style="${container.backgroundStyle @ context='styleString'}"`

in [responsive grid](https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/container/v1/container/responsiveGrid.html#L19) and [simple](https://github.com/adobe/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/container/v1/container/simple.html#L21) container as `\27`:

`url(\27\2f content\2f dam\2fname with space.jpg\27)`

Can't use [URL.Encoder](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URLEncoder.html) to encode special characters either as as it encodes whitespace as `+` instead of `%20`. For now  its implemented as a simple search replace  which seems a bit hacky, any suggestions welcome.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractContainerImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractContainerImpl.java
@@ -196,7 +196,11 @@ public abstract class AbstractContainerImpl extends AbstractComponentImpl implem
     public String getBackgroundStyle() {
         if (this.backgroundStyle == null) {
             StringBuilder styleBuilder = new StringBuilder();
-            getBackgroundImage().ifPresent(image -> styleBuilder.append("background-image:url(").append(image).append(");background-size:cover;background-repeat:no-repeat;"));
+            getBackgroundImage().ifPresent(image -> {
+                styleBuilder.append("background-image:url(")
+                    .append(image.replace(" ","%20"))
+                    .append(");background-size:cover;background-repeat:no-repeat;");
+            });
             getBackgroundColor().ifPresent(color -> styleBuilder.append("background-color:").append(color).append(";"));
             this.backgroundStyle = styleBuilder.toString();
         }


vladbailescu (approved):
Should we also add a small unit test for this?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractContainerImpl.java
```diff
@@ -196,7 +196,11 @@
     public String getBackgroundStyle() {
         if (this.backgroundStyle == null) {
             StringBuilder styleBuilder = new StringBuilder();
-            getBackgroundImage().ifPresent(image -> styleBuilder.append("background-image:url(").append(image).append(");background-size:cover;background-repeat:no-repeat;"));
+            getBackgroundImage().ifPresent(image -> {
+                styleBuilder.append("background-image:url(")
+                    .append(image.replace(" ","%20"))
```
kwin, line 201: why not using `java.net.URI` (as outlined in https://stackoverflow.com/a/65805013) for doing the encoding. That would also take care of other characters not valid in a URI.

kwin: resourceResolver.map() would probably even better as that takes into account both mapping and does URL escaping.

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/AbstractContainerImpl.java
msagolj: @kwin both work so thank you for your input :)

msagolj: .. and @vladbailescu even suggested a 3rd one :) let me check this one as well.

vladbailescu: Proposed a new approach in https://github.com/adobe/aem-core-wcm-components/pull/1493

vladbailescu:
Proposed a new approach in https://github.com/adobe/aem-core-wcm-components/pull/1493

PR #1487, opened 2021-03-31
[Accessibility][Search] Search component keyboard trap and tab navigation 

### LINKED ISSUE #1437: [Accessibility][Search] Search component keyboard trap and tab navigation
## Bug Report

**Current Behavior**
In the search component, after text is entered in the search input field and/or the search results have been populated, the clear input button is revealed but cannot be reached via tab navigation due to a keyboard trap. It can be reached and activated via screen readers but for WCAG compliance it should be usable via tab navigation. After search results have been populated, this keyboard trap persists even if you clear the search input field.

**Expected behavior/code**
The tab button should allow you to proceed to the clear input button and below and the input field should not trap the user in the input field either when the field is populated or empty. In addition to preventing access to the clear input field button this also prevents tab navigation anywhere below the search component.

**Environment**
- AEM version and patch level: 6.5.6.0 & AEM as a cloud service
- Core Components version: 2.15.0
- JRE version: Java(TM) SE Runtime Environment (build 1.8.0_241-b07)

**Possible Solution**
Disabled tab keystroke checking on the input field 

Issue appears to be related to the keydown event handlers around lines 209-213 found at: /apps/core/wcm/components/search/v1/search/clientlibs/site/js/search.js

Screen Shot 2021-03-09 at 4 18 15 PM

You may also wish to account for shift+tab preventing navigating the page in reverse.

comanV:
Fixes #1437

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/search/v1/search/clientlibs/site/js/search.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/search/v1/search/clientlibs/site/js/search.js
@@ -207,11 +207,6 @@
         var self = this;
 
         switch (event.keyCode) {
-            case keyCodes.TAB:
-                if (self._resultsOpen()) {
-                    event.preventDefault();
-                }
-                break;
             case keyCodes.ENTER:
                 event.preventDefault();
                 if (self._resultsOpen()) {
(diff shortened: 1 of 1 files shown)


content/src/content/jcr_root/apps/core/wcm/components/search/v1/search/clientlibs/site/js/search.js
```diff
@@ -207,11 +207,6 @@
         var self = this;
 
         switch (event.keyCode) {
-            case keyCodes.TAB:
```
vladbailescu, line 210: This seems to have been intentionally removed before. We should make sure it does not interfere with results navigation (up-down arrows) and tab selection and up-down results selection look and behave consistently.

vladbailescu (changes requested):

vladbailescu (approved):

PR #1491, opened 2021-03-31
[ID] Create contentfragmentlist v2

### LINKED ISSUE #1254: Do not build `Component` model in HTL if current component implements `Component`
When a component model implements the `Component` interface, then the component ID should be available directly from the component's model. There is no need to create (within the component HTL) an additional sling model for `Component`. Doing so causes extra unnecessary work, and leaves open the possibility that the component model may override the `#getID()` method, and that override would not be honoured by the HTL unless the component model also marked `Component` interface in the model adapters.   

Remove the `data-sly-use.component="..."` from component HTL where there already exists a model that implements the `Component` interface.

e.g.:
```HTML
<nav data-sly-use.breadcrumb="com.adobe.cq.wcm.core.components.models.Breadcrumb"
     data-sly-use.component="com.adobe.cq.wcm.core.components.models.Component"
     data-sly-use.template="core/wcm/components/commons/v1/templates.html"
     id="${component.id}"  ...>
    ...
</nav>
```
vs:
```HTML
<nav data-sly-use.breadcrumb="com.adobe.cq.wcm.core.components.models.Breadcrumb"
     data-sly-use.template="core/wcm/components/commons/v1/templates.html"
     id="${breadcrumb.id}"  ...>
    ...
</nav>
```

**Are there alternatives?**
Resolving #1061 would achieve the same outcome by ensuring that the `data-sly-use.component="com.adobe.cq.wcm.core.components.models.Component"` resolves the same component model, and thus any override to the ID would be honoured; However, it does not resolve the unnecessary overhead of constructing the model twice under different interfaces.

I believe that both the changes suggested in this ticket, as well as those in #1061 should be implemented so as to eliminate that overhead, while ensuring that regardless of how you get the component (adapting to the `Component` model interface, or the specific component interface that extends the `Component` interface), the results are the same. 

For clarity, this should be the promise that is made (not real java)
```JAVA
String componentID = getModel(resource, Component.class).getID();
String breadcrumbID =  ((Component) getModel(resource, Breadcrumb.class)).getID();
Assertions.assertEquals(componentID, breadcrumbID);

// or more generally, being equivalent, not necessarily same.
Assertions.assertEquals(getModel(resource, Component.class), ((Component) getModel(resource, Breadcrumb.class)));
```

bpauli:
Create contentfragmentlist v2 and remove usage of Component model

| Fixed Issues?            | Fixes #1254
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


examples/ui.apps/src/content/jcr_root/apps/core-components-examples/components/contentfragmentlist/.content.xml
```diff
@@ -19,5 +19,5 @@
     jcr:description="Displays a list of Content Fragments"
     jcr:primaryType="cq:Component"
     jcr:title="Content Fragment List "
-    sling:resourceSuperType="core/wcm/components/contentfragmentlist/v1/contentfragmentlist"
+    sling:resourceSuperType="core/wcm/components/contentfragmentlist/v2/contentfragmentlist"
```
vladbailescu, line 22: We should probably also update the Embed page texts and links in Library.

PR #1496, opened 2021-04-01
issue/1488 add check for large images to return smaller rendition

### LINKED ISSUE #1488: [Image] Large "Portrait" images throws "Cannot process rendition original due to size" when loading
## Bug Report
**Steps to reproduce**
Reproducible on a clean environment with latest core components installed. 

1. Create a page with the image core component
2. Drag and drop an image with size 4000x5000

**Current Behavior**
Image is loaded with the original rendition which is larger than 3840 and (with default settings) will cause the "Cannot process rendition XXX due to size" error. 

**Expected behavior/code**
Image is loaded with (by default) the 1280 rendition

**Environment**
- Adobe Experience Manager (6.5.5.0)
- Core Components version 2.13.2 & 2.15.3
- Java(TM) SE Runtime Environment (build 1.8.0_221-amd64)

**Possible Solutions**
The `getBestRendition` method in `AdaptiveImageServlet` shouldn't fallback to the "original" rendition if width is larger than the `DEFAULT_MAX_SIZE` AND height is larger than width. In our case we would like to return a 1920 rendition. 

error-image

Semperdecus:
Fixes #1488

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -103,7 +103,7 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
     private static final String SELECTOR_WIDTH_KEY = "width";
     private int defaultResizeWidth;
     private int maxInputWidth;
-    
+
     private AdaptiveImageServletMetrics metrics;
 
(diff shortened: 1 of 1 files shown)


msagolj:
@Semperdecus added link to issue in PR description so they get linked, also see my comment in the issue.

vladbailescu:
As discussed in #1488, this is not the preferred solution as it assumes the second largest rendition is suitable as a replacement. It could be too large itself, it could lead to unwanted upscaling and so on, so we're not sure it works for all projects. Closing this for now.

PR #1497, opened 2021-04-01
Create latest version ImplementationPicker

### LINKED ISSUE #546: [Architecture] Use newest model implementation by default
## Feature Request

If I want to reuse a model from core wcm components in another context (i.e. not in a resource using or inheriting from core wcm component's resource types) Sling Models will pick the first implementation by default.
I.e. if you use the following code snippet in HTL
```
<sly data-sly-use.page="com.adobe.cq.wcm.core.components.models.Page"/>
<sly data-sly-test="${page.clientLibCategoriesJsHead}"
         data-sly-call="${clientLib.js @ categories=page.clientLibJsHead}">
</sly>
```
which is not bound to resource type `core/wcm/components/page/v2/page` by default the implementation from `com.adobe.cq.wcm.core.components.internal.models.v1.PageImpl` is picked instead of `com.adobe.cq.wcm.core.components.internal.models.v2.PageImpl`

In this particular case this leads to an `UnsupportedOperationException` as the v1 PageImpl does not override the default impl of `getClientLibCategoriesJsHead`

I would be nice if the packages are ordered in a way so that always the newest (not the oldest) version is being picked by default by the `org.apache.sling.models.impl.FirstImplementationPicker`

**Are there alternatives?**
You cannot reference the model with the implementation name i.e. `com.adobe.cq.wcm.core.components.internal.models.v2.PageImpl` as that is not exported.

bpauli:
- add implementation picker which returns the latest core component model if no resource type is matching
- add unit test

| Fixed Issues?            | Fixes #546 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
@@ -32,7 +32,7 @@ import org.osgi.service.component.propertytypes.ServiceRanking;
 @ServiceRanking(value = 1)// this must come after ResourceTypeBasedResourcePicker
 public class LatestVersionImplementationPicker implements ImplementationPicker {
 
-    private static final Pattern INTERNAL_MODEL_PATTERN = Pattern.compile("com\\.adobe\\.cq\\.wcm\\.core\\.components\\.internal\\.models\\.v(\\d+)\\.");
+    private static final Pattern INTERNAL_MODEL_PATTERN = Pattern.compile("^com\\.adobe\\.cq\\.wcm\\.core\\.components\\.internal\\.models\\.v(\\d+)\\.\\S*$");
     private static final String CORE_COMPONENTS_MODELS = "com.adobe.cq.wcm.core.components.models";
 
     @Override
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
@@ -45,7 +45,7 @@ public class LatestVersionImplementationPicker implements ImplementationPicker {
                 if (m1.matches() && m2.matches()) {
                     return Integer.parseInt(m2.group(1)) - Integer.parseInt(m1.group(1));
                 } else {
-                    return (m1.find() ? 1 : (m2.find() ? -1 : 0));
+                    return (m1.matches() ? 1 : (m2.matches() ? -1 : 0));
                 }
             }).orElse(implementationsTypes[0]);
         }
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
@@ -16,7 +16,6 @@
 package com.adobe.cq.wcm.core.components.internal.services;
 
 import java.util.Arrays;
-import java.util.Comparator;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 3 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
```diff
@@ -0,0 +1,56 @@
...
+@Component(property=Constants.SERVICE_RANKING+":Integer=1") // this must come after ResourceTypeBasedResourcePicker
+public class LatestVersionImplementationPicker implements ImplementationPicker {
+
+    private static final Pattern pattern = Pattern.compile("\\.v(\\d+)\\.");
+
+    @Override
+    public Class<?> pick(Class<?> adapterType, Class<?>[] implementationsTypes, Object adaptable) {
+        // make sure picker is only responsible for core component models
+        if (adapterType.getPackage().getName().equals("com.adobe.cq.wcm.core.components.models")) {
+            Arrays.sort(implementationsTypes, new Comparator<Class<?>>() {
```
kwin, line 41: Why not using Java 8 stream with https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#maxBy-java.util.Comparator-. Should be less effort than sorting the array and afterwards just returning the first item.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.sling.models.spi.ImplementationPicker;
+import org.osgi.framework.Constants;
+import org.osgi.service.component.annotations.Component;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Picks latest model version of a core component model
+ */
+@Component(property=Constants.SERVICE_RANKING+":Integer=1") // this must come after ResourceTypeBasedResourcePicker
```
kwin, line 32: Using the annotation http://docs.osgi.org/javadoc/osgi.cmpn/7.0.0/org/osgi/service/component/propertytypes/ServiceRanking.html is much easier to read and write

kwin:
This would fix #546.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
```diff
@@ -21,15 +21,14 @@
...
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.osgi.service.component.propertytypes.ServiceRanking;
 
 /**
  * Picks latest model version of a core component model
  */
-@Component(property=Constants.SERVICE_RANKING+":Integer=1") // this must come after ResourceTypeBasedResourcePicker
+@Component
+@ServiceRanking(value = 1)// this must come after ResourceTypeBasedResourcePicker
```
kwin, line 31: This will only work together with Sling Model Impl 1.4.12 and later (https://github.com/apache/sling-org-apache-sling-models-impl/commit/89ae8beba7f6bd03e3af13c45124cbe62f122b3b), in older versions the ResourceTypeBasedResourcePicker was registered at `Integer.MAX_VALUE - 1`. Not sure if all supported AEM versions already come with that version.

bpauli: Since v2.11.0 the core components depend on 1.4.12 of the org.apache.sling.models.impl (see system requirements in README)

bpauli:
@kwin Thank you for the review! Please check latest commits for a final review. Thanks!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
```diff
@@ -0,0 +1,51 @@
...
+    public Class<?> pick(Class<?> adapterType, Class<?>[] implementationsTypes, Object adaptable) {
+        // make sure picker is only responsible for core component models
+        if (adapterType.getPackage().getName().equals("com.adobe.cq.wcm.core.components.models")) {
+            return Arrays.stream(implementationsTypes).min((Class<?> o1, Class<?> o2) -> {
+                Matcher m1 = pattern.matcher(o1.getName());
+                Matcher m2 = pattern.matcher(o2.getName());
+                if (m1.find() && m2.find()) {
+                    return Integer.parseInt(m2.group(1)) - Integer.parseInt(m1.group(1));
+                }
+                return 0;
```
kwin, line 46: I think a deliberate order for external implementations not only relying on the version string would be good as well. So all implementations shipped with Core WCM should be sorted by the version number in it but the ones outside package `com.adobe.cq.wcm.core.components.internal` should be preferred (independent of any version number in the class name). WDYT?

bpauli: Indeed, that makes sense. I changed the logic of the Implementation Picker in the last commit and added a unit test for this case.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
```diff
@@ -0,0 +1,54 @@
...
+public class LatestVersionImplementationPicker implements ImplementationPicker {
+
+    private static final Pattern INTERNAL_MODEL_PATTERN = Pattern.compile("com\\.adobe\\.cq\\.wcm\\.core\\.components\\.internal\\.models\\.v(\\d+)\\.");
+    private static final String CORE_COMPONENTS_MODELS = "com.adobe.cq.wcm.core.components.models";
+
+    @Override
+    public Class<?> pick(Class<?> adapterType, Class<?>[] implementationsTypes, Object adaptable) {
+        // make sure picker is only responsible for core component models
+        if (adapterType.getPackage().getName().equals(CORE_COMPONENTS_MODELS)) {
+            return Arrays.stream(implementationsTypes).min((Class<?> o1, Class<?> o2) -> {
```
vladbailescu, line 42: I don't think you should use `min` here, this causes the opposite effect:

```
Error:    LatestVersionImplementationPickerTest.testPickExternalModel:60 expected: <com.sun.proxy.$Proxy72> but was: <com.adobe.cq.wcm.core.components.internal.models.v1.PageImpl>
Error:    LatestVersionImplementationPickerTest.testPickLatestModelVersion:47 expected: <com.adobe.cq.wcm.core.components.internal.models.v2.PageImpl> but was: <com.adobe.cq.wcm.core.components.internal.models.v1.PageImpl>
```

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/LatestVersionImplementationPicker.java
bpauli: Sorry, there was an issue with the regex. It has been fixed with the latest commit.

vladbailescu (approved):

PR #1503, opened 2021-04-07
Inconsistent component ID in model.json output

### LINKED ISSUE #1502: Inconsistent component ID in model.json output
It is a feature of the component ID feature that the components inherited from the template structure have the same ID throughout all pages resulting from that template. This works as expected for the HTML output, but doesn't for the JSON output.

As an example, look at the language navigation component of following pages that has the same `languagenavigation-ac5c6a67dc` ID on both pages:
* https://wknd.site/content/wknd/us/en.html
* https://wknd.site/content/wknd/us/en/adventures.html

But now look at the corresponding JSON output, and see how this container has each time a different ID:
* https://wknd.site/content/wknd/us/en.model.json has ID: `languagenavigation-bdcfb864f8`
* https://wknd.site/content/wknd/us/en/adventures.model.json has ID: `languagenavigation-d1527c62a0`

It would be expected that the JSON output always consistently has the same `languagenavigation-ac5c6a67dc` ID than on the HTML output.

vladbailescu:
* Added passing of parent resource to the underlying components when working with Experience Fragments (XFs). This will be used when computing the ids of the components within the XF.
* Added test for XF with items.
* Removed unused dummy models (serialization is done with actual Core Components models when properly setup).
* Updated models that had the request injected so they pick it up from the base class.

Fixes #1502


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
```diff
@@ -70,21 +75,39 @@ public static boolean isDataLayerEnabled(@NotNull final Resource resource) {
...
+    }
+
     /**
      * Get the ID property value if set (using {@link #getPropertyId(Resource)},
-     * otherwise generate a new ID (using {@link #generateId(Resource, Page, ComponentContext)}.
+     * otherwise generate a new ID (using {@link #generateId(Resource, Page, String, ComponentContext)}.
      *
      * @param resource The resource for which to get or generate an ID.
      * @param currentPage The current request page.
+     * @param resourceCallerPath The resource caller path(s).
```
jckautzmann, line 100: Is it possible to have multiple paths? Otherwise it should be: The resource caller path.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
```diff
@@ -168,6 +194,9 @@ private static String generateId(@NotNull final Resource resource,
      */
     @NotNull
     public static String generateId(@NotNull final String prefix, @NotNull final String path) {
+        if ("search".equals(prefix)) {
```
jckautzmann, line 197: Is this logging needed?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/ContentFragmentUtils.java
```diff
@@ -245,19 +252,32 @@ public static String getGridResourceType(Resource fragmentResource) {
...
+        @NotNull final SlingHttpServletRequest slingHttpServletRequest,
+        @NotNull final Resource callerResource) {
         final LinkedHashMap<String, ComponentExporter> componentExporterMap = new LinkedHashMap<>();
 
+        SlingHttpServletRequest wrappedSlingHttpServletRequest = new SlingHttpServletRequestWrapper(slingHttpServletRequest) {
+
+            @Override
+            public Object getAttribute(String name) {
+                if (ATTR_RESOURCE_CALLER_PATH.equals(name)) {
+                    String resourceCallerPath = (String)super.getAttribute(ATTR_RESOURCE_CALLER_PATH);
```
jckautzmann, line 271: Why is this line needed? In case this request attribute is set in a custom component?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/ContentFragmentUtils.java
```diff
@@ -245,19 +252,32 @@ public static String getGridResourceType(Resource fragmentResource) {
      * @param resourceIterator Iterator of resources for which to get the component exporters.
      * @param modelFactory Model factory service.
      * @param slingHttpServletRequest Current request.
+     * @param callerResource The caller (XF) resource.
```
jckautzmann, line 255: I would be more specific. e.g.:
The caller resource, i.e. the page or template resource referencing the experience fragment or the content fragment.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
vladbailescu: Good catch? It was a debugging left-over, removed it in 4d14728.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
vladbailescu: Updated in 4d14728.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/ContentFragmentUtils.java
vladbailescu: Updated in 4d14728.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/ContentFragmentUtils.java
vladbailescu: For HTML rendering, for XFs we only use the root-most resource path (obtained through the `componentContext`) concatenated with the path of the component inside the leaf-most XF (in case of nesting). I added a comment bellow explaining why we return the existing caller path in 4d14728.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/ContentFragmentUtils.java
```diff
@@ -245,19 +252,36 @@ public static String getGridResourceType(Resource fragmentResource) {
...
     public static LinkedHashMap<String, ComponentExporter> getComponentExporters(
         @NotNull final Iterator<Resource> resourceIterator,
         @NotNull final ModelFactory modelFactory,
-        @NotNull final SlingHttpServletRequest slingHttpServletRequest) {
+        @NotNull final SlingHttpServletRequest slingHttpServletRequest,
+        @NotNull final Resource callerResource) {
         final LinkedHashMap<String, ComponentExporter> componentExporterMap = new LinkedHashMap<>();
 
+        SlingHttpServletRequest wrappedSlingHttpServletRequest = new SlingHttpServletRequestWrapper(slingHttpServletRequest) {
+
```
jckautzmann, line 267: Instead of ovewriting the getAttribute method, wouldn't it be easier to set the attribute in the request? E.g.:
```
        String resourceCallerPath = (String) slingHttpServletRequest.getAttribute(ATTR_RESOURCE_CALLER_PATH);
        SlingHttpServletRequest wrappedSlingHttpServletRequest = new SlingHttpServletRequestWrapper(slingHttpServletRequest);
        if (StringUtils.isEmpty(resourceCallerPath)) {
            wrappedSlingHttpServletRequest.setAttribute(ATTR_RESOURCE_CALLER_PATH, callerResource.getPath());
        }
```

bpauli (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/ContentFragmentUtils.java
vladbailescu: The advantage of using a request wrapper and overriding `getAttribute` is that the attribute will get automatically removed when the original request is unwrapped, thus no need to clean up (which is something we should always do).

jckautzmann (approved):

Issue #1469, opened 2021-03-26
[Accessibility][Container] Allow to define ARIA labels and role attributes on the container


gabrielwalt:
As an author, I want to be able to add ARIA labels on the container, so that my site structure has accessible landmarks.

To avoid requiring a new version, insert the "Label" field to the current tab of the edit dialog.

gabrielwalt:
We actually also need a "Role" text field in the dialog that outputs a `role` attribute on the container element.

comanV:
> We actually also need a "Role" text field in the dialog that outputs a `role` attribute on the container element.

I opened PR for the enhancement: https://github.com/adobe/aem-core-wcm-components/pull/1507

gabrielwalt:
Thanks @comanV, your PR looks good to me, just very quickly looking at it!

PR #1507, opened 2021-04-09
[Accessibility][Container] Allow authors to define role attribute on the container

comanV:
Fixes the latest requirement from #1469

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/LayoutContainerImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/LayoutContainerImpl.java
@@ -62,6 +62,13 @@ public class LayoutContainerImpl extends AbstractContainerImpl implements Layout
     private String accessibilityLabel;
 
     /**
+     * The role attribute.
+     */
+    @ValueMapValue(injectionStrategy = InjectionStrategy.OPTIONAL)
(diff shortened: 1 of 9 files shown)


vladbailescu (changes requested):
Please also add a test for the `accessibilityLabel` and `roleAttribute`, including JSON export.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/LayoutContainerImpl.java
```diff
@@ -61,6 +61,13 @@
     @Nullable
     private String accessibilityLabel;
```
vladbailescu, line 62: The new properties should be documented in README.md

content/src/content/jcr_root/apps/core/wcm/components/container/v1/container/README.md
```diff
@@ -46,6 +46,9 @@ It is also possible to define the allowed components for the Container.
 ### Edit Dialog Properties
 The following properties are written to JCR for this Container component and are expected to be available as `Resource` properties:
 
+1. `./accessibilityLabel` - defines an accessibility label for the container.
```
vladbailescu, line 49: I'd move this to a #### Accessibility section bellow (after Common Properties).

vladbailescu (approved):

PR #1515, opened 2021-04-15
[Title][Accessibility] Possibility of adding description related to link in Title component

### LINKED ISSUE #668: [Title][Accessibility] Possibility of adding description related to link in Title component
## Feature Request

**Current behavior**
As an author I would like to add description (like Alt text) to link in Title component. When I try to configure Title component there is only field to fill link (without description for this link).

**Expected behaviour**
Author can add purpose of the link (for instance by some Alt text field).

**Are there alternatives?**
Raised issue is related to 2.4.4 point from WCAG 2.0 accessibility standard. Is there any buisness goal that this component doesn't require 'Alt text' field?

**Environment**
AEM version 6.5.0
Core Components 2.5.0

comanV:
- provided possibility of adding Aria Label and Title attributes to the Title's Link
Fixes #668

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
@@ -35,6 +35,8 @@ import com.google.common.collect.ImmutableSet;
 
 import static com.adobe.cq.wcm.core.components.commons.link.Link.PN_LINK_TARGET;
 import static com.adobe.cq.wcm.core.components.commons.link.Link.PN_LINK_URL;
+import static com.adobe.cq.wcm.core.components.commons.link.Link.PN_LINK_ACCESSIBILITY_LABEL;
+import static com.adobe.cq.wcm.core.components.commons.link.Link.PN_LINK_TITLE_ATTRIBUTE;
 
 /**
  * Simple implementation for resolving and validating links from model's resources.
@@ -62,7 +64,7 @@ public class LinkHandler {
     @ScriptVariable
     @org.apache.sling.models.annotations.Optional
     private PageManager pageManager;
-    
+
     /**
      * Resolves a link from the properties of the given resource.
      * @param resource Resource
@@ -86,7 +88,9 @@ public class LinkHandler {
         ValueMap props = resource.getValueMap();
         String linkURL = props.get(linkURLPropertyName, String.class);
         String linkTarget = props.get(PN_LINK_TARGET, String.class);
-        return Optional.ofNullable(getLink(linkURL, linkTarget).orElse(null));
+        String linkAccessibilityLabel = props.get(PN_LINK_ACCESSIBILITY_LABEL, String.class);
+        String linkTitleAttribute = props.get(PN_LINK_TITLE_ATTRIBUTE, String.class);
+        return Optional.ofNullable(getLink(linkURL, linkTarget, linkAccessibilityLabel, linkTitleAttribute).orElse(null));
     }
 
     /**
@@ -101,22 +105,24 @@ public class LinkHandler {
             return Optional.empty();
         }
         String linkURL = getPageLinkURL(page);
-        return Optional.of(new LinkImpl<>(linkURL, null, page));
+        return Optional.of(new LinkImpl<>(linkURL, null, null, null, page));
     }
 
     /**
-     * Builds a link with the given Link URL and target.
+     * Builds a link with the given Link URL, target, accessibility label, title.
      * @param linkURL Link URL
      * @param target Target
+     * @param linkAccessibilityLabel Link Accessibility Label
+     * @param linkTitleAttribute Link Title Attribute
      *
      * @return {@link Optional} of  {@link Link<Page>}
      */
     @NotNull
-    public Optional<Link<Page>> getLink(@Nullable String linkURL, @Nullable String target) {
+    public Optional<Link<Page>> getLink(@Nullable String linkURL, @Nullable String target, @Nullable String linkAccessibilityLabel, @Nullable String linkTitleAttribute) {
         String resolvedLinkURL = validateAndResolveLinkURL(linkURL);
         String resolvedLinkTarget = validateAndResolveLinkTarget(target);
         Page targetPage = getPage(linkURL).orElse(null);
-        return Optional.of(new LinkImpl<>(resolvedLinkURL, resolvedLinkTarget, targetPage));
+        return Optional.of(new LinkImpl<>(resolvedLinkURL, resolvedLinkTarget, linkAccessibilityLabel, linkTitleAttribute, targetPage));
     }
 
     /**
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
@@ -121,8 +121,10 @@ public class LinkHandler {
     public Optional<Link<Page>> getLink(@Nullable String linkURL, @Nullable String target, @Nullable String linkAccessibilityLabel, @Nullable String linkTitleAttribute) {
         String resolvedLinkURL = validateAndResolveLinkURL(linkURL);
         String resolvedLinkTarget = validateAndResolveLinkTarget(target);
+        String resolvedLinkAccessibilityLabel = validateAndResolveLinkAccessibilityLabel(linkAccessibilityLabel);
+        String resolvedLinkTitleAttribute = validateAndResolveLinkTitleAttribute(linkTitleAttribute);
         Page targetPage = getPage(linkURL).orElse(null);
-        return Optional.of(new LinkImpl<>(resolvedLinkURL, resolvedLinkTarget, linkAccessibilityLabel, linkTitleAttribute, targetPage));
+        return Optional.of(new LinkImpl<>(resolvedLinkURL, resolvedLinkTarget, resolvedLinkAccessibilityLabel, resolvedLinkTitleAttribute, targetPage));
     }
 
     /**
@@ -149,6 +151,36 @@ public class LinkHandler {
     private String validateAndResolveLinkTarget(String linkTarget) {
         if (linkTarget != null && VALID_LINK_TARGETS.contains(linkTarget)) {
             return linkTarget;
+        }
+        else {
+            return null;
+        }
+    }
+
+    /**
+     * Validates and resolves the link accessibility label.
+     * @param linkAccessibilityLabel Link accessibility label
+     *
+     * @return The validated link accessibility label or {@code null} if not valid
+     */
+    private String validateAndResolveLinkAccessibilityLabel(String linkAccessibilityLabel) {
+        if (!StringUtils.isBlank(linkAccessibilityLabel)) {
+            return linkAccessibilityLabel;
+        }
+        else {
+            return null;
+        }
+    }
+
+    /**
+     * Validates and resolves the link title attribute.
+     * @param linkTitleAttribute Link title attribute
+     *
+     * @return The validated link title attribute or {@code null} if not valid
+     */
+    private String validateAndResolveLinkTitleAttribute(String linkTitleAttribute) {
+        if (!StringUtils.isBlank(linkTitleAttribute)) {
+            return linkTitleAttribute;
         }
         else {
             return null;
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
@@ -87,6 +87,9 @@ public class LinkHandler {
     public Optional<Link> getLink(@NotNull Resource resource, String linkURLPropertyName) {
         ValueMap props = resource.getValueMap();
         String linkURL = props.get(linkURLPropertyName, String.class);
+        if (linkURL == null) {
+            return Optional.empty();
+        }
         String linkTarget = props.get(PN_LINK_TARGET, String.class);
         String linkAccessibilityLabel = props.get(PN_LINK_ACCESSIBILITY_LABEL, String.class);
         String linkTitleAttribute = props.get(PN_LINK_TITLE_ATTRIBUTE, String.class);
@@ -105,7 +108,22 @@ public class LinkHandler {
             return Optional.empty();
         }
         String linkURL = getPageLinkURL(page);
-        return Optional.of(new LinkImpl<>(linkURL, null, null, null, page));
+        return Optional.of(new LinkImpl<>(linkURL, null, page));
+    }
+
+    /**
+     * Builds a link with the given Link URL and target.
+     * @param linkURL Link URL
+     * @param target Target
+     *
+     * @return {@link Optional} of  {@link Link<Page>}
+     */
+    @NotNull
+    public Optional<Link<Page>> getLink(@Nullable String linkURL, @Nullable String target) {
+        String resolvedLinkURL = validateAndResolveLinkURL(linkURL);
+        String resolvedLinkTarget = validateAndResolveLinkTarget(target);
+        Page targetPage = getPage(linkURL).orElse(null);
+        return Optional.of(new LinkImpl<>(resolvedLinkURL, resolvedLinkTarget, targetPage));
     }
 
     /**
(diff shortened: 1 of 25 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Title.java
```diff
@@ -81,4 +81,24 @@ default boolean isLinkDisabled() {
         return false;
     }
 
+    /**
```
vladbailescu, line 84: This is one case where the newly introduced `Link` interface and the `LinkHandler` should help. Rather than defining these attributes inline for this component only, we should add them to the `Link#getHtmlAttributes`. We can then return a `Link` from the `Title` (something that's already due to happen in https://github.com/adobe/aem-core-wcm-components/pull/1419) and use these attributes from there.

content/src/content/jcr_root/apps/core/wcm/components/title/v2/title/title.html
```diff
@@ -22,6 +22,8 @@
      class="cmp-title">
     <h1 class="cmp-title__text" data-sly-element="${title.type}"><a
             data-sly-unwrap="${!title.linkURL || title.linkDisabled}" class="cmp-title__link" href="${title.linkURL}"
+            aria-label="${title.linkAccessibilityLabel}"
```
vladbailescu, line 25: If we make the `Title` return a `Link` then we can either do: `aria-label="title.link.htmlAttributes['aria-label']"` or just batch add all like `data-sly-attribute="${title.link.htmlAttributes}"`

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -101,22 +105,26 @@
...
      * @param linkURL Link URL
      * @param target Target
+     * @param linkAccessibilityLabel Link Accessibility Label
+     * @param linkTitleAttribute Link Title Attribute
      *
      * @return {@link Optional} of  {@link Link<Page>}
      */
     @NotNull
-    public Optional<Link<Page>> getLink(@Nullable String linkURL, @Nullable String target) {
+    public Optional<Link<Page>> getLink(@Nullable String linkURL, @Nullable String target, @Nullable String linkAccessibilityLabel, @Nullable String linkTitleAttribute) {
```
vladbailescu, line 121: To avoid changes in files unrelated to this issue, we should still provide a method with the old signature.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/DownloadImpl.java
```diff
@@ -170,7 +170,7 @@ private void initFileDownload(Resource file) {
                         extension = mimeTypeService.getExtension(format);
                     }
 
-                    url = linkHandler.getLink(getDownloadUrl(file) + "/" + filename, null).map(Link::getURL).orElse(null);
+                    url = linkHandler.getLink(getDownloadUrl(file) + "/" + filename, null, null, null).map(Link::getURL).orElse(null);
```
vladbailescu, line 173: Change not necessarily needed.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/DownloadImpl.java
```diff
@@ -203,7 +203,7 @@ private void initAssetDownload(String fileReference) {
                         extension = FilenameUtils.getExtension(filename);
                     }
 
-                    url = linkHandler.getLink(getDownloadUrl(downloadResource), null).map(Link::getURL).orElse(null);
+                    url = linkHandler.getLink(getDownloadUrl(downloadResource), null, null, null).map(Link::getURL).orElse(null);
```
vladbailescu, line 206: Change not necessarily needed.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TeaserImpl.java
```diff
@@ -235,7 +235,7 @@ private void initModel() {
                     // target page doesn't exist
                     if (this.isActionsEnabled()) {
                         return this.getActions().stream().findFirst()
-                                .map(action -> Optional.ofNullable(linkHandler.getLink(action.getURL(), null).orElse(null)))
+                                .map(action -> Optional.ofNullable(linkHandler.getLink(action.getURL(), null, null, null).orElse(null)))
```
vladbailescu, line 238: Change not necessarily needed.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TitleImpl.java
```diff
@@ -82,6 +82,14 @@
     @Nullable
     private String type;
 
+    @ValueMapValue(injectionStrategy = InjectionStrategy.OPTIONAL)
```
vladbailescu, line 85: These are not needed anymore, the `LinkHandler`  should handle them.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Title.java
```diff
@@ -62,12 +65,13 @@ default String getType() {
...
-     * Returns the Title's link URL, if one was set.
+     * Returns the Title's link, if one was set.
      *
-     * @return the title's link URL, if one was set, or {@code null}
-     * @since com.adobe.cq.wcm.core.components.models 12.4.0
+     * @return the title's link, if one was set, or {@code null}
+     * @since com.adobe.cq.wcm.core.components.models 13.0.0
      */
-    default String getLinkURL() {
+    @Nullable
```
vladbailescu, line 73: We need to keep backwards compatiblity. Instead of replacing this method with `getLink` we should keep it, mark it deprecated and make sure it defers to `getLink` in the implementation.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/package-info.java
```diff
@@ -34,7 +34,7 @@
  *      version, is bound to this proxy component resource type.
  * </p>
  */
-@Version("12.20.0")
+@Version("13.0.0")
```
vladbailescu, line 37: No major bumps, this will cause customer bundles to no longer start on an upgrade.

bundles/core/src/test/resources/title/exporter-title-jcr-title-type.json
```diff
@@ -10,5 +10,8 @@
             "dc:title": "Hello World"
         }
     },
+    "link": {
```
vladbailescu, line 13: Invalid links should return `null` so that they don't get rendered in JSON export.

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -149,6 +157,36 @@ private String validateAndResolveLinkTarget(String linkTarget) {
         }
     }
 
+    /**
+     * Validates and resolves the link accessibility label.
+     * @param linkAccessibilityLabel Link accessibility label
+     *
+     * @return The validated link accessibility label or {@code null} if not valid
+     */
+    private String validateAndResolveLinkAccessibilityLabel(String linkAccessibilityLabel) {
```
jckautzmann, line 166: As this method only validates the label, I'd call it `validateLinkAccessibilityLabel`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -149,6 +157,36 @@ private String validateAndResolveLinkTarget(String linkTarget) {
...
+        }
+    }
+
+    /**
+     * Validates and resolves the link title attribute.
+     * @param linkTitleAttribute Link title attribute
+     *
+     * @return The validated link title attribute or {@code null} if not valid
+     */
+    private String validateAndResolveLinkTitleAttribute(String linkTitleAttribute) {
```
jckautzmann, line 181: As this method only validates the title attribute, I'd call it `validateLinkTitleAttribute`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -101,22 +105,26 @@
...
+     * @param linkTitleAttribute Link Title Attribute
      *
      * @return {@link Optional} of  {@link Link<Page>}
      */
     @NotNull
-    public Optional<Link<Page>> getLink(@Nullable String linkURL, @Nullable String target) {
+    public Optional<Link<Page>> getLink(@Nullable String linkURL, @Nullable String target, @Nullable String linkAccessibilityLabel, @Nullable String linkTitleAttribute) {
         String resolvedLinkURL = validateAndResolveLinkURL(linkURL);
         String resolvedLinkTarget = validateAndResolveLinkTarget(target);
+        String resolvedLinkAccessibilityLabel = validateAndResolveLinkAccessibilityLabel(linkAccessibilityLabel);
```
jckautzmann, line 124: I'd rename it `validatedLinkAccessibilityLabel`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -101,22 +105,26 @@
...
+        String resolvedLinkTitleAttribute = validateAndResolveLinkTitleAttribute(linkTitleAttribute);
```
jckautzmann, line 125: I'd rename it `validatedLinkTitleAttribute`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TitleImpl.java
```diff
@@ -125,8 +133,8 @@ public String getType() {
     }
 
     @Override
-    public String getLinkURL() {
```
jckautzmann, line 128: This method cannot be removed for backwards compatibility reasons.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
```diff
@@ -327,7 +327,7 @@ protected void buildAreas() {
                     String href = StringUtils.removeAll(remainingTokens[0], "\"");
                     String target = remainingTokens.length > 1 ? StringUtils.removeAll(remainingTokens[1], "\"") : "";
 
-                    Link link = linkHandler.getLink(href, target).orElse(null);
+                    Link link = linkHandler.getLink(href, target, null, null).orElse(null);
```
jckautzmann, line 330: this change is not needed as getLink with 2 parameters should still be supported.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Title.java
jckautzmann: `getLink` should not be introduced in Title v1. It will be introduced in Title v3:
https://github.com/adobe/aem-core-wcm-components/pull/1419

bundles/core/src/test/resources/title/exporter-title-jcr-title-link-v2.json
```diff
@@ -1,13 +1,16 @@
...
     "linkDisabled": false,
-    "linkURL": "https://www.adobe.com",
     "dataLayer": {
         "title-6032c7e80a": {
             "repo:modifyDate": "2016-01-13T15:14:51Z",
             "xdm:linkURL": "https://www.adobe.com",
             "@type": "core/wcm/components/title/v2/title"
         }
     },
+    "link": {
```
jckautzmann, line 11: As `getLink` should not be introduced in v1 (see comment above), the `link` property won't be part of the JSON rendering in v1. This applies here and below.

content/src/content/jcr_root/apps/core/wcm/components/title/v2/title/title.html
```diff
@@ -19,9 +19,10 @@
      data-sly-test.text="${title.text}"
      data-cmp-data-layer="${title.data.json}"
      id="${component.id}"
-     class="cmp-title">
+    class="cmp-title">
     <h1 class="cmp-title__text" data-sly-element="${title.type}"><a
-            data-sly-unwrap="${!title.linkURL || title.linkDisabled}" class="cmp-title__link" href="${title.linkURL}"
+            data-sly-unwrap="${!title.link.htmlAttributes['href'] || title.linkDisabled}" class="cmp-title__link"
```
jckautzmann, line 24: `title.link` cannot be used in v1 (see comment above)

comanV: but this is v2, am I missing something?

bundles/core/src/test/resources/title/exporter-title-jcr-title-link-v2.json
comanV: also, this json is for v2

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Title.java
vladbailescu: I think it's ok to introduce `getLink` in v2 as well as long as we also keep the old `getLinkURL` method. We already did that in https://github.com/adobe/aem-core-wcm-components/blob/master/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/ListItem.java#L33-L42

content/src/content/jcr_root/apps/core/wcm/components/title/v2/title/title.html
vladbailescu: @jckautzmann is right, we should keep the old `title.linkURL` for `href` (for backwards compatibility) and only use `title.link.htmlAttributes` for the two new attributes. It will not output any errors, just skip the attributes if `title.link` is not defined/null.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Title.java
jckautzmann: IIRC we had decided to introduce the new `getLink` API and use it in HTL templates in a new component version e.g. #1419 ) to avoid any backwards compatibility issue. Am I missing something?

vladbailescu: I think we can keep backwards compatibility if we only use it to get the new accessibility attributes and use the old/deprecated method for the `href`. It will create some merge conflicts with #1419 but that can be addressed afterwards.

I see no point in adding the new attributes to the title model when we know the proper way forward is to get them via `Link`.

jckautzmann: >I think we can keep backwards compatibility if we only use it to get the new accessibility attributes
How would you do that as there is no API to get only the 2 accessibility properties?

jckautzmann: Answering to myself: we could instantiate the link object with only the  2 accessibility properties, the other ones being `null`. A bit hacky but I guess it would work until we introduce the new version of the Title component that supports the link handler.

vladbailescu: @jckautzmann , we should instantiate the Link as usual, continue to get the `href` from `getLinkURL` (but change the implementation there to get the value from `Link#getUrl`) and inject the rest of the attributes like: `aria-label="${title.link.htmlAttributes['aria-label']}"`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TitleImpl.java
```diff
@@ -125,10 +125,17 @@ public String getType() {
     }
 
     @Override
+    @JsonIgnore
```
vladbailescu, line 128: Don't ignore it, it still needs to be part of the JSON export for compatibility reasons. We will `@JsonIgnore` it in v3.

bundles/core/src/test/resources/title/exporter-title-jcr-title-link-v2.json
```diff
@@ -1,13 +1,16 @@
 {
     "id": "title-6032c7e80a",
     "linkDisabled": false,
-    "linkURL": "https://www.adobe.com",
```
vladbailescu, line 4: Don't remove this, see comment for `@JsonIgnore`!

vladbailescu (changes requested):

vladbailescu (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
```diff
@@ -58,8 +58,19 @@ public LinkImpl(String url, String target) {
...
+    }
+
+    /**
+     * @param url Link URL
+     * @param target Target
+     * @param reference Referenced WCM/DAM entity
+     * @param linkAccessibilityLabel Accessibility Label
+     * @param linkTitleAttribute Title attribute
+     */
+    LinkImpl(String url, String target, T reference, String linkAccessibilityLabel, String linkTitleAttribute) {
```
jckautzmann, line 71: I'd put the `T reference` argument at the beginning, as all other arguments are part of the link properties

PR #1519, opened 2021-04-19
[Navigation][Accessibility] Inconsistent isActive behavior

### LINKED ISSUE #688: [Navigation][Accessibility] Inconsistent isActive behavior

bpauli:
- add isCurrent flag to NavigationItem to check if the item is the current page or a redirect target
- adapt unit tests
- use current flag in templates

fixes #688

| Fixed Issues?            | Fixes #688 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


testing/it/ui-js/package-lock.json
```diff
@@ -130,7 +130,7 @@
         },
         "balanced-match": {
             "version": "1.0.0",
-            "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
+            "resolved": "https://artifactory.corp.adobe.com:443/artifactory/api/npm/npm-aem-release/balanced-match/-/balanced-match-1.0.0.tgz",
```
vladbailescu, line 133: Please revert these changes!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/NavigationImpl.java
```diff
@@ -254,8 +255,9 @@ public String getExportedType() {
      */
     private NavigationItem createNavigationItem(@NotNull final Page page, @NotNull final List<NavigationItem> children) {
         int level = page.getDepth() - (this.getNavigationRoot().getDepth() + structureStart);
+        boolean current = checkCurrent(page);
         boolean selected = checkSelected(page);
```
vladbailescu, line 259: This causes `checkCurrent` to be called twice in a row, I'd rather we pass the value of `current` to `checkSelected`.

vladbailescu (changes requested):

vladbailescu (approved):

PR #1522, opened 2021-04-20
[Link Handling][List] Apply URL mapping and vanity URLs to list links

### LINKED ISSUE #559: [Link Handling][List] Apply URL mapping and vanity URLs to list links
## Bug Report

When adding a page to a Fixed List using the List Component v2, if the page has a vanity URL, the list component is replacing the normal page name + .html with the vanity name. 

When clicked, this generates a 404 error because the URL generated by the component includes the full path to the page which is not part of the vanity URL. The correct vanity URL is http;//site.com/vanity while the full path is http://site.com/content/site/folder1/folder2/vanity.html. My link URL is being output as http://site.com/content/site/folder1/folder2/vanity instead of a functional URL.

When I remove the vanity URL, then the full path URL is used, and the link will work correctly.

Is there a way to override this or tell it not to use the vanity URL?

**Environment**
- AEM 6.3.3.3
- Core Components version 2.3.0 and 2.3.2
- Java(TM) SE Runtime Environment(build 1.8.0_171-b11)

vladbailescu:
* This is fixed by the new LinkHandler (#713), just adding specific tests

Closes #559

### DIFF
--- a/bundles/core/src/test/resources/list/v2/exporter-staticWithVanityPaths.json
+++ b/bundles/core/src/test/resources/list/v2/exporter-staticWithVanityPaths.json
@@ -1,0 +1,43 @@
+{
+    "id": "list-d860131bb1",
+    "dateFormatString": "yyyy-MM-dd",
+    "items": [
+        {
+            "id": "list-d860131bb1-item-29f192d859",
+            "path": "/content/list/pages/page_3",
+            "title": "Page 3",
+            "url": "/context/page_3.html",
+            ":type": "cq:PageContent",
+            "dataLayer": {
+                "list-d860131bb1-item-29f192d859": {
+                    "@type": "cq:PageContent",
+                    "dc:title": "Page 3",
+                    "xdm:linkURL": "/context/page_3.html"
+                }
+            }
+        },
+        {
+            "id": "list-d860131bb1-item-86df63380b",
+            "path": "/content/list/pages/page_4",
+            "title": "Page 4",
+            "url": "/context/4_page.html",
+            ":type": "cq:PageContent",
+            "dataLayer": {
+                "list-d860131bb1-item-86df63380b": {
+                    "@type": "cq:PageContent",
+                    "dc:title": "Page 4",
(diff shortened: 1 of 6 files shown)


bundles/core/src/test/resources/list/v2/exporter-staticWithVanityPaths.json
```diff
@@ -0,0 +1,43 @@
...
+                    "dc:title": "Page 3",
+                    "xdm:linkURL": "/context/page_3.html"
+                }
+            }
+        },
+        {
+            "id": "list-d860131bb1-item-86df63380b",
+            "path": "/content/list/pages/page_4",
+            "title": "Page 4",
+            "url": "/context/4_page.html",
```
jckautzmann, line 23: Shouldn't it be page_4.html as for page_3.html?

bundles/core/src/test/resources/list/v2/test-content.json
```diff
@@ -81,6 +95,22 @@
...
+                    "jcr:title": "Page 3",
+                    "sling:vanityPath": "/page_3.html"
+                }
+            },
+            "page_4": {
+                "jcr:primaryType": "cq:Page",
+                "jcr:content": {
+                    "jcr:primaryType": "cq:PageContent",
+                    "jcr:title": "Page 4",
+                    "sling:vanityPath": "/4_page.html"
```
jckautzmann, line 112: Shouldn't it be page_4.html as for page_3.html?

bundles/core/src/test/resources/list/v2/exporter-staticWithVanityPaths.json
vladbailescu: No, I specifically named the vanity path to be `4_page`

bundles/core/src/test/resources/list/v2/test-content.json
vladbailescu: No, I specifically named the vanity path to be `4_page`

jckautzmann (approved):

PR #1536, opened 2021-04-26
Expose LinkHandler Model

bpauli:
- add LinkHandler interface
- provide default link handler implementation without mapping to be backward compatible
- add mapping aware linkhandler


jckautzmann:
IMHO the link handler should not be a model as it is not bound to a resource/component but is rather a utility to manage links.

PR #1538, opened 2021-04-27
[Title V3][Accessibility] Possibility of adding description related to link in Title component

### LINKED ISSUE #668: [Title][Accessibility] Possibility of adding description related to link in Title component

comanV:
Fixes #668 for Title V3 Component


jckautzmann:
@comanV Could you please merge the `development` branch into `feature/title-v3` so that there are less files to review? Thx

content/src/content/jcr_root/apps/core/wcm/components/title/v3/title/_cq_dialog/.content.xml
```diff
@@ -18,9 +18,9 @@
     jcr:primaryType="nt:unstructured"
     jcr:title="Title"
     sling:resourceType="cq/gui/components/authoring/dialog"
-    extraClientlibs="[core.wcm.components.title.v2.editor]"
+    extraClientlibs="[core.wcm.components.title.v3.editor]"
```
jckautzmann, line 21: Instead of defining a specific client library for Title v3, we should use the client library of v2 (`core.wcm.components.title.v2.editor`) as the logic for the new link attributes is already available in v2.

jckautzmann: Is it necessary to create an edit dialog for v3? If not, we could remove it and automatically use the one from v2.

content/src/content/jcr_root/apps/core/wcm/components/title/v3/title/clientlibs/editor/js/title.js
```diff
@@ -0,0 +1,233 @@
+/*******************************************************************************
```
jckautzmann, line 1: Does it duplicate `title.js` of Title v2? If so, we could remove it and use the v2 client lib (see comment above)

content/src/content/jcr_root/apps/core/wcm/components/title/v3/title/_cq_dialog/.content.xml
vladbailescu: I think it makes sense the new version is as independent as possible from the previous version(s), just in case we might want to remove the previous versions from the content package (and only ship them in a `-deprecated` package). Hence, I would keep the v3 clientlibs and dialog.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/jackson/LinkHtmlAttributesSerializer.java
```diff
@@ -0,0 +1,62 @@
...
+    public LinkHtmlAttributesSerializer() { this(null); }
+
+    protected LinkHtmlAttributesSerializer(Class<Map<String, String>> t) { super(t); }
+
+    private Map<String, String> filteredMap = new LinkedHashMap<>();
+
+    @Override
+    public void serialize(Map map, JsonGenerator gen, SerializerProvider serializers) throws IOException {
+        gen.writeStartObject();
+            for (Map.Entry<String, String> entry : filteredMap.entrySet()) {
```
vladbailescu, line 47: I believe assuming `isEmpty` will **ALWAYS** get called before `serialize`, with the same input, is a weak-point.

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
```diff
@@ -100,7 +102,8 @@ public boolean isValid() {
      * @return {@link Map} of HTML attributes, may include the URL as {@code href}
      */
     @Override
-    @JsonIgnore  // exclude HTML-specific attributes in JSON
+    @JsonInclude(Include.NON_EMPTY)
+    @JsonSerialize(using = LinkHtmlAttributesSerializer.class)
```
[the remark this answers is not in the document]
comanV: Done

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/jackson/LinkHtmlAttributesSerializer.java
comanV: Made some changes in this regard.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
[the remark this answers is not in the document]
vladbailescu: 👍  You'll need to update the test content too to pass the tests.

content/src/content/jcr_root/apps/core/wcm/components/title/v3/title/_cq_dialog/.content.xml
comanV: Made the changes based on the latest clarifications: we will duplicate HTL scripts and dialogs for the new component versions, but we won't duplicate the clientlibs, instead we will use the former version by referencing the category.

vladbailescu (approved):

jckautzmann (approved):

PR #1541, opened 2021-04-27
[LinkHandling] Version 2.16.0 automatic path/URL mapping is backwards…

bpauli:
… incompatible with existing mapping configuration of wknd.site

- add LinkProcessor and LinkProcessor factory service
- add service user mapping for MappingLinkProcessor

| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/LinkProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public interface LinkProcessor {
+
+    /**
+     * Process a given link
+     * @param linkUrl the link url
+     * @return the processed link if the processor can handle the link, otherwise null
+     */
+    @Nullable String process(@Nullable String linkUrl);
+}
```
jckautzmann, line 33: we may want to introduce a method that verifies if the processor can process the given URL. E.g:
```
boolean canProcess(@Nullable String linkUrl);
```
This would enable defining which kind of links the processor can process based on regex in OSGI configuration.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
```diff
@@ -84,6 +86,17 @@ default String getProcessedURL() {
...
     }
 
+
+    /**
+     * The full URL which also contains the scheme and host information.
+     * This is usually created with a {@link com.day.cq.commons.Externalizer} service
+     * @return Full link URL or {@code null} if link is invalid or can't be externalized.
+     */
+    @Nullable
+    default String getFullURL() {
```
vladbailescu, line 96: I would rather we stick to the following names: `getURL`, `getMappedURL` and `getExternalizedURL`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/LinkRequest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * Simple PoJo which contains required information to create a {@link com.adobe.cq.wcm.core.components.commons.link.Link} by the
+ * {@link LinkProcessor}
+ *
+ * @since com.adobe.cq.wcm.core.components.services.link 1.0.0
+ */
+@ProviderType
+public final class LinkRequest<T> {
```
vladbailescu, line 33: I'm not sure having a wrapper object here is that helpful, I find it a bit confusing, maybe we should pass the path, reference and attributes directly.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkProcessorFactoryImpl.java
```diff
@@ -0,0 +1,60 @@
...
+import org.osgi.service.component.annotations.ReferenceCardinality;
+import org.osgi.service.component.annotations.ReferencePolicy;
+
+import com.adobe.cq.wcm.core.components.commons.link.Link;
+import com.adobe.cq.wcm.core.components.services.link.LinkProcessor;
+import com.adobe.cq.wcm.core.components.services.link.LinkProcessorFactory;
+import com.adobe.cq.wcm.core.components.services.link.LinkRequest;
+
+@Component(immediate = true, service = LinkProcessorFactory.class)
+public class LinkProcessorFactoryImpl implements LinkProcessorFactory {
```
vladbailescu, line 33: It should not be necessary to have a factory for getting references to services. We should just inject them in the Sling model and the best ranked one should be resolved by OSGi.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkProcessorFactoryImpl.java
```diff
@@ -0,0 +1,60 @@
...
+    protected void unbindLinkProcessor(final LinkProcessor linkProcessor, final Map<String, Object> props) {
+        synchronized (linkProcessors) {
+            linkProcessors.unbind(linkProcessor, props);
+        }
+    }
+
+    @Override
+    public LinkProcessor getProcessor(LinkRequest linkRequest) {
+        for (LinkProcessor linkProcessor: linkProcessors) {
+            Link link = linkProcessor.process(linkRequest);
```
vladbailescu, line 53: Even if we would need a factory to iterate through processors, calling `process` when deciding is not a good approach if we are also going to call it again when doing the actual processing. Either we add an `accept` method or we handle the processing result here directly.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/DefaultLinkProcessor.java
```diff
@@ -0,0 +1,79 @@
...
+
+@Component(property = Constants.SERVICE_RANKING + ":Integer=" + Integer.MAX_VALUE,
+           service = LinkProcessor.class)
+public class DefaultLinkProcessor<T> implements LinkProcessor<T> {
+
+    @Reference
+    Externalizer externalizer;
+
+    @Override
+    public @Nullable Link<T> process(@NotNull LinkRequest<T> linkRequest) {
```
vladbailescu, line 41: I would rather we define three Link processing stages: `validation`, `mapping` and `externalizing` and allow developers to override (through service ranking) whichever stage(s) they need.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
jckautzmann: We may want to introduce `getCustomURL` as an extension point to allow customized processing specific to a project.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/LinkRequest.java
jckautzmann: I find the name confusing. If we decide to keep it, I'd rename it `LinkData`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
```diff
@@ -49,16 +50,51 @@
...
     private final Map<String, String> htmlAttributes;
-    private String processedUrl;
-    private boolean processed = false;
+    private final String externalizedUrl;
 
-    public LinkImpl(String url, T reference, LinkProcessor linkProcessor, Map<String, String> htmlAttributes) {
+    /**
+     * @param url Link URL
+     */
+    public LinkImpl(String url) {
```
vladbailescu, line 61: Is there a reason for introducing all these constructors? I don't see them being used (outside tests) and we could do without.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
```diff
@@ -90,21 +126,20 @@ public boolean isValid() {
...
-            processed = true;
-        }
-        return processedUrl;
+    public @Nullable String getMappedURL() {
+        return mappedUrl;
+    }
+
+    @Override
+    @JsonIgnore
+    public @Nullable String getExternalizedURL() {
```
vladbailescu, line 135: We should add some test for extenalization.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    boolean canHandle(String path, SlingHttpServletRequest request);
+
+    /**
+     * Path is prefixed with the context path and escaped
+     * @param path the path which needs to be checked /fixed.
+     * @param request the current request
+     * @return the escaped absolute URL path with context path
+     */
+    @NotNull String fixPath(String path, SlingHttpServletRequest request);
```
jckautzmann, line 50: I would not keep `fix` in the method name as it gives the impression that the path is wrong. I'd rename it `prefixPath` or `addContextPath`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+ * @since com.adobe.cq.wcm.core.components.services.link 1.0.0
+ */
+@ConsumerType
+public interface PathProcessor {
+
+    /**
+     * Checks if the current processor can handle the requested path
+     * @param path the path which should be processed
+     * @param request the current request
+     * @return {@code true} if the processor can handle the request, otherwise {@code false} and the next path processor is used by the
```
vladbailescu, line 39: I think it should be: *if the processor can handle the path*

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+     * {@link com.adobe.cq.wcm.core.components.internal.link.LinkHandler}
+     */
+    boolean canHandle(String path, SlingHttpServletRequest request);
+
+    /**
+     * Path is prefixed with the context path and escaped
+     * @param path the path which needs to be checked /fixed.
```
jckautzmann, line 46: I would say it as: the path which needs to be prefixed.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
vladbailescu, line 50: I would name this `sanitize`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+public interface PathProcessor {
+
+    /**
+     * Checks if the current processor can handle the requested path
+     * @param path the path which should be processed
+     * @param request the current request
+     * @return {@code true} if the processor can handle the request, otherwise {@code false} and the next path processor is used by the
+     * {@link com.adobe.cq.wcm.core.components.internal.link.LinkHandler}
+     */
+    boolean canHandle(String path, SlingHttpServletRequest request);
```
vladbailescu, line 42: I would name this `accepts`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    @NotNull String fixPath(String path, SlingHttpServletRequest request);
+
+    /**
+     * Map the fixed path to the internal resource. Usually this is done be the {@link ResourceResolver#map(String)} method.
+     * @param path the resource path
+     * @param request the current request
+     * @return the mapped path
+     */
+    @NotNull String mapPath(String path, SlingHttpServletRequest request);
```
vladbailescu, line 58: I would name this `map`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Externalize the given mapped path.
+     * @param path the resource path
+     * @param request the current request
+     * @return the external link of the given path
+     */
+    @NotNull String externalizeLink(@NotNull String path, @NotNull SlingHttpServletRequest request);
```
vladbailescu, line 66: I would name this `externalize`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Path is prefixed with the context path and escaped
+     * @param path the path which needs to be checked /fixed.
+     * @param request the current request
+     * @return the escaped absolute URL path with context path
+     */
+    @NotNull String fixPath(String path, SlingHttpServletRequest request);
+
+    /**
+     * Map the fixed path to the internal resource. Usually this is done be the {@link ResourceResolver#map(String)} method.
```
jckautzmann, line 53: I'd write it as:
Maps the path. This is usually done using the {@link ResourceResolver#map(String)} method.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkImpl.java
```diff
@@ -49,16 +50,51 @@
...
+     * @param mappedUrl mapped Link URL
+     * @param externalizedUrl external Link URL
+     * @param reference Referenced WCM/DAM entity
+     */
+    LinkImpl(String url, String mappedUrl, String externalizedUrl, T reference) {
+        this(url, mappedUrl, externalizedUrl, reference, null);
+    }
+
+    public LinkImpl(@Nullable String url, @Nullable String mappedUrl, @Nullable String externalizedUrl, @Nullable T reference,
+                    @Nullable Map<String, Optional<String>> htmlAttributes) {
```
vladbailescu, line 93: Is there a reason for making `htmlAttributes` values optional?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
jckautzmann, line 66: I'd rename it `getExternalizedPath()`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param path the resource path
+     * @param request the current request
+     * @return the mapped path
+     */
+    @NotNull String mapPath(String path, SlingHttpServletRequest request);
+
+    /**
+     * Externalize the given mapped path.
```
jckautzmann, line 61: Does the path need to be mapped? Otherwise I'd keep it generic:
Externalizes the given path.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
```diff
@@ -74,13 +72,24 @@ default String getURL() {
...
-    default String getProcessedURL() {
+    default String getMappedURL() {
+        return null;
+    }
+
+
+    /**
+     * The externalized URL which also contains the scheme and host information.
+     * This is usually created with a {@link com.day.cq.commons.Externalizer} service
+     * @return Full link URL or {@code null} if link is invalid or can't be externalized.
```
vladbailescu, line 89: There's a disconnect here in documentation. On one hand `Link#getExternalizedURL` claims to return `null` if link can't be externalized, on the other hand the `PathProcessor#externalizePath` is annotated to return `@NotNull` (and the default implementation will return the original/nonexternalized path in case of errors).

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+     * Checks if the current processor can handle the requested path
+     * @param path the path which should be processed
+     * @param request the current request
+     * @return {@code true} if the processor can handle the request, otherwise {@code false} and the next path processor is used by the
+     * {@link com.adobe.cq.wcm.core.components.internal.link.LinkHandler}
+     */
+    boolean canHandle(String path, SlingHttpServletRequest request);
+
+    /**
+     * Path is prefixed with the context path and escaped
```
jckautzmann, line 45: I'd rename it as:
Escapes the path and prefixes it with the context path.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
vladbailescu: Also, do we expect the `request` would even be needed here?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
jckautzmann: I just saw Vlad's comment :)
I'd go with: 
`prefix()`
`map()`
`externalize()`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
bpauli: I would keep the request. I can think of scenarios where customers would like to check if the path should be handled by a custom path processor based on the request parameter or request path info.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
bpauli: If the url is null and the link is invalid also the Link#getExteralizedURL will return null now.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
```diff
@@ -64,7 +62,7 @@ default boolean isValid() {
     }
 
     /**
-     * The link URL, supports context path and vanity paths.
+     * The link URL, supports context path and escaping.
```
jckautzmann, line 65: I'd go with: Returns the link URL ...

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
```diff
@@ -74,13 +72,24 @@ default String getURL() {
     }
 
     /**
-     * The processed URL, in case any extra processing can be done.
+     * The mapped URL, which supports mapping and vanity path.
```
jckautzmann, line 75: I'd go with: Returns the mapped URL ...

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
```diff
@@ -74,13 +72,24 @@ default String getURL() {
...
      */
     @Nullable
-    default String getProcessedURL() {
+    default String getMappedURL() {
+        return null;
+    }
+
+
+    /**
+     * The externalized URL which also contains the scheme and host information.
```
jckautzmann, line 87: I'd go with: Returns the externalized URL ...

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+     * Checks if the given path should be handled by this processor.
+     * @param path the path which should be processed
+     * @param request the current request
+     * @return {@code true} if the processor should handle the given path, otherwise {@code false} and the next path processor is applied by
+     * the {@link com.adobe.cq.wcm.core.components.internal.link.LinkHandler} if present
+     */
+    boolean accepts(@NotNull String path, @NotNull SlingHttpServletRequest request);
+
+    /**
+     * Sanitize the given path by doing proper escaping and prepend the context path if needed.
```
jckautzmann, line 45: minor: Sanitizes ... and prepends ...

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param path the path which needs to be sanitized
+     * @param request the current request
+     * @return the escaped absolut path with optional context path information
```
jckautzmann, line 48: absolut -> absolute

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    @NotNull String sanitize(@NotNull String path, @NotNull SlingHttpServletRequest request);
+
+    /**
+     * Apply mappings to the given path. Usually this is done with the {@link ResourceResolver#map(String)} method.
```
jckautzmann, line 53: Apply -> Applies

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param path the path which should be mapped
+     * @param request the current request
+     * @return the mapped path
+     */
+    @NotNull String map(@NotNull String path, @NotNull SlingHttpServletRequest request);
+
+    /**
+     * Externalize the given path.
```
jckautzmann, line 61: Externalize -> Externalizes

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
bpauli: The @return tag should make this clear. It's also not aligned with other interfaces like OEmbedClient for example.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
bpauli: The @return tag should make this clear. It's also not aligned with other interfaces like OEmbedClient for example.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
bpauli: The @return tag should make this clear. It's also not aligned with other interfaces like OEmbedClient for example.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
bpauli: fixed typo in last commit.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
bpauli: fixed typo in last commit.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
bpauli: fixed typo in last commit.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/link/PathProcessor.java
bpauli: fixed typo in last commit.

vladbailescu (approved):

PR #1552, opened 2021-05-10
[Core] [Unit Test] Core module unit tests take very long time due to heavy GC activity #1544

### LINKED ISSUE #1544: [Core] [Unit Test] Core module unit tests take very long time due to heavy GC activity
## Bug Report

**Current Behavior**
Once running a full set of tests within the core module, each test is gradually taking more time to execute, to some very long times like minutes. For low memory limit (i.e. 1G) it stops executing. Running them separately, or in a small set, does not bring this effect. The problem seems to be caused by some either memory leak or heavy memory work - connected monitor, such as VisualVM, shows the significant activity of GC  (more than 10% of processor's time) related to the moment when the test suite slows down.

**Expected behavior/code**
These tests are not complicated and should run smoothly, taking normally around 100ms per test and they should not require that long time to execute.

**Environment**
- current master branch, running it on a local machine, special EC2 instance free of other dependencies, and it also has the same effect (although I couldn't connect VisuamVM there) in GitHub actions (based on logs). 

**Possible Solution**
Tracking down the memory leak and fixing this problem either by removing the library that causes it or fixing the code.

I will add screenshots of VisualVM if necessary, but I think this may be reproduced everywhere, without any significant difference between forkCount values in Maven or MAVEN_OPTS settings. It is more visible when forkCount=0, given that JVM is already running for some time, but it impacts the most when forkCount = 1, and the process has no that much memory to allocate that potential memory leak.

vladbailescu:
* Created extension for clearing out self4j-test loggers

Closes #1544

### DIFF
--- a/bundles/core/src/test/java/com/adobe/cq/wcm/core/components/Slf4jTestCleanupExtension.java
+++ b/bundles/core/src/test/java/com/adobe/cq/wcm/core/components/Slf4jTestCleanupExtension.java
@@ -1,0 +1,28 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2019 Adobe
+ ~
+ ~ Licensed under the Apache License, Version 2.0 (the "License");
+ ~ you may not use this file except in compliance with the License.
+ ~ You may obtain a copy of the License at
+ ~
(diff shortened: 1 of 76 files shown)


bpauli (commented):
This approach is not thread safe. If tests are running in parallel the memory state can be cleared by another thread which leads to failing tests.

jckautzmann (approved):

vladbailescu:
@bpauli , you're right, that was not a threadsafe approach (but calling `clear` was not working). I removed the `uk.org.lidalia.slf4j-test` and replaced it with Mockito verification.

PR #1556, opened 2021-05-11
Multi-Page application support for PWA

kbalasub:
This PR adds support for multi-page applications (MPAs) to be recognized as Progressive Web Apps(PWAs) without having to configure each page. With this change you can configure PWA properties once at the project level or even at the localization level in the tree (e.g., us/en). When you access any page within the MPA this solution will traverse up the tree to see if any parent page has PWA properties configured and if so use those properties to show up as a PWA with the installable icon in the browser.


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PWAImpl.java
```diff
@@ -48,26 +48,31 @@
...
-        this.manifestPath = replaceSuffix(startURL, MANIFEST_NAME);
-
-        PageManager pageManager = resource.getResourceResolver().adaptTo(PageManager.class);
-        if (pageManager != null) {
-            Page page = pageManager.getContainingPage(resource);
-            if (page != null) {
+        Page page = pageManager.getContainingPage(resource);
+        while (page != null) {
+            Resource contentResource = page.getContentResource();
+            if (contentResource == null) {
```
vladbailescu, line 59: I find the logic here (L59-75, using `continue` and `break`) unnecessary complicated. I would write it as:
```java
if (contentResource != null) {
    ValueMap valueMap = contentResource.getValueMap();
    Boolean isPWAEnabled = valueMap.get(PN_PWA_ENABLED, Boolean.class);
    if (isPWAEnabled != null && isPWAEnabled) {
        ...
        return; //could be break if you need more operations after the loop
    }
}
page = page.getParent();
```

vladbailescu (changes requested):

testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/PageIT.java
```diff
@@ -55,4 +64,47 @@ public void testBrandSlug() throws ClientException {
...
         GraniteAssert.assertRegExFind(content, "<title>Simple SubPage \\| Core Components</title>");
     }
+
+    @Test
+    public void testPWAProperties() throws ClientException {
+        String content = adminAuthor.doGet("/content/core-components/simple-page.html", 200).getContent();
+        adminAuthor.getUrl("/");
+
+        // Manifest
+        GraniteAssert.assertRegExFind(content, "<link rel=\"manifest\" href=\".*/content/foo/us/en/manifest.webmanifest\" crossorigin=\"use-credentials\"/>");
```
vladbailescu, line 74: Added support for context path, would fail in our testing pipeline otherwise as we always use a context path.

testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/PageIT.java
```diff
@@ -55,4 +64,47 @@ public void testBrandSlug() throws ClientException {
...
+        Pattern swconfigPattern = Pattern.compile("const swconfig = (?<swconfigjson>\\{.*})", Pattern.MULTILINE);
+        Matcher swconfigMatcher= swconfigPattern.matcher(content);
+        assertTrue(swconfigMatcher.find());
+        Gson gson = new Gson();
+        JsonObject swconfigJson = gson.fromJson(swconfigMatcher.group("swconfigjson"), JsonObject.class);
+        assertNotNull(swconfigJson);
+        assertEquals("staleWhileRevalidate", swconfigJson.get("pwaCachestrategy").getAsString());
+        assertNotNull(swconfigJson.get("pwaPrecache").getAsJsonArray());
+        assertNotNull(swconfigJson.get("pwaCachingpaths").getAsJsonArray());
+        //assertNotNull(swconfigJson.get("pwaOfflineClientlibs").getAsJsonArray());
```
vladbailescu, line 108: Added better handling of JSON as the previous was failing in my local development environment.

vladbailescu: You might want to expand assertions by walking the expected tree.

testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/PageIT.java
```diff
@@ -55,4 +64,47 @@ public void testBrandSlug() throws ClientException {
         content = adminAuthor.doGet("/content/core-components/simple-page/simple-subpage.html", 200).getContent();
         GraniteAssert.assertRegExFind(content, "<title>Simple SubPage \\| Core Components</title>");
     }
+
+    @Test
+    public void testPWAProperties() throws ClientException {
```
vladbailescu, line 69: These tests should probably be skipped on 6.4 and 6.5 if the PWA features were not backported. @msagolj, any pointers here?

vladbailescu (changes requested):
I reduced the service worker assertions because the previous test was failing on my environment.

Also, please see inline comments!

testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/PageIT.java
msagolj: - create an Empty Interface named e.g. `IgnoreOn64` and `IgnoreOn65`
- Mark tests to ignore with `@Category({IgnoreOn64.class, IgnoreOn65})`
- In the Jenkinsfile we then have to add a` .withCategoryFilter("IgnoreOn64")` etc. for the corresponding `IntergrationTestRun `configurations.

testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/PageIT.java
kbalasub: Done.

testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/PageIT.java
kbalasub: Done. @msagolj Grateful if you could update the JenkinsFile accordingly.

vladbailescu (approved):

PR #1599, opened 2021-06-03
Test Content for Selenium Testing

absarasw:
This PR adds a separate test content package for Selenium and other type of testing. Its a copy of the test content found in testing/it/ui-js for hobbes tests.


msagolj:
the new package needs to be added in the root pom as well :
https://github.com/absarasw/aem-core-wcm-components/blob/hobbes-to-selenium/pom.xml#L64-L77

pom.xml
```diff
@@ -69,6 +69,7 @@
         <module>testing/it/http</module>
         <module>testing/it/e2e</module>
         <module>testing/it/e2e-selenium</module>
+        <module>testing/it/core-wcm-components-it-content</module>
```
vladbailescu, line 72: I would rename it to just `it-content`

testing/it/core-wcm-components-it-content/src/content/META-INF/vault/filter.xml
```diff
@@ -0,0 +1,32 @@
...
+    <!-- Sample content -->
+    <filter root="/conf/core-components"/>
+    <filter root="/content/core-components"/>
+    <filter root="/content/campaigns/core_component_brand"/>
+    <filter root="/content/dam/core-components"/>
+    <filter root="/content/cq:tags/core-components"/>
+    <filter root="/content/experience-fragments/core-components-test"/>
+    <filter root="/home/users/core-component"/>
+    <filter root="/etc/blueprints/corecomp_blueprint"/>
+    <!-- Hobbes IT folder -->
```
vladbailescu, line 27: This comment is misleading, it no longer holds Hobbes ITs

testing/it/core-wcm-components-it-content/src/content/jcr_root/apps/core/wcm/tests/components/.content.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vladbailescu, line 1: Are the components here needed anymore? Maybe we can drop them

testing/it/core-wcm-components-it-content/src/content/META-INF/vault/filter.xml
```diff
@@ -0,0 +1,32 @@
...
+    <filter root="/apps/core/wcm/tests"/>
+    <filter root="/apps/msm">
+        <include pattern="/apps/msm/corecomp_blueprint(/.*)?"/>
```
vladbailescu, line 30: Are the blueprints used? Perhaps we can drop them

testing/it/core-wcm-components-it-content/src/content/META-INF/vault/filter.xml
```diff
@@ -0,0 +1,32 @@
...
+<workspaceFilter version="1.0">
+    <!-- Sample content -->
+    <filter root="/conf/core-components"/>
+    <filter root="/content/core-components"/>
+    <filter root="/content/campaigns/core_component_brand"/>
+    <filter root="/content/dam/core-components"/>
+    <filter root="/content/cq:tags/core-components"/>
+    <filter root="/content/experience-fragments/core-components-test"/>
+    <filter root="/home/users/core-component"/>
+    <filter root="/etc/blueprints/corecomp_blueprint"/>
```
vladbailescu, line 26: This should be obsolete

testing/it/core-wcm-components-it-content/src/content/META-INF/vault/filter.xml
absarasw: Not required now. For 6.5 having this configuration is necessary to have Blueprint Tab in properties page. But in 6.6 the dependency was removed since this configuration exists in /libs or /apps and it can not be modified on the run. This dependency removal change need to be backported in 6.5 as well.
I have marked the test failing on 6.5 for now. We can enable the test once that fix is released.

absarasw: @vladbailescu I think we can remove the /apps folder completely. Is there any dependency on that path for http tests you are aware of?

PR #1601, opened 2021-06-03
[Button] Allow to open links in new tabs

### LINKED ISSUE #1600: [Button] Allow to open links in new tabs
As an author, I want to configure whether links do open in a new tab for the visitors.

Requirements:
- Authors need a checkbox to select whether the link is to be opened in a new tab or not.
- When selected, the generated link element gets the attribute target="_blank".
- By default, the checkbox would be unchecked.
- This behavior doesn't need to be configurable via the policy.

SITES-925

bpauli:
- add checkbox to button dialog to set link target attribute
- add button e2e test to test new checkox
- update readme of button v2 component
- adapt unit test

| Fixed Issues?            | Fixes #1600 
| Patch: Bug Fix?          | 👎
| Minor: New Feature?      |  👍
| Any Dependency Changes?  |


kwin:
Shouldn't this rather be part of the link handler? It is useful in all contexts where links are used.

jckautzmann:
@kwin The logic to handle the `linkTarget` property is already part of the `LinkHandler`. This PR adds the capability for an author to set the `linkTarget` property of the button.

jckautzmann (approved):

PR #1626, opened 2021-06-15
GH-1625 - [Image] The featured image is broken when an allowed width …

### LINKED ISSUE #1625: [Image] The featured image is broken when an allowed width is configured
Steps to reproduce:
- Open a page and define a featured image
- Set an allowed width to 400 in the policy dialog of the image component
- Add an image component to the page

-> the image is broken.

Root cause analysis: 
The src attribute is set to something like:
```
/content/core-components-examples/library/page-authoring/image/_jcr_content/_cq_featuredimage.coreimg.80.400.jpeg/1623320944578/lava-into-ocean.jpeg
```

The `cq:featured` resource has no policy defined, which makes the Adaptive Image Servlet (AIS) return a broken image.

Possible fix: the `src` attribute should be based on the image resource of the page (the caller resource), not on the featured image resource.

jckautzmann:
…is configured

- the src attribute is based on the path of the image component
- if the image has no content, the AIS takes the content of the featured imapge of the page

| Fixed Issues?            | Fixes #1625 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -182,13 +183,26 @@ protected void doGet(@NotNull SlingHttpServletRequest request, @NotNull SlingHtt
...
-                response.sendError(HttpServletResponse.SC_NOT_FOUND);
-                return;
+                PageManager pageManager = resourceResolver.adaptTo(PageManager.class);
+                if (pageManager != null) {
+                    Page page = pageManager.getContainingPage(component);
+                    if (page != null) {
+                        Resource featuredImage = ComponentUtils.getFeaturedImage(page);
+                        if (featuredImage != null) {
+                            imageComponent = new ImageComponent(featuredImage);
+                            if (imageComponent.source == Source.NONEXISTING) {
```
vladbailescu, line 196: I would move this after the attempt of getting the `imageComponent` (Line 206) from the featured image, to cover the null-checks branching as well.

vladbailescu (changes requested):
Tests need to be updated as well.

bpauli (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
jckautzmann: Addressed in https://github.com/adobe/aem-core-wcm-components/pull/1626/commits/a5167ad95344a28b7008cb4e885be29ec99a5e2c

jckautzmann:
@vladbailescu Tests have been adapted in https://github.com/adobe/aem-core-wcm-components/pull/1626/commits/a5167ad95344a28b7008cb4e885be29ec99a5e2c

PR #1647, opened 2021-06-25
CIF-2102: add robots, canonical and language alternates

Buuhuu:
| Fixed Issues?            |  CIF-2102
| Patch: Bug Fix?          | 
| Major: Breaking Change?  | 
| Documentation Provided   | 👍
| Any Dependency Changes?  | Yes

With the introduction of xml sitemaps a couple of implicit requirements needed to be implemented: 

- sitemaps should not include pages, which robots meta tag is set to noindex
- sitemaps should include the canonical urls of pages 
- when available sitemaps should include language alternates

This ticket is about adding the information now available to sitemaps to the page's head.


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -88,6 +92,11 @@
      */
     public static final String PN_MAIN_CONTENT_SELECTOR_PROP = "mainContentSelector";
 
+    /**
+     * Property name of the style property that enables/disables rendering of the alternate language links
+     */
+    public static final String PN_RENDER_ALTERNATE_LANGUAGES = "renderAlternateLanguageLinks";
```
vladbailescu, line 98: I know we are not very consistent with this (see above) but I prefer we have the constants that describe properties as members of the model interface. Also, in this case, the naming should probably begin with `PN_STYLE_*`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -311,4 +326,42 @@ public String getMainContentSelector() {
...
+        try {
+            SeoTags seoTags = resource.adaptTo(SeoTags.class);
+            return seoTags != null ? seoTags.getCanonicalUrl() : null;
+        } catch (NoClassDefFoundError ex) {
+            return pathProcessor.externalize(currentPage.getPath(), request);
+        }
+    }
+
+    @Override
+    public @NotNull Map<Locale, String> getAlternateLanguageLinks() {
```
vladbailescu, line 341: Would be great, as general practice, if we lazy-cached the output of these methods so that we don't have to do all the processing on each call.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/seo/LanguageNavigationSiteRootSelectionStrategy.java
```diff
@@ -0,0 +1,146 @@
...
+            Integer structureDepth = getPropertyLazy(languageNavigation, LanguageNavigation.PN_STRUCTURE_DEPTH, Integer.class);
+            if (structureDepth != null) {
+                return structureDepth;
+            }
+        }
+        return defaultSelectionStrategy != null ? defaultSelectionStrategy.getStructuralDepth(page) : 1;
+    }
+
+    private Resource findLanguageNavigation(Resource contentResource) {
+        Resource templatedResource = contentResource.adaptTo(TemplatedResource.class);
```
vladbailescu: At some point it would make sense to have some sort of caching as this is potentially expensive performance-wise.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/seo/LanguageNavigationSiteRootSelectionStrategy.java
```diff
@@ -0,0 +1,146 @@
...
+            Resource result = findFirst(child, condition);
+            if (result != null) {
+                return result;
+            }
+        }
+
+        return null;
+    }
+
+    private static <T> T getPropertyLazy(Resource resource, String property, Class<T> type) {
```
vladbailescu, line 126: This is probably something that should be in a common helper. Also, the name is a bit misleading, perhaps something more like `getPropertyOrStyle` would be better.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/seo/LanguageNavigationSiteRootSelectionStrategy.java
Buuhuu: True, for a start it would be good to cache the returned language navigation for the given `Page` so that when it is called more than once with the same object it does not traverse again. I will use https://guava.dev/releases/18.0/api/docs/com/google/common/cache/CacheBuilder.html#weakKeys() to build a short lived cache.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
Buuhuu: Will do that

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/seo/LanguageNavigationSiteRootSelectionStrategy.java
Buuhuu: Will move it to `Utils` and rename it as suggested.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
vladbailescu: It's not just the `adaptTo` part but also `SeoTags#getAleternateLanguages` and even the `Collections#unmodifiable*`

Buuhuu: If the `SeoTags#getAlternateLanguages()` is an expensive implementation worth caching, is an implementation detail and should be implemented by the implemented by the SeoTags implementation. The `PageImpl` as consumer should trust the implementation to perform optional. The `Collections#unmodifiableMap()` and `Collections#emptyMap()` should be inlined and so only in the former case we would safe a single object creation (a matter of nanoseconds in modern JMVs). 

However I don't have a strong opinion on this I will implement the change. In theory, it introduces complexity as you will have to test that subsequent calls return the same object.

vladbailescu (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ExperienceFragmentDataImpl.java
```diff
@@ -0,0 +1,207 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2019 Adobe
```
bpauli, line 2: please update copyright year.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -311,4 +329,55 @@ public String getMainContentSelector() {
...
+            String canonicalUrl;
+            try {
+                SeoTags seoTags = resource.adaptTo(SeoTags.class);
+                canonicalUrl = seoTags != null ? seoTags.getCanonicalUrl() : null;
+            } catch (NoClassDefFoundError ex) {
+                canonicalUrl = null;
+            }
+            this.canonicalUrl = canonicalUrl != null
+                ? canonicalUrl
+                : pathProcessor.externalize(currentPage.getPath(), request) + LinkHandler.HTML_EXTENSION;
```
bpauli, line 345: The HTML extension is already added by the LinkHandler#getPageLinkURL method.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/seo/LanguageNavigationSiteRootSelectionStrategy.java
```diff
@@ -0,0 +1,140 @@
...
+        target = "(objectClass=*DefaultSiteRootSelectionStrategy)")
+    private SiteRootSelectionStrategy defaultSelectionStrategy;
+
+    @Reference
+    private LanguageManager languageManager;
+
+    @Reference
+    private LiveRelationshipManager liveRelationshipManager;
+
+    private final Cache<Page, Optional<Resource>> languageNavigationCache = CacheBuilder.newBuilder().weakKeys().build();
```
bpauli, line 75: A cache with the page key should not be used in combination with weakKeys as this will end up in a `==` comparisons and will create new keys without needed (https://guava.dev/releases/18.0/api/docs/com/google/common/cache/CacheBuilder.html)

bpauli (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
Buuhuu: Afaik the `DefaultPathProcessor` does not call the `LinkHandler`. The `SeoTags` implementation returns urls that include the `.html` extension, thats why it is added here in the fallback case as well.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/seo/LanguageNavigationSiteRootSelectionStrategy.java
Buuhuu: This is by intent to keep the complexity small (no need for an eviction strategy, cache entry live-time or max-size etc.). It should only address cases where the service is called with the same page object multiple times, but was not supposed to introduce a long-living caching beyond the scope of a request. 

I would recommend to keep it like that as this addresses a proven performance concern and improve it when we see an actual need.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ExperienceFragmentDataImpl.java
Buuhuu: done

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
```diff
@@ -88,6 +92,11 @@
      */
     public static final String PN_MAIN_CONTENT_SELECTOR_PROP = "mainContentSelector";
 
+    /**
+     * Property name of the style property that enables/disables rendering of the alternate language links
+     */
+    public static final String PN_STYLE_RENDER_ALTERNATE_LANGUAGES = "renderAlternateLanguageLinks";
```
jckautzmann, line 98: I'd call it `PN_STYLE_RENDER_ALTERNATE_LANGUAGE_LINKS`, in sync with the repository property.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/seo/LanguageNavigationSiteRootSelectionStrategy.java
```diff
@@ -0,0 +1,147 @@
...
+import com.day.cq.wcm.api.TemplatedResource;
+import com.day.cq.wcm.msm.api.LiveRelationshipManager;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+
+/**
+ * An implementation of {@link SiteRootSelectionStrategy} that looks for a language navigation component on the given page and uses it's
+ * configured navigation root as site root.
+ * <p>
+ * This Component must be explicitly enabled by providing an (empty) configuration for it. This should be the case for sites the use the
```
jckautzmann, line 49: There seems to be y typo in the sentence:
`This should be the case for sites **the** use the`

bpauli (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/seo/LanguageNavigationSiteRootSelectionStrategy.java
Buuhuu: fixed in https://github.com/adobe/aem-core-wcm-components/pull/1647/commits/1237ae32873b8dec2d8446f4b5d8b21d69e8454b

jckautzmann (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/PageImpl.java
bpauli: The LinkHandler should used instead of calling the path processor directly. The purpose of the LinkHandler is to have unified handler to generate all links in the core components project.

PR #1649, opened 2021-06-28
DataLayer Intergration broken in 2.17.0 

comanV:
Fixes #1648
-removes `vanilla-lazyload` library

PR #1655, opened 2021-06-30
Selenium test for DataLayer Integration

comanV:
Adds Selenium test for DataLayer Integration #1649

### DIFF
--- a/testing/it/e2e-selenium/src/test/java/com/adobe/cq/wcm/core/components/it/seljup/tests/commons/DataLayerIT.java
+++ b/testing/it/e2e-selenium/src/test/java/com/adobe/cq/wcm/core/components/it/seljup/tests/commons/DataLayerIT.java
@@ -16,22 +16,24 @@
 
 package com.adobe.cq.wcm.core.components.it.seljup.tests.commons;
 
-import org.apache.http.HttpStatus;
-import org.apache.sling.testing.clients.ClientException;
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
-import org.openqa.selenium.JavascriptExecutor;
-import org.openqa.selenium.WebDriver;
-
 import com.adobe.cq.testing.selenium.pageobject.EditorPage;
 import com.adobe.cq.testing.selenium.pageobject.PageEditorPage;
 import com.adobe.cq.wcm.core.components.it.seljup.AuthorBaseUITest;
 import com.adobe.cq.wcm.core.components.it.seljup.constant.CoreComponentConstants;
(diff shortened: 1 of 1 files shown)


jckautzmann (approved):

vladbailescu (approved):

testing/it/e2e-selenium/src/test/java/com/adobe/cq/wcm/core/components/it/seljup/tests/commons/DataLayerIT.java
```diff
@@ -0,0 +1,64 @@
...
+    public void setupBeforeEach() throws ClientException {
+        testPage = authorClient.createPage("testPage-" + System.currentTimeMillis(), "Test Page Title", "/content/core-components", "/conf/core-components/settings/wcm/templates/simple-template").getSlingPath();
+        editorPage = new PageEditorPage(testPage);
+        editorPage.open();
+    }
+
+    @Test
+    public void testDataLayerInitialized() throws InterruptedException {
+        editorPage.enterPreviewMode();
+        assertTrue(isDataLayerInitialized());
```
msagolj, line 51: The check seems a bit simple and the assert is not catching many cases (e.g. when `adobeDataLayer` is `null` you get a `JavascriptException` instead). I would do a more step by step approach e.g.:

```
 public void testDataLayerInitialized() throws InterruptedException {
        editorPage.enterPreviewMode();
        Commons.switchContext("ContentFrame");
        final WebDriver webDriver = WebDriverRunner.getWebDriver();
        // check if adobeDataLayer is defined at all
        Object adobeDataLayer =  ((JavascriptExecutor) webDriver).executeScript("return window.adobeDataLayer;");
        assertNotNull(adobeDataLayer,"adobeDataLayer is undefined!");
        try {
            // check if adobeDataLayer is initialized i.e. getState() function is defined
            Object state = ((JavascriptExecutor) webDriver).executeScript("return window.adobeDataLayer.getState();");
            // check if returned state object is not empty
            assertNotNull(state, "adobeDataLayer.getState(): returned state object is null!");
            Map<?,?> stateMap = (Map)(state);
            // check some basic properties of the state object
            assertNotNull(stateMap.get("page"), "returned State is missing 'page' property!");
            assertNotNull(stateMap.get("component"), "returned State is missing 'component' property!");
            Map<?,?> components = (Map)stateMap.get("component");
            assertEquals(2,components.size(), "returned state.components should have 2 entries!");
        } catch (JavascriptException e){
            // this would return 'window.adobeDataLayer.getState() is not a function'
            fail(e);
        } catch (ClassCastException e) {
            fail("returned State object is not properly initialized!");
        }
    }
```

msagolj (approved):

PR #1686, opened 2021-07-15
[Image] Finalize image v3 component

### LINKED ISSUE #1684: [Image] Finalize image v3 component
**Edit toolbar**
Remove the Edit pencil icon that opens the in-context image editor.

**Edit Dialog**
Remove the width and height options

**Design Dialog**
Remove the threshold option
Add a resizeWidth option that is used to define the width of the image in the src attribute. This option should allow to override the defauttResizeWidth provided by OSGi, and when not set, it should default to the OSGi defauttResizeWidth.
Make that lazy loading is enabled by default when no policy is defined.
Rename "Enable lazy loading" to "Disable lazy loading", and make that checkbox unchecked by default in the dialog option, so that when opening the dialog for the first time and saving it without changes, the lazy loading stays enabled.
Remove the "Disable UUID tracking" option.
Make the "JPEG Quality" option optional.
Remove the "Features" tab

**Clean up the HTML**
Have the width and height attributes set automatically from the main image, depending on the actual dimensions of the image in the src attribute. The width should therefore be equal to the resizeWidth or to the defauttResizeWidth, and the height attribute should be set accordingly to the image aspect ratio.
As we remove the threshold capability, remove the following attribute:
data-cmp-lazythreshold
As we don't need any JS initialization anymore, we can remove the following attributes were used by the old JS script:
data-cmp-is="image"
data-cmp-src="..."
data-cmp-hook-image="image"
data-cmp-hook-image="link"
data-cmp-hook-image="noscript"
data-cmp-hook-image="map"
data-cmp-hook-image="area"
As the data layer now exists, the following attributes can be removed:
data-asset="..."
data-title="..."
Remove all parts about click areas that map to the image as these techniques aren't being used much anymore and as they'd require JavaScript to resize the area when the image is responsive. Alternatively, authors can prepare an SVG file that contains the clickable areas, and use it in our image component to display that SVG.

**Clean up the JSON**
Similarly to the HTML, we also need to cleanup the JSON:

remove srcUriTemplate
remove lazyThreshold
remove widths
remove dmImage
remove areas when it's empty
remove dataLayer xdm:tags and xdm:smartTags when either is empty

comanV:
| Fixed Issues?            | Fixes #1684
| Patch: Bug Fix?          |
| Major: Breaking Change?  | yes
| Any Dependency Changes?  |

Makes changes in the Image V3 component, as described in #1684.


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
```diff
@@ -161,14 +168,65 @@ public String getSrcset() {
...
-    public String getWidth() {
-        return width;
+    @JsonIgnore
+    public String[] getBaseImageResolution() {
+        String baseImageAbsoluteUrl = externalizer.publishLink(resourceResolver, super.getSrc());
+        URL url;
+        InputStream in;
+        BufferedImage image = null;
+        try {
+            url = new URL(baseImageAbsoluteUrl);
```
vladbailescu, line 178: We can load the file locally and compute the original width and height, no need to do a HTTP request. For DAM assets we can even leverage `new EnhancedRendition(asset.getOriginal())`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -169,6 +170,12 @@
      */
     String PN_DESIGN_DYNAMIC_MEDIA_ENABLED = "enableDmFeatures";
 
+    /**
+     * Name of the configuration policy property that the {@link AdaptiveImageServlet} will use for resizing images that don't provide a width information in their request.
+     * @since com.adobe.cq.wcm.core.components.models 12.22.0
+     */
+    String PN_RESIZE_WIDTH = "resizeWidth";
```
vladbailescu, line 177: We generally prefix policy properties with `PN_DESIGN_<name>` and use `PN_<name>` for simple/resource properties.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -316,23 +323,13 @@ default String getSrcset() {
...
+     * Returns an array with the base image width and height.
      *
-     * @return the value for the image's {@code height} attribute, if one was set, or {@code null}.
-     * @since com.adobe.cq.wcm.core.components.models 12.21.0;
+     * @return an array with the base image width and height, if image exists, or {@code new String[] {null, null}}.
+     * @since com.adobe.cq.wcm.core.components.models 13.0.0;
      */
-    default String getHeight() {
-        return null;
+    default String[] getBaseImageResolution() {
```
vladbailescu, line 331: I'd rather we use a better-specified format than `String[]`, for `EnhancedRendition` we use `java.awt.Dimension`

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
```diff
@@ -15,36 +15,44 @@
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.internal.models.v3;
 
-import java.io.UnsupportedEncodingException;
-import java.net.URLDecoder;
-import java.nio.charset.StandardCharsets;
-
-import javax.annotation.PostConstruct;
-
+import com.adobe.cq.export.json.ComponentExporter;
```
jckautzmann, line 18: Please reorganize the imports according to the AEM code conventions:
https://wiki.corp.adobe.com/display/tsg/Code+Conventions+For+Java

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
jckautzmann: The image dimensions should be saved in the repository to avoid having to compute it for each request.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
```diff
@@ -161,14 +168,65 @@ public String getSrcset() {
...
+
+    @Override
+    public boolean isLazyEnabled() {
+        return !currentStyle.get(PN_DESIGN_LAZY_LOADING_ENABLED, false);
+    }
+
+    @Override
+    @JsonIgnore
+    public String getSrcUriTemplate() {
+        return super.getSrcUriTemplate();
```
jckautzmann, line 204: is it needed to call the method from the parent class, as the method doesn't add any other logic?
Same question applies to the below methods.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -113,7 +113,7 @@
     @SuppressFBWarnings(justification = "This field needs to be transient")
     private transient AssetStore assetStore;
 
-    public AdaptiveImageServlet(MimeTypeService mimeTypeService, AssetStore assetStore, AdaptiveImageServletMetrics metrics, 
+    public AdaptiveImageServlet(MimeTypeService mimeTypeService, AssetStore assetStore, AdaptiveImageServletMetrics metrics,
```
jckautzmann, line 116: Please document the new parameters added to the constructor.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -863,6 +863,16 @@ private Integer getAllowedJpegQuality(@NotNull Resource imageResource) {
         return allowedJpegQuality;
     }
 
+    private int getResizeWidth(@NotNull Resource imageResource){
```
jckautzmann, line 866: Please add some javadocs

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
jckautzmann, line 331: Shouldn't the method name reflect that it returns the image dimensions? E.g. `getBaseImageDimension`.

content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/README.md
```diff
@@ -102,57 +95,29 @@ Publish:
...
 
-It also reuses the `core.wcm.components.image.v2.editor` editor client library category that includes JavaScript
+It also provides a `core.wcm.components.image.v3.editor` editor client library category that includes JavaScript
 handling for dialog interaction. It is already included by its edit dialog.
 
 ## BEM Description
 ```
 BLOCK cmp-image
-    ELEMENT cmp-image__link
+    ELEMENT cmp-image__linki
```
jckautzmann, line 107: is this update intentionnal?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
comanV: You're right, `null` can be returned.

content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/README.md
comanV: no, good catch

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -316,23 +323,13 @@ default String getSrcset() {
...
+     * Returns {@code new Dimension(width, height)} containing the base DAM asset's width and height (the DAM asset which is used as the {@code src} html attribute of the image).
      *
-     * @return the value for the image's {@code height} attribute, if one was set, or {@code null}.
-     * @since com.adobe.cq.wcm.core.components.models 12.21.0;
+     * @return {@code new Dimension(width, height)} containing the base DAM asset's width and height, if the image exists and is a DAM asset, or {@code new Dimension(0, 0)}.
+     * @since com.adobe.cq.wcm.core.components.models 13.0.0;
      */
-    default String getHeight() {
-        return null;
+    default Dimension getBaseDamAssetDimension() {
```
vladbailescu, line 331: I would name this `getOriginalDimension`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
```diff
@@ -15,20 +15,21 @@
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.internal.models.v3;
 
+import java.awt.*;
```
jckautzmann, line 18: Please list all classes

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
jckautzmann: @comanV Did you see the question above?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -15,13 +15,14 @@
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.models;
 
+import java.awt.*;
```
jckautzmann, line 18: Please list all the classes.

content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/_cq_design_dialog/.content.xml
```diff
@@ -98,15 +98,17 @@
...
                                         jcr:primaryType="nt:unstructured"
-                                        sling:resourceType="granite/ui/components/coral/foundation/form/checkbox"
-                                        fieldDescription="Check to disable tracking of asset's UUID."
-                                        name="./uuidDisabled"
-                                        text="Disable UUID tracking"
-                                        checked="{Boolean}true"
-                                        uncheckedValue="false"
-                                        value="{Boolean}true"/>
+                                        sling:resourceType="granite/ui/components/coral/foundation/form/numberfield"
+                                        fieldDescription="Define a resize width (in pixels). Its value will be used for resizing the width of the base images that are DAM assets. The aspect ratio of the images will be preserved. If the value is bigger than the natural width of an image, it will have no effect. Note that this setting no effect on SVG images."
```
jckautzmann, line 104: There seems to be a typo:
Note that this setting no effect on SVG images. -> "Note that this setting has no effect on SVG images."

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
comanV: The getSrcUriTemplate() and getWidths() methods need  to call the method from the parent class because their logic is relevant for Image V3 too, they are used in the getSrcset() method and in the unit tests. I agree that is not needed to call the method from the parent class for getAreas(), because we won’t use Image Maps in Image V3. I modified the method to return `null` and deprecated it.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
comanV: `java.awt` is the exception from the rule. This is the fragment from the wiki page about code conventions (https://wiki.corp.adobe.com/display/tsg/Code+Conventions+For+Java) .
`The order of the package imports is:
	1.	java.*
	2.	_blank line_
	3.	javax.*
	4.	_blank line_
	5.	org.*
	6.	_blank line_
	7.	*
	8.	_blank line_
	9.	import static *
Each group of imports is sorted alphabetically. Never use wildcard imports (import java.util.*) except for java.awt`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
comanV: Implemented `getOriginalDimension` for now to get the image dimensions only if the image is a DAM asset.

jckautzmann (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Image.java
```diff
@@ -316,23 +323,13 @@ default String getSrcset() {
...
+     * Returns {@code new Dimension(width, height)} containing the base DAM asset's width and height (the DAM asset which is used as the {@code src} html attribute of the image).
      *
-     * @return the value for the image's {@code height} attribute, if one was set, or {@code null}.
-     * @since com.adobe.cq.wcm.core.components.models 12.21.0;
+     * @return {@code new Dimension(width, height)} containing the base DAM asset's width and height, if the image exists and is a DAM asset, or {@code new Dimension(0, 0)}.
+     * @since com.adobe.cq.wcm.core.components.models 13.0.0;
      */
-    default String getHeight() {
-        return null;
+    default Dimension getOriginalDimension() {
```
vladbailescu, line 331: If the expectation is that this will always return a dimension it should be annotated as not null

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/package-info.java
```diff
@@ -34,7 +34,7 @@
  *      version, is bound to this proxy component resource type.
  * </p>
  */
-@Version("12.22.0")
+@Version("13.0.0")
```
vladbailescu, line 37: -1, we should not have breaking changes like this. Why is this needed?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
```diff
@@ -161,14 +157,70 @@ public String getSrcset() {
 
     @Nullable
```
vladbailescu, line 158: Obsolete

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
```diff
@@ -161,14 +157,70 @@ public String getSrcset() {
...
+                        return dimension;
+                    }
+                }
+            }
+        }
+        return new Dimension(0, 0);
+    }
+
+    @Override
+    public boolean isLazyEnabled() {
```
vladbailescu, line 190: Can't this be inherited from super class?

richardhand (approved):

PR #1703, opened 2021-07-28
[Accessibility][Advisory][Button] Icon text may not be appropriate

### LINKED ISSUE #1678: [Accessibility][Advisory][Button] Icon text may not be appropriate
**Steps to Reproduce**
1. With NVDA operating, navigate to the "Contact Us" button, and note announcement of "email" CSS content.

**Current Behavior**
Text alternatives for icons in buttons are provided through the CSS `content` property, which screen reader users may or may not hear announced. 

Text content in the CSS `content` property is supported by modern Chrome and Firefox browsers and JAWS and NVDA versions, but is not supported in IE with either screen reader. Announcements will vary from one assistive technology/browser combination to another. 

**Expected behavior/code**
Icons that provide meaning have textual alternatives that are consistently announced by screen readers; icons that are decorative are not announced by screen readers.

**Environment**
Chrome 91, Windows 10, NVDA 2020.4 
AEM Core Components v. 2.17  

**Affected Users**
Screen reader users

**WCAG SC**
1.1.1

**Code snippet**
```
<span class="cmp-button**icon cmp-button**icon--email"></span>
```

``{{
.t-cmp-clean .cmp-button__icon--email:before {
    content: 'email';
}
}}``

**Possible Solution**
If a text alternative for an icon is necessary to understand the purpose of a button, provide that alternative through a mechanism that has more consistent browser/assistive technology support, such as visually-hidden text. If the icon is purely decorative, it should be hidden from assistive technology using `aria-hidden="true"`.

If one of these options is not possible, ensure that any text alternative for icons added through CSS is a) appropriately descriptive to convey the purpose of the icon (so it does not interfere with understanding for those users who hear it announced) and b) not necessary to understand the purpose of the button (so those users who do not hear it announced will not lose important information).

paniciu:
- Icon text may not be appropriate
Fixes #1678

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/button/v1/button/icon.html
+++ b/content/src/content/jcr_root/apps/core/wcm/components/button/v1/button/icon.html
@@ -14,5 +14,5 @@
   ~ limitations under the License.
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
 <sly data-sly-template.icon="${@ icon}">
-    <span data-sly-test="${icon}" class="cmp-button__icon cmp-button__icon--${icon}"></span>
+    <span data-sly-test="${icon}" class="cmp-button__icon cmp-button__icon--${icon}" aria-hidden="true"></span>
 </sly>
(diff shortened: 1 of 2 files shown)


bartoszglow:
hi @paniciu . Are those the only icons that do not provide "content" CSS property and require to be ignored by screen readers?

bartoszglow (approved):

bpauli (approved):

Issue #1713, opened 2021-08-03
[List][Navigation] Display external URL redirection targets


gabrielwalt:
User story:
As an author, I want the navigation and list components to display pages that redirect to external URLs as links that point directly to the external URL, so that visitors directly navigate to the right target URL and don't get redirected.

Requirement:
When a page has a redirection target (regardless whether it is pointing to an external URL or to another AEM page), then when rendering a link to that page, the link URL should point directly to the redirection target.

Example:
* Create a page A that redirects to page B.
* Create a page C that redirects to http://google.com/
* On a page D, insert a list and/or navigation component that lists page A & C
* The respective links that are generated should then point directly to page B & http://google.com/

glasert:
@adobe export issue to Jira project SITES as Story

PR #1772, opened 2021-08-20
[List][Navigation] Display external URL redirect targets (#1713)

vladbailescu:
* Resolve redirect chain in the link handler

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
@@ -145,7 +145,10 @@ public class LinkHandler {
      */
     @NotNull
     public Optional<Link<Page>> getLink(@Nullable Page page) {
-        Page resolved = resolveRedirects(page);
+        Page resolved = page;
+        if (!isShadowingDisabled()) {
+            resolved = resolveRedirects(page);
+        }
         if (resolved == null && page != null) {
             String redirectTarget = page.getProperties().get(PageImpl.PN_REDIRECT_TARGET, "");
             if (StringUtils.isNotEmpty(redirectTarget)) {
@@ -324,9 +327,9 @@ public class LinkHandler {
      *         even {@code null} if the redirect chain does not resolve to a valid page.
      */
     @Nullable
-    private Page resolveRedirects(@Nullable final Page page) {
+    public Page resolveRedirects(@Nullable final Page page) {
         Page result = page;
-        if (page != null && !isShadowingDisabled()) {
+        if (page != null) {
             String redirectTarget;
             Set<String> redirectCandidates = new LinkedHashSet<>();
             redirectCandidates.add(page.getPath());
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
@@ -23,6 +23,8 @@ import java.util.Optional;
 import java.util.Set;
 
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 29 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -73,6 +99,11 @@
     @OSGiService
     private List<PathProcessor> pathProcessors;
 
+    /**
+     * Variable that defines how
```
richardhand, line 103: some docs missing here

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -114,11 +145,20 @@
      */
     @NotNull
     public Optional<Link<Page>> getLink(@Nullable Page page) {
-        if (page == null) {
+        Page resolved = resolveRedirects(page);
+        if (resolved == null && page != null) {
+            String redirectTarget = page.getProperties().get(PageImpl.PN_REDIRECT_TARGET, "");
```
richardhand, line 153: what about the situation where:
page A -> page B -> external URL
looks like this only handles external redirects if they're at the first level, which is the more common scenario.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -276,4 +316,51 @@ private String getPageLinkURL(@NotNull Page page) {
         return Optional.ofNullable(pageManager.getPage(path));
     }
 
+    /**
+     * Attempts to resolve the redirect chain starting form the given page, avoiding loops.
```
richardhand, line 323: starting from the given page

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
```diff
@@ -276,4 +316,51 @@ private String getPageLinkURL(@NotNull Page page) {
...
+     *
+     * @param page The starting {@link Page}
+     * @return The {@link Page} the redirect chain resolves too. Can be the original page, if no redirect target is defined or
```
richardhand, line 326: chain resolves to.

richardhand:
@vladbailescu - this is much cleaner, thanks for adding it. Added a few, mainly minor comments.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
vladbailescu: Fixed in 06d5f73

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
vladbailescu: Fixed in 06d5f73

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
vladbailescu: Fixed in 06d5f73

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
vladbailescu: I'll prepare a test case and handling for this as well

vladbailescu: Should be solved by 56f5eab

richardhand (approved):

PR #1776, opened 2021-08-25
#1775 Alt attribute removed from skip to main content link

### LINKED ISSUE #1775: [Accessibility] Page v3 skip to the main content link to the has an unnecessary alt attribute
## Bug Report

**Current Behavior**
Page v3 skip to main content link (which is an anchor) has an unnecessary alt attribute.

According to [HTML spec](https://www.w3.org/TR/html5/text-level-semantics.html#the-a-element) anchor tag is not allowed to have an `alt` attribute.

This is not present in page v2. It applies only to page v3.

**Expected behavior/code**
Page v3 skip to main content link does not have an alt attribute.

**Environment**
- Core Components version 2.17.10

bartoszglow:
Alt attribute removed from skip to main content link in page v3.

Fixes #1775


richardhand:
This was fixed in https://github.com/adobe/aem-core-wcm-components/commit/78ff7a17bfa3816632b6fdbc27eac27d5083b125

PR #1819, opened 2021-09-07
Issues/1818

### LINKED ISSUE #1818: [Link Hander] expose interface of Link Handler to be able to use it in other Models

**The issue**

**Solution**

jkoebner:
| Q                        | A 🤹‍♂️
| Fixed Issues?            | `Fixes #1818` 
| Minor:  New Feature?   | Yes

Added interface for all publicly exposed functions of LinkHandler and moved LinkHandler.class to LinkHandlerImpl.class implementing the newly created interface.

Also, refactored other classes to use the interface instead of the model directly. 

ImplementationPicker should pick the Impl class automatically.


jkoebner:
PR  #1851 will include the changes from this PR.

PR #1850, opened 2021-09-29
CIF-2300: move jackson annotations from LinkImpl to Link

Buuhuu:
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/Link.java
@@ -22,12 +22,19 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.osgi.annotation.versioning.ConsumerType;
 
+import com.adobe.cq.wcm.core.components.internal.jackson.LinkHtmlAttributesSerializer;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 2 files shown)


jckautzmann:
Not sure about this change. IIRC we've discussed it some time ago and decided to not annotate the interface and let the model implementation decide which annotation it needs.

Buuhuu:
Our particular use case it to extend the button component and add a data attribute to it in the CIF Core Components. We don't want to overlay the htl file as this will cause long term efforts adapting to upstream changes in the the WCM Core Components (which we are trying to get rid of in other places currently). The component however supports adding attributes via the button link's html attributes, which we also could extend using delegation. That would require us to implement our own `Link` implementation and annotate it the same way the `LinkImpl` is annotated (copying implementation details, esp. the `LinkHtmlAttributesSerializer`). 

Afacis there are many other interfaces annotated with Jackson annotations, why not `Link`? For example the datalayer models, the page model, and some like the Component even use `JsonSerialize` with an internal serializer implementation.

vladbailescu:
@Buuhuu , one of the reasons why we wanted the annotations to be on the implementation was to allow implementors to decide which of the getters would be the *default* for them.

Buuhuu:
> the getters would be the default for them.
If you are referring to the "url" field, then this is still possible as the annotations of the implementation overrule the ones on the interface. I added a test that verifies this behaviour in https://github.com/adobe/aem-core-wcm-components/pull/1850/commits/ef6c7d78ded33f37cc771d52a21a47bdd4c0c97f

PR #1851, opened 2021-10-01
[LinkHandler] Expose `LinkHandler` interface (#1818)

### LINKED ISSUE #1818: [Link Hander] expose interface of Link Handler to be able to use it in other Models

ky940819:
| Fixed Issues?            | Fixes #1818 

This pull request includes the commits made by @jkoebner in PR #1819. 
The commits made in that PR have been rebased to resolve conflicts, but have otherwise been unchanged and attribution is intact.

This PR expands on #1819 by:

- Removing the raw `Link` return types from the `LinkHandler` interface, and replacing them with `Link<Page>`
- Updating model implementations that use the `LinkHandler` to be compatible with this change. Other exported interfaces that return a raw `Link` are *not* modified so as to retain backwards compatibility for the time being.
- Adding additional tests for the `LinkHandlerImpl`. Specifically, tests for redirect chains and shadowing. 
- Correcting a testing bug where the resource injected into the `LinkHandlerImpl` is not the correct resource - this issue only surfaces once you start testing the ability to disable redirect shadowing in the link handler.
- Significantly reworking `LinkHandlerImpl` so as to:
  - apply consistent formatting
  - add `@Override` annotations where appropriate
  - add JavaDoc comments to all non-override methods/fields
  - make private methods static where possible
  - Update the `LinkHandlerImpl#getLink(String, String)` method to reduce duplication of code

There is one change that might be contentious. The `LinkHandlerImpl` class has been changed from field injection to constructor injection. I understand that constructor injection is not the typical pattern for this project; however, the benefit of this change is that it permits (internally) the direct instantiation of a `LinkHandlerImpl` without having to rely on a model - as long as you can satisfy the constructor requirements.
 In truth, the `LinkHandler` feels more like a service than a model to me, and so using constructor injection allows it to continue functioning as a model for backwards compatibility, but also opens up the possibility (in the future) for creating a `LinkHandlerFactory` service that can provide `LinkHandler`s as a factory service that can be used outside the context of sling models. If constructor injection is used (as it is in this PR) then the link handler _can_ function both ways.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
@@ -38,7 +38,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.adobe.cq.wcm.core.components.commons.link.Link;
-import com.adobe.cq.wcm.core.components.internal.link.LinkHandler;
+import com.adobe.cq.wcm.core.components.services.link.LinkHandler;
 import com.adobe.cq.wcm.core.components.internal.resource.CoreResourceWrapper;
 import com.adobe.cq.wcm.core.components.models.ExperienceFragment;
 import com.adobe.cq.wcm.core.components.models.Teaser;
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
@@ -46,7 +46,6 @@ import com.adobe.cq.wcm.core.components.util.ComponentUtils;
 import com.day.cq.commons.DownloadResource;
 import com.day.cq.commons.ImageResource;
 import com.day.cq.wcm.api.Page;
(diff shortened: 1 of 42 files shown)


stefanseifert:
imho this PR mixes different issues. i would recommend to split the topic of making the LinkHandler public from improving it's current implementation.

making it public is a far bigger discussions, and it should be decided first if the current API is ready for it (started a discussion item on this: #1861). this may take a bit to come to a conclusion.

improving the implementation is always a good thing and could and should be done before making it public, and is independent of this discussion.

ky940819:
@stefanseifert , I don't disagree - I think it's important that the link handler becomes usable to client bundles, but it is - as you said - two different things. 

I can pull out the changes that don't expose the interface and make that as a first request, then rebase the changes that do expose the interface into a different PR so that they can be merged (if suitable) when any discussion around that has been resolved. 

Do you have any thoughts about keeping the interface but just moving it to the implementation package so that it's not exposed and can be just moved to an exported package when the API is ready for it? Or would you rather that the interface is eliminated entirely for now?

vladbailescu:
Superseded by #2196

PR #1864, opened 2021-10-10
[LinkHandler] General improvements + add non-exported interface (#1863)

### LINKED ISSUE #1863: [LinkHandler] Refactor / improve `LinkHandler`
## Feature Request

This issue is not a bug fix - just optimizations / improvements:

1. The current `LinkHandler` model does not implement an interface. #1818 requests that an interface be made public; however, until it is made public a non-exported interface should exist so that when it is time to make the interface public it does not require a major refactoring to accomplish.
2. The link handler model implementation contains some unnecessary code duplication that can be eliminated.
3. The link handler model implementation - in some cases - computes fall-back values before knowing that they are required, thus incurring the overhead even when they are not needed. 
4. The link handler model implementation returns the raw type `Link` where that generic type should be parameterized as `Link<Page>`. The ambiguity in the API does not provide any benefit and the weakness of the interface (once exposed) will lead to confusion by clients trying to make use of it, and to possible breaking-changes that cannot be expressed via the API. 
5. The test coverage for the `LinkHandlerImpl` is incomplete - it does not test linking to redirect pages / chains of redirect pages with/without shadowing.
6. A silent bug exists in the `LinkHandlerImpl` tests that results in the incorrect resource being injected into the test models - this bug only surfaces when testing testing link redirect shadowing.

1. Create an interface `LinkHandler` and rename the current link handler model implementation to `LinkHandlerImpl`. The new interface should not (at this time) be exported from the bundle (see #1818  
2. Refactor `LinkHandlerImpl` to reduce code duplication
3. Refactor `LinkHanlderImpl` to only compute fall-back value if needed.
4. Change the `LinkHandler` to return `Link<Page>` instead of the raw type `Link` - do not update any exported model interfaces that already use the raw type so that this change is non-breaking. 
5. Improve test coverage prior to any refactoring of the link handler implementation so as to be able to reliably assert no break in functionality due to the refactoring / rewriting. 
6. Correct the testing bug and ensure that the resource injected into the link model is the link resource, and not the page content resource.

**Are there alternatives?**
No

**Documentation**
N/A - no functional change

ky940819:
| Fixed Issues?            | Fixes #1863

This pull request is in response to the comments left on PR #1851 . 
Functionally, this is the same PR with the key exception of _not_ exporting the interface.

- `LinkHandler` interface is added (but not exported). The old implementation is renamed to `LinkHandlerImpl`
- Removing the raw `Link` return types from the `LinkHandler` interface, and replacing them with `Link<Page>`
- Updating model implementations that use the `LinkHandler` to be compatible with this change (where required). Other exported interfaces that return a raw `Link` are *not* modified so as to retain backwards compatibility for the time being.
- Adding additional tests for the `LinkHandlerImpl`. Specifically, tests for redirect chains and shadowing. 
- Correcting a testing bug where the resource injected into the `LinkHandlerImpl` is not the correct resource - this issue only surfaces once you start testing the ability to disable redirect shadowing in the link handler.
- Significantly reworking `LinkHandlerImpl` so as to:
  - apply consistent formatting
  - add `@Override` annotations where appropriate
  - add JavaDoc comments to all non-override methods/fields
  - make private methods static where possible
  - Update the `LinkHandlerImpl#getLink(String, String)` method to reduce duplication of code
  - Avoid computing fall-back values before they are needed,

There is one change that might be contentious. The `LinkHandlerImpl` class has been changed from field injection to constructor injection. I understand that constructor injection is not the typical pattern for this project; however, the benefit of this change is that it permits (internally) the direct instantiation of a `LinkHandlerImpl` without having to rely on a model - as long as you can satisfy the constructor requirements.

 In truth, the `LinkHandler` feels more like a service than a model to me, and so using constructor injection allows it to continue functioning as a model for backwards compatibility, but also opens up the possibility (in the future) for creating a `LinkHandlerFactory` service that can provide `LinkHandler`s as a factory service that can be used outside the context of sling models. If constructor injection is used (as it is in this PR) then the link handler _can_ function both ways (i.e. instantiated by the sling model service, or manually by a service).

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
@@ -46,7 +46,6 @@ import com.adobe.cq.wcm.core.components.util.ComponentUtils;
 import com.day.cq.commons.DownloadResource;
 import com.day.cq.commons.ImageResource;
 import com.day.cq.wcm.api.Page;
-import com.day.cq.wcm.api.PageManager;
 import com.day.cq.wcm.api.Template;
 import com.day.cq.wcm.api.designer.Designer;
 import com.day.cq.wcm.api.designer.Style;
@@ -323,7 +322,7 @@ public class Utils {
 
         if (imageFromPageImage) {
             Resource inheritedResource = null;
-            Optional<Link> link = linkHandler.getLink(resource);
+            Optional<Link<Page>> link = linkHandler.getLink(resource);
             boolean actionsEnabled = (currentStyle != null) ?
                     !currentStyle.get(Teaser.PN_ACTIONS_DISABLED, !properties.get(Teaser.PN_ACTIONS_ENABLED, false)) :
                     properties.get(Teaser.PN_ACTIONS_ENABLED, false);
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkHandler.java
@@ -32,8 +32,10 @@ import java.util.Optional;
 @ConsumerType
 public interface LinkHandler {
 
+    /**
+     * Page extension.
+     */
     String HTML_EXTENSION = ".html";
-
 
     /**
      * Name of the resource property that for redirecting pages will indicate if original page or redirect target page should be returned.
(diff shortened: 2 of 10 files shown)


vladbailescu:
@ky940819 , it would have been so much easier to review the changes if you moved the original `LinkHandler` into `LinkHandlerImpl` and add the interface as a new file.

vladbailescu:
Superseded by #2196

Issue #1884, opened 2021-11-01
[DMAssetPostProcessor] Property 'imageServerUrl' is created on every resource where a property 'fileReference' is saved


stefanseifert:
## Bug Report

**Current Behavior**

The current Dynamic Media implementation contains a [DMAssetPostProcessor](https://github.com/adobe/aem-core-wcm-components/blob/development/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/DMAssetPostProcessor.java) that is implemented as SlingPostProcessor.

It listenes to *every* POST that is done on the AEM instance, and kicks in if it finds a changed property `fileReference` in the updated properties *whatever component or action has triggered this POST*. if the referenced asset is a dynamic media asset, it stores an additional `imageServerUrl` property in each and every content resource where that asset is referenced.

The result is that any component, also custom components, that use a `fileReference` property in their edit dialog get this additional property written into the resource. This is completely unexpected and usually not what is desired.

In general, it is *very bad practice* to write instance/domain-specific stuff directly into the content. If this content is transferred to other environments all those properties might need to be updated because the target environment might be configured to a different DM catalog using a different host name. Additionally, this host name is already stored in each DM asset itself and can just be read from there (also on publish instances).

**Expected behavior/code**

Ideally, the implementation of the core component's DM support should be updated to get completely rid of the DMAssetPostProcessor, e.g. reading the DM host name from the references asset.

If this is not possible or desired, the DMAssetPostProcessor should be updated to handle only posts for a configurable list of resource types, by default only the resource types of the core components that have DM support enabled.

**Environment**
- Core Components 2.17.12

bpauli:
@adobe export issue to Jira project SITES as Bug

PR #1904, opened 2021-12-02
Updated Image component DM url creation logic to get rid of the 'imageServerUrl' property dependency generated by DMAssetPostProcessor

shubhanshu23:
Fixes the Issue mentioned on : https://github.com/adobe/aem-core-wcm-components/issues/1884

"If content is transferred to other AEM environments all those properties might need to be updated because the target environment might be configured to a different DM catalog using a different host name. Additionally, this host name is already stored in each DM asset itself and can just be read from there (also on publish instances)."

| Fixed Issues?            | Fixes https://github.com/adobe/aem-core-wcm-components/issues/1884 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/ImageImpl.java
@@ -176,17 +176,20 @@ public class ImageImpl extends com.adobe.cq.wcm.core.components.internal.models.
                         dmImage = true;
                         //check for publish side
                         boolean isWCMDisabled =  (com.day.cq.wcm.api.WCMMode.fromRequest(request) == com.day.cq.wcm.api.WCMMode.DISABLED);
+                        //sets to '/is/image/ or '/is/content' based on dam:scene7Type property
+                        String dmServerPath;
+                        if (asset.getMetadataValue(Scene7Constants.PN_S7_TYPE).equals(Scene7AssetType.ANIMATED_GIF.getValue())) {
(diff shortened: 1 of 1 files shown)


stefanseifert:
this changes the logic in ImageImpl to no longer read the imageServerUrl property from the components content.

but [DMAssetPostProcessor](https://github.com/adobe/aem-core-wcm-components/blob/development/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/DMAssetPostProcessor.java) is still - in place and writes this properties everywhere - can we just delete this class with the new logic in place?

shubhanshu23:
@stefanseifert , we can not remove the DMAssetPostProcessor entirely as it is responsible to update the 'smartCropRenditions' as well as per Javadoc 
"Additionally this post processor monitors drag and drop actions by checking if fileReference update is accompanied by file@Delete update. In these cases it resets the value of smartCropRendition." property as well.
What I can do is remove the logic to update 'imageServerUrl' in DMAssetPostProcessor.

stefanseifert:
imho the DMAssetPostProcessor should be removed completely. the logic about smartCropRendition is only to remove it, if another file was drag&dropped to the component? for this, it has be [added here](https://github.com/adobe/aem-core-wcm-components/blob/6388d2212c3050b8f8cf6750bc5759f8bd4f7fdd/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/_cq_editConfig.xml#L10-L16) (it's there already it seems). probably check the other v1 and v3 of the image component as well.

as the post processor is configured currently it may have lot's of unwanted side effect as it does not actually check on which resource/resource type it operates, it just blindly checks for certain property names.

PR #1912, opened 2021-12-13
Updated dmServerUrl creation logic to remove dependency on 'imageServ…

### LINKED ISSUE #1884: [DMAssetPostProcessor] Property 'imageServerUrl' is created on every resource where a property 'fileReference' is saved

bpauli:
…erUrl' property (#1904)

Issue : https://github.com/adobe/aem-core-wcm-components/issues/1884

| Fixed Issues?            | Fixes #1884 
| Any Dependency Changes?  |


shubhanshu23:
Hi @bpauli , 'dam:scene7Domain' property is save with a '/' in metadata, this why I added .substring(1) while doing the concatenation in  #1904

PR #1926, opened 2022-01-10
[SITES-3860] - Add cmp:click event to Text Component

### LINKED ISSUE #1925: [Text] Add cmp:click event to Text Component
Without the cmp:click event there is no way to track links that have been added to the text.
The event should pass the actual link clicked because there could be more than one anchor link within the Text component..

comanV:
Fixes #1925

### DIFF
--- a/content/src/scripts/datalayer/v1/datalayer.js
+++ b/content/src/scripts/datalayer/v1/datalayer.js
@@ -47,6 +47,20 @@
         var element = event.currentTarget;
         var componentId = getClickId(element);
 
+        // add clicked link contained by the Text Component to Data Layer
+        if (element.classList.contains("cmp-text")) {
+            if (event.target.tagName === "A") {
+                dataLayer.push({
+                    event: "cmp:click",
+                    eventInfo: {
+                        path: "component." + componentId,
+                        link: event.target.getAttribute("href")
+                    }
+                });
+            }
+            return;
+        }
+
         dataLayer.push({
             event: "cmp:click",
             eventInfo: {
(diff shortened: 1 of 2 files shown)


content/src/scripts/datalayer/v1/datalayer.js
```diff
@@ -47,6 +47,20 @@
         var element = event.currentTarget;
         var componentId = getClickId(element);
 
+        // add clicked link contained by the Text Component to Data Layer
```
jckautzmann, line 50: In order to keep the logic of the `datalayer.js` agnostic of specific components, we might as well place this logic as part of the Text component, as we do when managing the `show` and `hide` events of the Accordion component [0].
@comanV @vladbailescu @bartoszglow WDYT?

[0] https://github.com/adobe/aem-core-wcm-components/blob/development/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js#L367

comanV: I agree, I think it's better.

content/src/content/jcr_root/apps/core/wcm/components/text/v2/text/clientlibs/site/js/text.js
```diff
@@ -0,0 +1,84 @@
...
+     *
+     * @private
+     * @param {HTMLElement} item, the Text item
+     * @returns {String} dataLayerId or undefined
+     */
+    function getDataLayerId(item) {
+        if (item && item.dataset.cmpDataLayer) {
+            return Object.keys(JSON.parse(item.dataset.cmpDataLayer))[0];
+        } else {
+            return item.id;
```
jckautzmann, line 66: the item might be null

comanV: Addressed in c1428f34751b405a731b3630b1dc7a4005ed6a61.

jckautzmann (approved):

PR #1927, opened 2022-01-12
[Image] issue with adaptive renditions of small tiff images

### LINKED ISSUE #1911: [Image] issue with adaptive renditions of small tiff images
## Bug Report

**Current Behavior**
When using tiff images in the Image Core Component the AdaptiveImageServlet will convert this into jpeg renditions according to the widths configured in the Image component policy.  This works fine when the original tiff image is large enough, but when the requested rendition width is bigger than the original image this breaks.  

For example, in the We.Retail example site, the image policy has the following widths configured: 128, 256, 512, 1024, 1280, 1440, 1920, 2048.  If you upload a small image to AEM (example [autumn.tif.zip](https://github.com/adobe/aem-core-wcm-components/files/7694825/autumn.tif.zip), dimensions: 345 × 206) and drag this onto an Image component in the We.Retail site, the image component will only render widths that are smaller than the original width (so in the example only 128 and 256) but larger renditions (e.g. 512 or higher) fail and return a jpeg file of 0 bytes.

Example URL that works for me: http://localhost:4502/content/we-retail/language-masters/en/about-us/_jcr_content/root/responsivegrid/image.coreimg.82.256.jpeg/1639159783293/autumn.jpeg

Example URL that fails for me: http://localhost:4502/content/we-retail/language-masters/en/about-us/_jcr_content/root/responsivegrid/image.coreimg.82.512.jpeg/1639159783293/autumn.jpeg

The problem seems to be that when a larger rendition is requested, the AdaptiveImageServlet is reverting to the original file, but this is then not converted into a jpeg rendition.

**Expected behavior/code**
I would expect that AdaptiveImageServlet to fallback on the largest available jpeg rendition (typically the 1280x1280 web rendition) instead of the tiff original.

**Environment**
Tested on:
- AEM 6.5.0.11 **and** AEMaaCS SDK 2021.11.6058.20211123T163652Z
- Core Components version 2.3.2 **and** 2.17.12
- Java version 8 **and** 11

**Possible Solution**

vladbailescu:
* Attempt to convert unchanged renditions if the rendition type and desired types are not the same

Fixes #1911


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -606,6 +635,29 @@ private EnhancedRendition filter(@NotNull EnhancedRendition rendition) throws IO
         throw new IOException(String.format("Cannot process rendition %s due to size %s", rendition.getName(), rendition.getDimension()));
     }
 
+    private void streamOrConvert(@NotNull SlingHttpServletResponse response, @NotNull EnhancedRendition rendition, @NotNull String imageType,
+                                 String imageName, int resizeWidth, double quality) throws IOException {
+        Dimension dimension = rendition.getDimension();
+        if (rendition.getMimeType().equals(imageType)) {
+            LOGGER.debug("Found rendition {}/{} has a width of {}px and does not require a resize for requested width of {}px",
+                    rendition.getAsset().getPath(), rendition.getName(), dimension != null ? dimension.getWidth() : null, resizeWidth);
+            stream(response, rendition.getStream(), imageType, imageName);
```
kwin, line 644: it would be better to use try..with...resources on the stream and close it where it has been opened (instead of closing it in the method stream (to which is is passed as parameter)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -492,12 +506,21 @@ private void resizeAndStreamLayer(SlingHttpServletResponse response, Layer layer
      *
      * @param rendition - the rendition
      * @return a layer for the rendition
-     * @throws IOException if a {@link Layer} cannot be created for the given rendition
+     *
+     * @return {@code null} if the rendition is not supported
      */
-    @NotNull
-    private Layer getLayer(@NotNull EnhancedRendition rendition) throws IOException {
+    @Nullable
```
vladbailescu, line 512: Added `null` return to avoid throwing an NPE when `assetHandler.getImage` was returning null (for image types that cannot be processed).

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -329,45 +335,53 @@ private void transformAndStreamAsset(SlingHttpServletResponse response, ValueMap
...
-                    Rectangle scaledRectangle = new Rectangle(
-                            (int) (rectangle.x * scaling),
-                            (int) (rectangle.y * scaling),
-                            (int) (rectangle.getWidth() * scaling),
-                            (int) (rectangle.getHeight() * scaling)
-                    );
-                    layer.crop(scaledRectangle);
-                } else {
-                    layer.crop(rectangle);
+                if (layer != null) {
```
vladbailescu, line 338: Added null-check because `getLayer` can return null, rest is unchanged.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -329,45 +335,53 @@ private void transformAndStreamAsset(SlingHttpServletResponse response, ValueMap
...
+                        layer.crop(rectangle);
+                    }
+                    appliedTransformation = true;
                 }
-                appliedTransformation = true;
             }
             if (rotationAngle != 0) {
                 if (layer == null) {
-                    layer = getLayer(getBestRendition(asset, resizeWidth));
+                    layer = getLayer(getBestRendition(asset, resizeWidth, imageType));
```
vladbailescu, line 355: Added null-check because `getLayer` can return null, rest is unchanged.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -329,45 +335,53 @@ private void transformAndStreamAsset(SlingHttpServletResponse response, ValueMap
...
+                }
+                if (layer != null) {
```
vladbailescu, line 357: Added null-check because `getLayer` can return null, rest is unchanged.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -329,45 +335,53 @@ private void transformAndStreamAsset(SlingHttpServletResponse response, ValueMap
...
-                layer.rotate(rotationAngle);
-                LOGGER.debug("Applied rotation transformation ({} degrees).", rotationAngle);
-                appliedTransformation = true;
             }
             if (flipHorizontally) {
                 if (layer == null) {
-                    layer = getLayer(getBestRendition(asset, resizeWidth));
+                    layer = getLayer(getBestRendition(asset, resizeWidth, imageType));
+                }
+                if (layer != null) {
```
vladbailescu, line 367: Added null-check because `getLayer` can return null, rest is unchanged.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -329,45 +335,53 @@ private void transformAndStreamAsset(SlingHttpServletResponse response, ValueMap
...
-                layer.flipHorizontally();
-                LOGGER.debug("Flipped image horizontally.");
-                appliedTransformation = true;
             }
             if (flipVertically) {
                 if (layer == null) {
-                    layer = getLayer(getBestRendition(asset, resizeWidth));
+                    layer = getLayer(getBestRendition(asset, resizeWidth, imageType));
+                }
+                if (layer != null) {
```
vladbailescu, line 377: Added null-check because `getLayer` can return null, rest is unchanged.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -378,41 +392,43 @@ private void transformAndStreamAsset(SlingHttpServletResponse response, ValueMap
                     int resizeHeight = calculateResizeHeight(originalWidth, originalHeight, resizeWidth);
                     if (resizeHeight > 0 && resizeHeight != originalHeight) {
                         layer = getLayer(rendition);
-                        if (layer.getBackground().getTransparency() != Transparency.OPAQUE &&
-                                ("jpg".equalsIgnoreCase(extension) || "jpeg".equalsIgnoreCase(extension))) {
-                            LOGGER.debug("Adding default (white) background to a transparent PNG: {}/{}", asset.getPath(),
-                                    rendition.getName());
-                            layer.setBackground(Color.white);
+                        if (layer != null) {
```
vladbailescu, line 395: Added null-check because `getLayer` can return null, rest is unchanged.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -606,16 +641,64 @@ private EnhancedRendition filter(@NotNull EnhancedRendition rendition) throws IO
...
                         String imageName)
             throws IOException {
         response.setContentType(contentType);
         response.setHeader("Content-Disposition", "inline; filename=" + URLEncoder.encode(imageName, CharEncoding.UTF_8));
-        try {
-            IOUtils.copy(inputStream, response.getOutputStream());
-        } finally {
-            IOUtils.closeQuietly(inputStream);
-        }
+        IOUtils.copy(inputStream, response.getOutputStream());
```
vladbailescu, line 701: We're always using try-with-resources outside this method.

msagolj (approved):

PR #1932, opened 2022-01-17
Use ExternalizableInputStream

### LINKED ISSUE #1931: DownloadServlet and AdaptiveImageServlet should support ExternalizableInputStream
[SLING-7140](https://issues.apache.org/jira/browse/SLING-7140) introduced an API for delivering untransformed binaries directly from external servers instead of spooling them through AEM. That mechanism should be supported in [DownloadServlet](https://github.com/adobe/aem-core-wcm-components/blob/c6756447df4ebcfd66d133385502cae3135f87be/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/DownloadServlet.java#L184-L203) (for all cases) and in [AdaptiveImageServlet](https://github.com/adobe/aem-core-wcm-components/blob/c6756447df4ebcfd66d133385502cae3135f87be/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L575-L585) (in case no transformation is necessary). 
The implementation just requires an instanceof check with `ExternalizableInputStream` like implemented in https://github.com/apache/sling-org-apache-sling-servlets-get/blob/edfaf307e4257dd34cc5b71121dfb8dfc43c12cb/src/main/java/org/apache/sling/servlets/get/impl/helpers/StreamRenderer.java#L159-L162.

The necessary Sling API version is shipped with the minimum required AEM Version 6.4.8.

kwin:
| Fixed Issues?            | Fixes #1931
| Minor: New Feature?     | Yes


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
```diff
@@ -297,7 +298,9 @@ private void transformAndStreamAsset(SlingHttpServletResponse response, ValueMap
         String extension = mimeTypeService.getExtension(imageType);
         if ("gif".equalsIgnoreCase(extension) || "svg".equalsIgnoreCase(extension)) {
             LOGGER.debug("GIF or SVG asset detected; will render the original rendition.");
-            stream(response, asset.getOriginal().getStream(), imageType, imageName);
+            try (InputStream inputStream = asset.getOriginal().getStream()) {
```
kwin, line 301: This leads to false positive in spotbugs with Java 11 (https://github.com/spotbugs/spotbugs/issues/1338).

vladbailescu:
Let's wait to have #1927 merged before we revisit this!

kwin:
@vladbailescu time to revisit now?

PR #1953, opened 2022-01-25
[Teaser] Enable inheritance by default

### LINKED ISSUE #1896: [Teaser] Enable inheritance by default
When adding a new teaser (v2) to a page, the different inheritance features of the edit dialog are disabled by default. For an author, it be more efficient if those were enabled by default, so that only the link to a page must be configured, and (provided that page defines a page image) the teaser would then automatically display the page title, description, image (and alt text).

bpauli:
- enable title and description from page per default
- fix issue with fallback to current page to get title and description if no link or action is provided

| Fixed Issues?            | Fixes #1896 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | 
| Any Dependency Changes?  |


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/TeaserImpl.java
```diff
@@ -57,6 +83,49 @@ public String getLinkURL() {
...
+
+    @Override
+    public String getDescription() {
+        if (this.description == null && !this.descriptionHidden) {
+            if (descriptionFromPage) {
+                this.description = this.getTargetPage().map(Optional::of).orElseGet(() -> Optional.ofNullable(getCurrentPage()))
+                        .map(Page::getDescription)
+                        // page properties uses a plain text field - which may contain special chars that need to be escaped in HTML
+                        // because the resulting description from the teaser is expected to be HTML produced by the RTE editor
+                        .map(Text::escapeXml)
```
vladbailescu, line 120: Shouldn't we use https://sling.apache.org/apidocs/sling9/org/apache/sling/xss/XSSAPI.html#encodeForHTML-java.lang.String- instead?

vladbailescu:
LGTM, expect for HTML encoding.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/TeaserImpl.java
jckautzmann: Doesn't Sightly take care of escaping the string in the HTML file? I wouldn't escape the string here as we don't know the context: it could be used in JSON and would need a different type of escaping.

vladbailescu: @jckautzmann , @bpauli mentioned that normally we expect the description to be edited with the RTE (which already does escape) so we skip HTL escaping. I too would love to be able to just rely on HTL, it's much safer.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/TeaserImpl.java
```diff
@@ -57,6 +83,49 @@ public String getLinkURL() {
...
+            }
+        }
+        return title;
+    }
+
+    @Override
+    public String getDescription() {
+        if (this.description == null && !this.descriptionHidden) {
+            if (descriptionFromPage) {
+                this.description = this.getTargetPage().map(Optional::of).orElseGet(() -> Optional.ofNullable(getCurrentPage()))
```
jckautzmann, line 116: the logic here should be the same as for the title: in case there is no target page, the description should be the one of the first CTA. If no CTA is defined, it should take the description of the current page.

jckautzmann: Please make sure the logic is the same as for the link:
```
        link = this.getTargetPage()
                .map(page -> Optional.of(linkHandler.getLink(page.getPath(), linkTarget).orElse(null)))
                .orElseGet(() -> {
                    // target page doesn't exist
                    if (this.isActionsEnabled()) {
                        return this.getActions().stream().findFirst()
                                .map(action -> Optional.ofNullable(linkHandler.getLink(action.getURL(), null).orElse(null)))
                                .orElse(Optional.empty());
                    } else {
                        // use the property value if actions are not enabled
                        return Optional.ofNullable(linkHandler.getLink(resource, Link.PN_LINK_URL).orElse(null));
                    }
                });

```

vladbailescu: @jckautzmann , `getTargetPage` already tries to get the first page from the CTAs. For the link we revisit the CTAs/actions because there might be an external/non-page link.

jckautzmann (approved):

jckautzmann:
@vladbailescu Forgot to say: shouldn't we add/update our Selenium tests?

bpauli:
@vladbailescu The link field get disabled in the dialog at the moment when you check the "Don't provide an alternative text" in the Asset tab. I think this needs to be fixed.
image

image

vladbailescu:
@jckautzmann , tests were updated in 658a143
@bpauli , that was a side-effect of importing the Image edit clientlibs, should be fixed by 504ca37

vladbailescu (approved):

PR #1963, opened 2022-01-31
[Teaser] Improve HTML semantics: display the content before the image

### LINKED ISSUE #906: [Teaser] Improve HTML semantics: display the content before the image
The teaser component lists the HTML order as image, title, description, CTA - with the title being a heading.

Semantically, headings are designed to start a content. Screen-reader users use headings for navigation so they would start a section with related child content underneath them (in this case, the image).

The markup should place the title first. If users want to reposition the image, they can do so with CSS.

There is a great article here for reference re: the future of the teaser component in general, including options to make the whole card clickable, etc. (https://inclusive-components.design/cards/) 

NB: I've referenced this as a comment in another thread, but this is standalone for easy locating.

comanV:
Fixes #906


jckautzmann:
The changes should be done on Teaser v2, we should not create a new Teaser v3 version.

comanV:
> The changes should be done on Teaser v2, we should not create a new Teaser v3 version.

There are changes in the html of the component. Now the `cmp-teaser__image` div comes after the `cmp-teaser__content` one.

jckautzmann (approved):

PR #2030, opened 2022-02-22
Table of Contents Component

vsarya:
| Fixed Issues?            | https://github.com/adobe/aem-core-wcm-components/wiki/Table-of-Content-Component
| Tests Added + Pass?      | UTs  and UI Selenium tests added
| Any Dependency Changes?  | Yes (JSOUP)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableOfContentsImpl.java
```diff
@@ -0,0 +1,123 @@
...
+    private String[] ignoreClasses;
+
+    @PostConstruct
+    private void initModel() {
+        restrictListType = currentStyle.get(PN_RESTRICT_LIST_TYPE, String.class);
+        restrictStartLevel = currentStyle.get(PN_RESTRICT_START_LEVEL, String.class);
+        restrictStopLevel = currentStyle.get(PN_RESTRICT_STOP_LEVEL, String.class);
+        includeClasses = currentStyle.get(PN_INCLUDE_CLASSES, String[].class);
+        ignoreClasses = currentStyle.get(PN_IGNORE_CLASSES, String[].class);
+        slingHttpServletRequest.setAttribute("contains-table-of-contents", true);
```
jckautzmann, line 86: I'd prefix with `cmp` to avoid clashes with custom names: `cmp:contains-table-of-contents`

jckautzmann, line 86: I'd prefix with `cmp` to avoid clashes with custom attributes: `cmp:contains-table-of-contents`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/CharResponseWrapper.java
```diff
@@ -0,0 +1,39 @@
...
+ ~ limitations under the License.
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+package com.adobe.cq.wcm.core.components.internal.servlets;
+
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpServletResponseWrapper;
+import java.io.CharArrayWriter;
+import java.io.PrintWriter;
+
+public class CharResponseWrapper extends HttpServletResponseWrapper {
```
jckautzmann, line 23: Please add some javadocs explaining why it is needed.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/CharResponseWrapper.java
```diff
@@ -0,0 +1,39 @@
...
+    private CharArrayWriter writer;
+
+    public CharResponseWrapper(HttpServletResponse response) {
+        super(response);
+        writer = new CharArrayWriter();
+    }
+
+    public PrintWriter getWriter() {
```
jckautzmann, line 31: Please add some javadocs

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/CharResponseWrapper.java
```diff
@@ -0,0 +1,39 @@
...
+        return new PrintWriter(writer);
+    }
+
+    public String toString() {
```
jckautzmann, line 35: Please add some javadocs

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -0,0 +1,231 @@
...
+    property = {Constants.SERVICE_RANKING + "Integer=999"}
+)
+@SlingServletFilter(
+    scope = {SlingServletFilterScope.REQUEST},
+    pattern = "/content/.*",
+    resourceTypes = "cq:Page",
+    extensions = {"html"},
+    methods = {"GET"}
+)
+public class TableOfContentsFilter implements Filter {
```
jckautzmann, line 57: Please add some javadocs explaining how the filter works.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -0,0 +1,231 @@
...
+
+            response.setContentLength(alteredContent.length());
+            response.getWriter().write(alteredContent);
+        } else {
+            response.setContentLength(originalContent.length());
+            response.getWriter().write(originalContent);
+        }
+    }
+
+    private Element getTableOfContents(Element tocPlaceholderElement) {
```
jckautzmann, line 114: Please add some javadocs explaining what the method does.

jckautzmann:
It would be nice to add a few Selenium tests.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/CharResponseWrapper.java
vladbailescu, line 23: Will this work across multiple encodings?

bundles/core/pom.xml
```diff
@@ -71,6 +72,7 @@
                             com.adobe.cq.commerce.*;resolution:=optional,
                             com.adobe.aem.wcm.seo.*;resolution:=optional,
                             javax.annotation;version=0.0.0,
+                            javax.annotation.meta;resolution:=optional,
```
vladbailescu, line 75: Is this used?

jckautzmann:
Please add a README for this component.

jckautzmann:
Please add a page to the component library with various different configurations of the TOC.

content/src/content/jcr_root/apps/core/wcm/components/tableofcontents/v1/tableofcontents/_cq_dialog/.content.xml
```diff
@@ -0,0 +1,148 @@
...
+                                                fieldLabel="ID"
+                                                name="./id"/>
+                                        </items>
+                                    </column>
+                                </items>
+                            </columns>
+                        </items>
+                    </properties>
+                </items>
+            </tabs>
```
jckautzmann, line 145: Please add a style tab for the edit dialog. See e.g.: https://github.com/adobe/aem-core-wcm-components/blob/main/content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/_cq_dialog/.content.xml#L253-L256

content/src/content/jcr_root/apps/core/wcm/components/tableofcontents/v1/tableofcontents/tableofcontents.html
```diff
@@ -0,0 +1,28 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
+<div data-sly-use.toc="com.adobe.cq.wcm.core.components.models.TableOfContents"
+     data-sly-use.component="com.adobe.cq.wcm.core.components.models.Component"
+     class = "table-of-contents-placeholder"
```
jckautzmann, line 18: See Vlad's remark above: the class name should be `cmp-toc__placeholder`

content/src/content/jcr_root/apps/core/wcm/components/tableofcontents/v1/tableofcontents/tableofcontents.html
```diff
@@ -0,0 +1,28 @@
...
+     id = "${component.id}"
+     data-list-type = "${toc.listType}"
```
jckautzmann, line 20: I'd prefix the data attribute with `cmp-toc-`:
```
data-cmp-toc-list-type
data-cmp-toc-start-level
data-cmp-toc-include-classes
data-cmp-toc-ignore-classes
```

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -0,0 +1,231 @@
...
+            (containsTableOfContents != null && containsTableOfContents)) {
+
+            Document document = Jsoup.parse(originalContent);
+
+            Elements tocPlaceholderElements = document.getElementsByClass("table-of-contents-placeholder");
+            for (Element tocPlaceholderElement : tocPlaceholderElements) {
+                Element tableOfContents = getTableOfContents(tocPlaceholderElement);
+                tocPlaceholderElement.empty();
+                tocPlaceholderElement.clearAttributes();
+                tocPlaceholderElement.addClass("table-of-contents");
```
jckautzmann, line 89: Please rename it by following the BEM style. E.g.: `cpm-toc__content`

jckautzmann:
It would be nice to increase the code coverage for `TableOfContentsFilter.java` as it is the core of the TOC processing. Current numbers:
- missed instructions: 92%
- missed branches: 76%

jckautzmann:
You are following below pattern when nesting lists:
```
<ul>
    <li>List item one</li>
    <li>List item two with subitems:</li>
    <ul>
        <li>Subitem 1</li>
        <li>Subitem 2</li>
    </ul>
    <li>Final list item</li>
</ul>
```

According to standard practices [0] it should be as follows:
```
<ul>
    <li>List item one</li>
    <li>List item two with subitems:
        <ul>
            <li>Subitem 1</li>
            <li>Subitem 2</li>
        </ul>
    </li>
    <li>Final list item</li>
</ul>
```

@vladbailescu @gabrielwalt : what is your take on this?

[0] https://stackoverflow.com/questions/5899337/proper-way-to-make-html-nested-list

jckautzmann:
Following use case triggers an error:
- edit dialog: start = 2, stop = 3
- design dialog: start = 4
Error:
```
java.lang.IllegalArgumentException: String must not be empty
	at org.jsoup.helper.Validate.notEmpty(Validate.java:94) [com.adobe.cq.core.wcm.components.core:2.18.1.SNAPSHOT]
	at org.jsoup.select.Selector.select(Selector.java:94)
	at org.jsoup.nodes.Element.select(Element.java:421) [com.adobe.cq.core.wcm.components.core:2.18.1.SNAPSHOT]
	at com.adobe.cq.wcm.core.components.internal.servlets.TableOfContentsFilter.getTableOfContents(TableOfContentsFilter.java:144) [com.adobe.cq.core.wcm.components.core:2.18.1.SNAPSHOT]
	at com.adobe.cq.wcm.core.components.internal.servlets.TableOfContentsFilter.doFilter(TableOfContentsFilter.java:86) [com.adobe.cq.core.wcm.components.core:2.18.1.SNAPSHOT]
```

jckautzmann:
Having to refresh the page to get the updated TOC is not a nice experience for an author. IMHO the page should be automatically refreshed when the TOC is updated.
cc: @vladbailescu @gabrielwalt

bundles/core/pom.xml
vsarya: No, only `javax.annotation.*` is used in this repo and only `javax.annotation.PostConstruct` is used in this PR.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/CharResponseWrapper.java
vsarya: This wrapper just overrides the `writer` of the original response. Rest everything is handled by the original `response` object.
`CharResponseWrapper` extends `HttpServletResponseWrapper` which further extends `ServletResponseWrapper` which has the method `setCharacterEncoding`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+        throws IOException, ServletException {
+
+        CharResponseWrapper responseWrapper = new CharResponseWrapper((HttpServletResponseWrapper) response);
+        chain.doFilter(request, responseWrapper);
+        String originalContent = responseWrapper.toString();
+        Boolean containsTableOfContents = (Boolean) request.getAttribute("contains-table-of-contents");
```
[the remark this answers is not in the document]
vsarya: While I understand to use the BEM style for all the CSS classnames used in this PR, this one is not a CSS class but a request attribute. Do we use the same style for request attributes as well or just prepend it with `cmp:` as pointed by @jckautzmann [here](https://github.com/adobe/aem-core-wcm-components/pull/2030/files#r816953415)?

testing/it/e2e-selenium/src/test/java/com/adobe/cq/wcm/core/components/it/seljup/tests/tableofcontents/v1/TableOfContentsIT.java
```diff
@@ -0,0 +1,211 @@
...
+
+import java.util.HashMap;
+import java.util.concurrent.TimeoutException;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+@Tag("group2")  // TODO
```
vsarya, line 42: @vladbailescu @jckautzmann Which group should I add this test to?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -0,0 +1,291 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Set;
+
+/**
+ * Intercepts all the HTTP requests made to /editor.html or a html page inside /content/.
+ * Creates a response wrapper - {@link CharResponseWrapper} in which all the servlets/filters called after this filter
```
jckautzmann, line 50: Typo? `called` `are called`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -0,0 +1,291 @@
...
+                : TableOfContentsImpl.DEFAULT_STOP_LEVEL;
+        String[] includeClasses = tocPlaceholderElement.hasAttr(TableOfContentsImpl.TOC_DATA_ATTR_INCLUDE_CLASSES)
+            ? tocPlaceholderElement.attr(TableOfContentsImpl.TOC_DATA_ATTR_INCLUDE_CLASSES).split(",")
+            : null;
+        String[] ignoreClasses = tocPlaceholderElement.hasAttr(TableOfContentsImpl.TOC_DATA_ATTR_IGNORE_CLASSES)
+            ? tocPlaceholderElement.attr(TableOfContentsImpl.TOC_DATA_ATTR_IGNORE_CLASSES).split(",")
+            : null;
+
+        if(startLevel.getValue() > stopLevel.getValue()) {
+            LOGGER.warn("Invalid start and stop levels, startLevel={%d}, stopLevel={%d}",
```
jckautzmann, line 158: I would set it as a log error (`LOGGER.error`) as it is a user error.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
vsarya: I think the language is correct. Proper punctuation is required.

`...in which all the servlets/filters called after this filter, store the response content.`

A `,` and `store` instead of `stores` would fix it.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableOfContentsImpl.java
```diff
@@ -0,0 +1,144 @@
...
+@Exporter(name = ExporterConstants.SLING_MODEL_EXPORTER_NAME,
+    extensions = ExporterConstants.SLING_MODEL_EXTENSION)
+public class TableOfContentsImpl extends AbstractComponentImpl implements TableOfContents {
+
+    /**
+     * The resource type of the component associated with this Sling model.
+     */
+    public static final String RESOURCE_TYPE = "core/wcm/components/tableofcontents/v1/tableofcontents";
+
+    public static final String TOC_REQUEST_ATTR_FLAG = "cmp-toc__present";
```
jckautzmann, line 53: I wouldn't use the BEM notation for the request attribute, I'd go with `cmp-toc-present`

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableOfContentsImpl.java
```diff
@@ -0,0 +1,144 @@
...
+
+    public static final String TOC_CONTENT_CLASS = "cmp-toc__content";
+    public static final String TOC_PLACEHOLDER_CLASS = "cmp-toc__placeholder";
+    public static final String TOC_TEMPLATE_PLACEHOLDER_CLASS = "cmp-toc__template-placeholder";
+
+    public static final String TOC_DATA_ATTR_LIST_TYPE = "data-cmp-toc-list-type";
+    public static final String TOC_DATA_ATTR_START_LEVEL = "data-cmp-toc-start-level";
+    public static final String TOC_DATA_ATTR_STOP_LEVEL = "data-cmp-toc-stop-level";
+    public static final String TOC_DATA_ATTR_INCLUDE_CLASSES = "data-cmp-toc-include-classes";
+    public static final String TOC_DATA_ATTR_IGNORE_CLASSES = "data-cmp-toc-ignore-classes";
```
jckautzmann, line 63: I'd put all those constants in the Java interface `TableOfContents.java` as the constants are part of the public CSS classes and HTML data attributes.

jckautzmann:
The start and stop levels are now strings (h1, ..., h6) but it can happen that a h3 is mapped to a start level of 1 (see the example in the TOC page of the component library).
I would keep start and stop level as integers as it was before to avoid confusion.

jckautzmann:
The dialogs (policy and edit) should display H1-H6 levels instead of 1-6 as the filtering is based on H1-H6 tags.

jckautzmann:
The wording for the list type in the dialogs (numbered/bulleted) is different than in the JSON (ORDERED/UNORDERED).
IMHO It should be in sync. I would change the JSON to: NUMBERED/BULLETED.

jckautzmann:
The TOC component should add some data to the data layer.

jckautzmann:
The hint in the edit dialog is as follows: "In order to update its preview, this component requires a page reload after edits are done to itself or to other components that affect it".
As the component automatically refreshes the page after it's updated, the hint should be updated, e.g.: "In order to update its preview, this component requires a page reload after edits are done to other components that affect it".

jckautzmann:
The sample TOC components in the example page in the component library should exclude the side navigation from the displayed TOC.

jckautzmann:
The Github and Adobe Help teasers are broken on the TOC page of the component library.
Screenshot 2022-03-18 at 12 35 58

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableOfContentsImpl.java
vsarya: Okay

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableOfContentsImpl.java
vsarya: I think all the classes except `cmp-toc__content` are part of the implementation details and should be kept in `Impl` class. `cmp-toc__content` should be moved to the interface class as this class is visible in the final TOC render.

vsarya:
> The dialogs (policy and edit) should display H1-H6 levels instead of 1-6 as the filtering is based on H1-H6 tags. The JSON output is correct (H1-H6)

@jckautzmann This is as per the specification doc provided by @gabrielwalt - [here](https://github.com/adobe/aem-core-wcm-components/wiki/Table-of-Content-Component#edit-dialog)
The JSON shows it `H1-H6` because it comes from the `enum HeadingLevel` which cannot have a field just `1` i.e. why I had to name it `H1-H6` there.

vsarya:
> The wording for the list type in the dialogs (numbered/bulleted) is different than in the JSON (ORDERED/UNORDERED).
> IMHO It should be in sync. I would change the JSON to: NUMBERED/BULLETED.

@jckautzmann This was also named as per the specification doc by @gabrielwalt - [here](https://github.com/adobe/aem-core-wcm-components/wiki/Table-of-Content-Component#edit-dialog) which was later changed to `NUMBERED/BULLETED`. 
I'll sync the JSON accordingly.

vsarya:
> The sample TOC components in the example page in the component library should exclude the side navigation from the displayed TOC.

Okay, I've made the relevant changes.

vsarya:
> In order to update its preview, this component requires a page reload after edits are done to other components that affect it

@jckautzmann I also noticed this and that's because the links for the TOC documentations do not exist right now.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableOfContentsImpl.java
jckautzmann: All those classes and data attributes are used in the initial HTML, which makes them part of the public API.

jckautzmann:
> This is as per the specification doc provided by @gabrielwalt - [here](https://github.com/adobe/aem-core-wcm-components/wiki/Table-of-Content-Component#edit-dialog)
The JSON shows it H1-H6 because it comes from the enum HeadingLevel which cannot have a field just 1 i.e. why I had to name it H1-H6 there.

The JSON is correct. I would also use H1-H6 in the dialogs to make it clear to authors that the filtering is based on the H1-H6 tags.

jckautzmann:
While testing the TOC component, I was sporadically hitting following issue:
- the TOC component library page wouldn't load anymore
- error in the developer console: `net::ERR_CONTENT_LENGTH_MISMATCH 200 (OK)`

jckautzmann:
The TOC component creates new IDs for all the H1-H6 elements and references them in the TOC component. E.g.:
`<li><a href="#list-type">List Type</a></li>`
It should instead reference the ID of the component or of its ancestor if it exists and only create a specific one if a component ID does not exist.

gabrielwalt:
The title levels should be named consistently with the [Title](https://www.aemcomponents.dev/content/core-components-examples/library/core-content/title.html) component where we seem to name them `h1` to `h6`. In the [design dialog of the title component](https://experienceleague.adobe.com/docs/experience-manager-core-components/using/components/title.html#sizes-tab) we however seem to display them as capital `H1` to `H6`, but the persistence in JCR still always is in small letters.

jckautzmann:
I discussed with @gabrielwalt and he agrees that the TOC page component should be placed below the `Core Structure` section of the side navigation.

jckautzmann:
@gabrielwalt also agrees that the pages in the side navigation should be ordered alphabetically. As you are updating the component library, could you please also address this request?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/TableOfContentsImpl.java
vsarya: Okay. moving them to the interface.

jckautzmann:
Regarding the IDs, there might be some misunderstandings. The requirements in the [wiki](https://github.com/adobe/aem-core-wcm-components/wiki/Table-of-Content-Component) state:
>Links to the corresponding component on the page by using the id attribute that is set on the closest parent from the title element.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -0,0 +1,291 @@
...
+                : TableOfContentsImpl.DEFAULT_STOP_LEVEL;
+        String[] includeClasses = tocPlaceholderElement.hasAttr(TableOfContentsImpl.TOC_DATA_ATTR_INCLUDE_CLASSES)
+            ? tocPlaceholderElement.attr(TableOfContentsImpl.TOC_DATA_ATTR_INCLUDE_CLASSES).split(",")
+            : null;
+        String[] ignoreClasses = tocPlaceholderElement.hasAttr(TableOfContentsImpl.TOC_DATA_ATTR_IGNORE_CLASSES)
+            ? tocPlaceholderElement.attr(TableOfContentsImpl.TOC_DATA_ATTR_IGNORE_CLASSES).split(",")
+            : null;
+
+        if(startLevel.getValue() > stopLevel.getValue()) {
+            LOGGER.error("Invalid start and stop levels, startLevel={%d}, stopLevel={%d}",
```
jckautzmann, line 158: This doesn't display correctly in the logs. It displays as:
```
Invalid start and stop levels, startLevel={%d}, stopLevel={%d}
```

jckautzmann:
Adding the class `cmp-examples-structure__aside` to both the included and ignored classes in the policy dialog renders an empty TOC. I would expect that it renders at least the content from the main panel.

vsarya:
> Adding the class `cmp-examples-structure__aside` to both the included and ignored classes in the policy dialog renders an empty TOC. I would expect that it renders at least the content from the main panel.

This is the correct behaviour. Include behaviour first filters the `h1-h6` tags inside `include` class names. It then removes all `h1-h6` tags which are inside `ignore` class names i.e. the intersection of `ignore` and `include` class names.
(Everything outside of `include class names` is ignored)

See the detailed behaviour - [here](https://github.com/adobe/aem-core-wcm-components/wiki/Table-of-Content-Component#include--ignore-behavior)

gabrielwalt:
> The TOC component should add some data to the data layer.

No, I see no value in the ToC component to use the data layer at that point. We haven't even yet enabled all components to work with the data layer, so I wouldn't make that a requirement for all new components.

jckautzmann:
>(Everything outside of include class names is ignored)

Ah ok, I thought that `includeClass` was a way to add more elements to the selection. It's working as designed then.

jckautzmann:
>No, I see no value in the ToC component to use the data layer at that point: I don't think that tracking the use of a ToC would generally be useful. Also, we haven't even yet enabled all components to work with the data layer, so I wouldn't make that a requirement for all new components.

@gabrielwalt Hm... ok. I had understood from a recent team discussion that you were in favour of adding TOC data to the data layer. Never mind, let's not make the TOC hydrate the data layer for now.

gabrielwalt:
> The TOC component creates new IDs for all the H1-H6 elements and references them in the TOC component. E.g.: `<li><a href="#list-type">List Type</a></li>` It should instead reference the ID of the component or of its ancestor if it exists and only create a specific one if a component ID does not exist.

I originally had specified the desired behavior as you also put it in your comment, and Vishal had considered it, but in a discussion where Vlad was also involved we then all decided against it. The reason was that it's more complex and less reliable to work with the nearest ancestor that has an ID.

First, that ancestor might have multiple titles inside of it. For example, an implementation could have components that render no IDs, excepted for their outer container. In such a case, for the ToC to distinguish the titles, it would again have to generate specific ID on the title elements, making the implementation more complex and less predictable: when does the ToC generate an ID for a title and when not?

Second, that ancestor might be containing other elements that take a lot of vertical space. For example, by default the Teaser v1 component first displays the teaser image in full width, and it's not uncommon for the title to be so far below that it isn't even within the visible area anymore. In such cases, the ToC would vertically scroll to that image that is not the title, making it less reliable and less user-friendly for the author.

For these reasons, we decided to always generate an ID on the title, excepted when it already exists. This is straight-forward, simple and predictable.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -0,0 +1,295 @@
...
+     * Creates list item element('li') from the heading element.
+     * Adds an internal link on the 'li' element to the provided heading element.
+     * Adds 'id' attribute on heading element if not already present, using its text content.
+     * @param headingElement - DOM heading element
+     * @return Independent 'li' element, not attached to the DOM
+     */
+    private Element getListItemElement(Element headingElement) {
+        String id = headingElement.attr("id");
+        if("".contentEquals(id)) {
+            id = headingElement.text()
```
jckautzmann, line 264: Two heading elements with the same title will currently have the same generated ID. IDs should be unique.
To make the ID unique you could e.g. suffix the current ID with a counter that is incremented for each heading element.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -0,0 +1,295 @@
...
+                        tocTemplatePlaceholderElement.remove();
+                    }
+                }
+            }
+
+            CharArrayWriter charWriter = new CharArrayWriter();
+            charWriter.write(document.outerHtml());
+            String alteredContent = charWriter.toString();
+
+            response.setContentLength(alteredContent.length());
```
jckautzmann, line 118: IMHO setting the content length of the request header should not be done in a filter: if different filters set the content length based on their internal logic, it might lead to inconsistencies. It may be the root cause of the sporadic issue [0] I observed while testing the TOC component.

[0] https://github.com/adobe/aem-core-wcm-components/pull/2030#issuecomment-1073706066

gabrielwalt:
A few comments from today's review meeting about the table of contents component. Most items have already been raised in this PR, but a few are new.
* Don’t allow authors to configure invalid options in the dialog and validate the options in the dialog to not allow a start level that is bigger than the stop level.
* Don’t throw errors on invalid configurations, but rather warnings.
* Order both the start-level and the end-level drop-downs the same way (ascending).
* Rename the title levels to H1-H6 in the UI and to h1-h6 in the JCR persistence and in the JSON returned by the Sling Models.
* Make the ID of titles that have the same text unique by appending a number, but don’t append anything to titles that don’t need to be made unique.

vsarya:
> * Don’t allow authors to configure invalid options in the dialog by validating the options and preventing a start level that is bigger than the stop level.

@gabrielwalt All improvements except the one mentioned are done. I'll be back working on this item after my next week's PTO.

cc - @jckautzmann , @vladbailescu

jckautzmann:
@vsarya I noticed a shift in the customID index. If you have 3 titles `mytitle` the generated IDs are currently : `mytitle`, `mytitle-2` and `mytitle-3`. IMHO they should be: `mytitle`, `mytitle-1` and `mytitle-2`.

vsarya:
> @vsarya I noticed a shift in the customID index. If you have 3 titles with the same text, e.g. `mytitle` the generated IDs are currently : `mytitle`, `mytitle-2` and `mytitle-3`. IMHO they should be: `mytitle`, `mytitle-1` and `mytitle-2`.

@jckautzmann Just pushed the required changes.
commit - https://github.com/adobe/aem-core-wcm-components/pull/2030/commits/acfbf8940fc9601bcaa53b65fd46165e5f5b7dd2

jckautzmann (approved):

jckautzmann:
@vsarya Please fix the build error due to the baseline error:
https://github.com/adobe/aem-core-wcm-components/runs/5971672041?check_suite_focus=true

jckautzmann:
@vsarya What is the status with the broken links to the 2 docs links (Github + Adobe)?
Screenshot 2022-04-11 at 15 49 43

jckautzmann:
I'd change the dialog validation message to: "Stop level is smaller than start level" as the tooltip applies to the stop level field.
Screenshot 2022-04-11 at 15 54 06

vsarya:
> @vsarya What is the status with the broken docs links (Github + Adobe)? Screenshot 2022-04-11 at 15 49 43

@jckautzmann I had already created a CQDOC issue for the same as discussed in our weekly meetings - https://jira.corp.adobe.com/browse/CQDOC-18998

vladbailescu (approved):

jckautzmann:
Thx @vsarya for implementing this TOC component!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -0,0 +1,305 @@
...
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+        throws IOException, ServletException {
+
+        CharResponseWrapper responseWrapper = new CharResponseWrapper((HttpServletResponseWrapper) response);
+        chain.doFilter(request, responseWrapper);
+        String originalContent = responseWrapper.toString();
+        Boolean containsTableOfContents = (Boolean) request.getAttribute(TableOfContentsImpl.TOC_REQUEST_ATTR_FLAG);
+        if (responseWrapper.getContentType().contains("text/html") &&
```
grubyak, line 89: @jckautzmann we observed NPE in this line during our on-call shift

jckautzmann: @grubyak thx for your feedback. Could you please create a ticket describing the issue and how it can be reproduced? Thx!

Issue #2056, opened 2022-03-06
If more than 1 form action dialog has the same name, on submit of dialog it saves as string array


punithshetty123:
## Bug Report

**Current Behavior**
Create 2-3 form actions with dialog and its property, make sure you have same property names in both the form dialog

Form Action1 -> Dialog (property1(name=."/p1", property2(name="./p2", property3(name="./p3")))
Form Action2 -> Dialog (property1(name=."/p1", property2(name="./p2", property3(name="./p4")))

Select one of the form field, fill the details and submit it
Ex: Select Form Action1 and submit, Below is what is saved on the AEM node,
p1 => String[] => property1 value, 
p2 => String[] => property2 value, 
p3 => String => property2 value

Since same name is stored in 2 different form action, on submit of form dialog, It sends both the properties from 2 form actions. Hence stores as String []

We need to make sure that only the properties from the form action select dropdown is submitted.

**Expected behavior/code**
p1 => String => property1 value
p2 => String => property2 value
p3 => String => property2 value

**Environment**
- AEM 6.5 SP10
- Core Components version: 2.18.0
- JRE version 
java version "1.8.0_202"
Java(TM) SE Runtime Environment (build 1.8.0_202-b08)

Issue #2052, opened 2022-03-06
AEM Core Components Form Action Dialog validation


punithshetty123:
## Bug Report

**Current Behavior**
Validation
Create 2-3 form actions with dialog and its property, make sure you have required property on one form dialog

Form Action1 -> Dialog (textfield(required), pathfield(required),select(required),checkbox(required))
Form Action2 -> Dialog (textfield, pathfield,select,checkbox)

Now in form Container choose Form Action 2 and try to submit the form. It fails in the current environment. 

**Expected behavior/code**
Since Form Action2 dialog did not have any required field, it was expected to submit successfully.

**Environment**
- AEM 6.5 SP10
- Core Components version: 2.18.0
- JRE version
java version "1.8.0_202"
Java(TM) SE Runtime Environment (build 1.8.0_202-b08)

PR #2057, opened 2022-03-06
[AEM Core Components Form Action Dialog] fix required validation for property on Form Action dialog and also fix the issue of saving the property as multiarray when the same name exists in diff form action dialog

punithshetty123:
issue#2052: Instead of targeting just the input[aria-required=true], coral-multifield[aria-required=true] selectors, we need to genralize and target all [aria-required=true],[required] so that other sling:resourceTypes are also covered

issue#2056: Enable only the input forms which is selected in the dropdown, Make the rest of them disabled so that it doesn't save as string[] if the same name exists in another form action

| Fixed Issues?            | https://github.com/adobe/aem-core-wcm-components/issues/2052, https://github.com/adobe/aem-core-wcm-components/issues/2056
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Tests Added + Pass?      |
| Documentation Provided   |
| Any Dependency Changes?  |


punithshetty123:
Can someone please approve the workflows? I am First-time contributors here

PR #2101, opened 2022-03-18
FIX: Issue 2099 - Prevents Accordion to throw an error when selecting an element by URL hash

### LINKED ISSUE #2099: [Accordion v1] Breaking the javascript execution when there is a hash in the URL starting with a number, like #3
## Bug Report

**Current Behavior**
When you have a hash in the URL starting with a number like that `your/page/path/index.html#3` the Accordion clientlib just throw an error that breaks the Javascript execution.

**Error:** `Uncaught DOMException: Failed to execute 'querySelector' on 'Document': '#3' is not a valid selector.`

This error is triggered because the Accordion script tries to select an element based on the hash from URL whithout check if the selector is a valid one and in this case, any css selector starting with a number is not valid for the `document.querySelector` method.

**Impact**
Every site that uses the Accordion core component and has its clientlib imported will have this problem.

**Expected behavior/code**
Do not break the code 

**Environment**
- AEM version and patch level - Adobe Experience Manager (6.5.8.0)
- Core Components version 2.18.0
- Java "11.0.1" 2018-10-16 OpenJDK Runtime Environment 18.9 (build 11.0.1+13)

**Possible Solution**

**Line of the code that is causing the error**
https://github.com/adobe/aem-core-wcm-components/blob/main/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js#L510

fagnervalente:
| Fixed Issues?            | Fixes #2099  
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

I've add a check if the hash starts with a number, if so, it will not go ahead with the `onHashChange` function and will avoid the error on try to select a element with a invalid CSS selector.

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
@@ -506,7 +506,11 @@
      */
     function onHashChange() {
         if (location.hash && location.hash !== "#") {
+            var unvalidSelectorRegex = /^#[0-9]/;
             var anchorLocation = decodeURIComponent(location.hash);
+            if (unvalidSelectorRegex.test(anchorLocation)) {
+                return;
+            }
             var anchorElement = document.querySelector(anchorLocation);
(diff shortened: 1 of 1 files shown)


bpauli:
@fagnervalente This issue was meanwhile fixed with https://github.com/adobe/aem-core-wcm-components/pull/2138. But thank you very much for the hands up and the PR!

PR #2127, opened 2022-04-07
SITES-5721 [Title] - Policy Change should not reset default type

### LINKED ISSUE #1700: [Title] - Policy Change should not reset default type
**Current Behavior**
When Template Editors make changes to a Title policy, the default type is always reset to the first heading that is checked, regardless of the previous value.

**Expected behavior/code**
If the current value of the default type is still a valid option from the list of allowed values, then no action should be taken.

**Environment**
- AEM SDK 
- Core Components version 2.17.2

**Possible Solution**
[This ](https://github.com/adobe/aem-core-wcm-components/blob/2cf246bdd2e37d66d111dc4da17b822bba1e2781/content/src/content/jcr_root/apps/core/wcm/components/title/v2/title/clientlibs/editor/js/title.js#L74-L81 )should check the list of checked values to determine if the value is still valid, rather than a simple reset.

comanV:
Fixes #1700


vladbailescu:
We should probably have an IT as well.

msagolj (approved):

PR #2196, opened 2022-05-30
make link handler API public

### LINKED ISSUE #2168: [LinkHandler] make the LinkHandler API public
Define the public API for the LinkHandler.

See discussion and other links relating to this topic at <0>.

<0> https://github.com/adobe/aem-core-wcm-components/discussions/1861

jckautzmann:
- draft API

| Fixed Issues?            | Fixes #2168
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
```diff
@@ -0,0 +1,55 @@
...
+public interface LinkHandler {
+
+    /**
+     * Returns a link builder based on the given resource.
+     *
+     * @param resource Resource to read the link properties from.
+     * @return {@link LinkBuilder}
+     */
+    @NotNull
+    LinkBuilder get(@NotNull Resource resource);
```
kwin, line 35: This is always a new(!) LinkBuilder, therefore `create()` would be a better fit than `get()`, but in this case I would just add a constructor to LinkBuilder (more obvious how to create) and get rid of this interface all together!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Returns a link builder pointing to the given target page.
+     *
+     * @param page Target page of the link.
+     * @return {@link LinkBuilder}
+     */
+    @NotNull
+    LinkBuilder get(@NotNull Page page);
```
kwin, line 44: Same as https://github.com/adobe/aem-core-wcm-components/pull/2196/files#r884858227

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
jckautzmann: I also prefer `get` as in essence it is returning a link based on properties defined by the Resource.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+    LinkBuilder setLinkTarget(@Nullable String target);
+
+    /**
+     * Sets the link attributes.
+     *
+     * @param attributes The link attributes
+     * @return {@link LinkBuilder}
+     */
+    @NotNull
+    LinkBuilder setLinkAttributes(@NotNull Map<String, String> attributes);
```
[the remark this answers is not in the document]
jckautzmann: This method provides a flexible way to add HTML attributes to the link, e.g. `aria-label` or `title`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.sling.api.resource.Resource;
+import org.jetbrains.annotations.NotNull;
+
+import com.day.cq.wcm.api.Page;
+
+/**
+ * Computes a link based on input configuration.
+ */
+public interface LinkHandler {
```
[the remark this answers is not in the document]
jckautzmann: It is meant to be a Sling model: I updated the javadocs at https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/50909be5b37fcbcd8d62ceeae55478d70fc3fe3f

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Returns the resolved link.
+     *
+     * @return {@link Link}
+     */
+    Link build();
```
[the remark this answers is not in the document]
jckautzmann: Addressed in https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/50909be5b37fcbcd8d62ceeae55478d70fc3fe3f

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
kwin: > it's not unlikely that the link builder impl will need to be injected some references or other stuff, so reducing this to a constructur of linkbuilder will make that difficult.

Ok, then what about renaming this class to `LinkBuilderFactory`?

stefanseifert: the goal is not to create `LinkBuilder` instances (this is only an intermediate step), but to create `Link` instances. in that direction `LinkFactory` could be ok. (on the other hand having two classes `LinkFactory` and `LinkBuilder` in the API is also a bit confusing for the user)

kwin: The API requires two steps:
First creating the link builder and then creating the link out of that (that is the case right now).
IMHO it would be clearer to express the different roles in the class name, therefore `LinkBuilderFactory` (to create the `LinkBuilder`) and then `LinkBuilder` (to create the actual link) is reasonable for me.
OTOH `LinkBuilder` and `LinkFactory` would be highly confusing as the latter would not create links but rather builder instances.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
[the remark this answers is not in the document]
stefanseifert: ok - still wondering if it might make sense to create dedicated methods like setLinkTarget for use cases like accessibility title which may affect lot of components.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
stefanseifert: i agree that `LinkFactory` is not a good name for this reasons.

in general i'm fine with both `LinkHandler` and `LinkBuilderFactory`. i've a slight preference to the former as it might be we need additional Link-related methods in the link handler in the future which do not return a LinkBuilder. for example, in https://wcm.io/handler/link/apidocs/io/wcm/handler/link/LinkHandler.html we have a method "invalid" which returns an invalid links. which looks odd, but as a link object is never null it's otherwise difficult if a component has to return an invalid link on purpose e.g. if no page, resource or string as input is available. maybe that's an edge case which could be solved otherwise.

kwin: For me `LinkHandler` is kind of fuzzy and not really explaining well the purpose of this class, therefore I have a strong preference towards `LinkBuilderFactory` :-)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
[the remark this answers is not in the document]
jckautzmann: As there might be requirements to add specific attributes to a link, I would keep a flexible way to add them. We could replace it with a method to only add one attribute at a time, e.g.:
```
LinkBuilder setLinkAttribute(String key, String value);
```
In a first step I wouldn't add specific methods other than `setLinkTarget`. We could add specific methods later if needed.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
jckautzmann: As @kwin noted, this proposal imposes a 2-steps definition for a link. We could simplify it by getting rid of the `LinkBuilder` and only introduce a `LinkHandler` interface as follows (if this idea gets traction I would create a 4th draft with detailed javadocs):
```
    Link getLink(@NotNull Resource resource);
    Link getLink(@NotNull Resource resource, @Nullable String linkURLPropertyName);
    Link getLink(@NotNull Page targetPage);
    Link getLink(@NotNull String URL);
    void setLinkTarget(Link link, @Nullable String linkTarget);
    void setLinkAttribute(Link link, String key, String value);
```
WDYT?

jckautzmann: any other ideas to avoid the 2 steps to get a link?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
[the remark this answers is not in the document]
stefanseifert: ok - yes, i like the setLinkAttribute approach better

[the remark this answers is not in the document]
jckautzmann: I updated the method in https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/9aef6a37a154aff71b0a68608de42a54dc83b1d7

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
```diff
@@ -0,0 +1,55 @@
...
+    LinkBuilder get(@NotNull Page page);
+
+    /**
+     * Returns a link builder pointing to the URL.
+     *
+     * @param URL URL string of the link.
+     * @return {@link LinkBuilder}
+     */
+    @NotNull
+    LinkBuilder get(@NotNull String URL);
```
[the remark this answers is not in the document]
stefanseifert: cosmetic: following java conventions the parameter should be lowercase `url`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Sets the HTML link attribute.
+     *
+     * @param name The attribute name
+     * @param value The attribute value
+     * @return {@link LinkBuilder}
+     */
+    @NotNull
+    LinkBuilder setLinkAttribute(@NotNull String name, String value);
```
stefanseifert, line 52: is it allowed to have a null-value? please make it explicit by adding @NotNull or @Nullable

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+public interface LinkBuilder {
+
+    /**
+     * Sets the property name to read the link URL from.
+     *
+     * @param name The property name
+     * @return {@link LinkBuilder}
+     */
+    @NotNull
+    LinkBuilder setLinkURLPropertyName(@Nullable String name);
```
stefanseifert, line 33: what's the usecase for passing null here? probably @NotNull is more appropriate?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Sets the HTML link target.
+     *
+     * @param target The link target
+     * @return {@link LinkBuilder}
+     */
+    @NotNull
+    LinkBuilder setLinkTarget(@Nullable String target);
```
stefanseifert, line 42: what's the usecase for passing null here? probably @NotNull is more appropriate?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+package com.adobe.cq.wcm.core.components.commons.link;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Builds a link and sets link properties.
+ */
+public interface LinkBuilder {
```
stefanseifert, line 24: please add for both interfaces a `@ConsumerType` or `@ProviderType` annotation. which to pick? is it allowed that 3rpartys provider their own implementations (like wcm.io) - than it should be `@ConsumerType`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
```diff
@@ -0,0 +1,56 @@
...
+
+}
```
stefanseifert, line 56: one use case is currently not covered by the interface: how to build a link **pointing to an asset** (e.g. PDF download)? there is no signature for it. it could be confused with `get(<Resource>)` by passing the resource pointing to the asset, but that will not work as the resource is only used to read the link properties from.
probably best to add a new method to avoid having to create a synthetic resource with a link property pointing to an asset.

alternatively `get(<String>)` could be made more magic by infer from the string if it is an external url, a page path or an asset path. parameter name and javadoc need to be updated in that case.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
jckautzmann: Addressed with https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/91c774f82f21b3b8a5c3116ee3776f21783d4f59

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
```diff
@@ -0,0 +1,56 @@
...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+package com.adobe.cq.wcm.core.components.commons.link;
+
+import org.apache.sling.api.resource.Resource;
+import org.jetbrains.annotations.NotNull;
+
+import com.day.cq.wcm.api.Page;
+
+/**
+ * Computes a link based on input configuration.
```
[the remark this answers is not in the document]
jckautzmann: Addressed at https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/91c774f82f21b3b8a5c3116ee3776f21783d4f59

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
jckautzmann: Addressed at https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/91c774f82f21b3b8a5c3116ee3776f21783d4f59

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
jckautzmann: Addressed at https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/91c774f82f21b3b8a5c3116ee3776f21783d4f59

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
jckautzmann: Addressed at https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/91c774f82f21b3b8a5c3116ee3776f21783d4f59

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
jckautzmann: good catch! I added a specific method to define a link based on an asset at https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/91c774f82f21b3b8a5c3116ee3776f21783d4f59

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
```diff
@@ -0,0 +1,69 @@
...
+     * Returns a link builder where the link points to an asset.
+     *
+     * @param asset Target asset of the link.
+     * @return {@link LinkBuilder}
+     */
+    @NotNull
+    LinkBuilder get(@NotNull Asset asset);
+
+    /**
+     * Returns a link builder where the link pointa to an URL.
```
stefanseifert, line 61: cosmetic: typo pointa -> points

stefanseifert (approved):
lgtm, thanks!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
jckautzmann: Addressed in https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/73adabead1a295cbe0ae9f76d2404781bb599330

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Sets the property name to read the link URL from.
+     *
+     * @param name The property name
+     * @return {@link LinkBuilder}
+     */
+    @NotNull
+    LinkBuilder setLinkUrlPropertyName(@NotNull String name);
```
vladbailescu, line 35: I would generally prefer the `withURL(...).withTarget(...).withAttribute(...)` syntax for a builder.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
vladbailescu: +1 for `LinkManager`

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
stefanseifert: that's fine from my side

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkHandler.java
stefanseifert: LinkManager is fine from my side as well

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkBuilder.java
jckautzmann: I renamed the methods in https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/fc7e4497db96ddcf3ec2e84e4e7b93417a52d2d2

jckautzmann:
I renamed `LinkHandler` to `LinkManager` in https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/fc7e4497db96ddcf3ec2e84e4e7b93417a52d2d2

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/commons/link/LinkManager.java
```diff
@@ -0,0 +1,78 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.osgi.annotation.versioning.ConsumerType;
+
+import com.day.cq.dam.api.Asset;
+import com.day.cq.wcm.api.Page;
+
+/**
+ * This interface offers a flexible way, based on the builder pattern, to compute links.
+ *
+ * This is a Sling model that can be injected into other models using the <code>@Self</code> annotation.
```
Buuhuu, line 28: I would probably help to mention what the adaptable should ne, Resource or SlingHttpServletRequest?

jckautzmann: Addressed in https://github.com/adobe/aem-core-wcm-components/pull/2196/commits/c591f65e23c98690da492530a6d611269962ac49

PR #2224, opened 2022-06-23
Implementation for the newly defined API: LinkHandler and LinkBuilder

### LINKED ISSUE #2168: [LinkHandler] make the LinkHandler API public

jckautzmann:
This PR adds implementations for the public API `LinkHandler` and `LinkBuilder` as defined in the pull request #2196:
- makes `LinkHandlerImpl` a Sling model that can be adapted from the request (similar behaviour as the former internal  `LinkHandler`)
- adapts the Core Component models to use the new `LinkHandler` API
- adds tests  for redirect chains and shadowing: cherry-picked from pull request #1864 created  by @ky940819
- `LinkHandlerImpl`: code has been cherry-picked from the former internal `LinkHandler` code and some bits from the pull request #1864  by @ky940819 
- removes the former internal `LinkHandler` and `LinkHandlerTest`
- adds `LinkTest` to test the default methods of the `Link` API

| Fixed Issues?            | Fixes #2168 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


jckautzmann:
@stefanseifert @kwin @ky940819 It would be great to get your feedback as you've been heavily involved in all the discussions about making the Link Handler part of the public API. Thx!

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
```diff
@@ -0,0 +1,282 @@
...
+    Resource resource;
+    String linkUrl;
+    Map<String, String> linkAttributes = new HashMap<>();
+    Object reference;
+
+    public LinkBuilderImpl(Resource res, SlingHttpServletRequest req, List<PathProcessor> pathProcs, boolean shadowingDisabled) {
+        if (res != null) {
+            resource = res;
+            ValueMap props = resource.getValueMap();
+            linkUrl = props.get(Link.PN_LINK_URL, String.class);
```
vladbailescu, line 65: As a a general rule, I would normally expect this kind of reads be done at `build` time to account for possible changes in the input data (resource, request). I know this is not the case, as we don't expect these to be manipulated after the builder is created, but to me they're a bit of a code smell.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
```diff
@@ -0,0 +1,282 @@
...
+        request = req;
+        pathProcessors = pathProcs;
+        this.shadowingDisabled = shadowingDisabled;
+    }
+
+    @Override
+    public @NotNull LinkBuilder setLinkUrlPropertyName(@NotNull String name) {
+        if (resource != null) {
+            ValueMap props = resource.getValueMap();
+            linkUrl = props.get(name, String.class);
```
vladbailescu, line 113: Same as https://github.com/adobe/aem-core-wcm-components/pull/2224/files#r906113819

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
```diff
@@ -0,0 +1,282 @@
...
+        }
+        return this;
+    }
+
+    @Override
+    public @NotNull LinkBuilder setLinkTarget(@NotNull String target) {
+        String resolvedLinkTarget = validateAndResolveLinkTarget(target);
+        return setLinkAttribute(PN_LINK_TARGET, resolvedLinkTarget);
```
vladbailescu, line 121: Same as https://github.com/adobe/aem-core-wcm-components/pull/2224/files#r906113819

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    @Override
+    public @NotNull LinkBuilder setLinkAttribute(@NotNull String name, @Nullable String value) {
+        String validatedLinkAttributeValue = validateLinkAttributeValue(value);
+        linkAttributes.put(name, validatedLinkAttributeValue);
```
vladbailescu, line 127: Same as https://github.com/adobe/aem-core-wcm-components/pull/2224/files#r906113819

vladbailescu (approved):
Except for a few small comments on the actual implementation of the builder, I like the approach and I find the usage pattern nice.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
```diff
@@ -359,17 +362,17 @@ public static Resource getWrappedImageResourceWithInheritance(Resource resource,
 
             if (StringUtils.isNotEmpty(linkURL)) {
                 // the inherited resource is the featured image of the linked page
-                Optional<Link> link = linkHandler.getLink(resource);
+                Optional<Link> link = Optional.of(linkHandler.get(resource).build());
```
stefanseifert, line 365: be careful with code like this (only commenting this here, other places affected as well): linkHandler.get never returns null, so wrapping it in Optional is quite useless. instead it leads to false assumption that and invalid link produces any empty optional (which is not the case). instead, it returns a link object that is marked with valid=false. this needs to be checked for further link processing, instead of Optional. or create a helper method for the legacy code that converts a link object o an Optional, setting it to empty if the link is not valid.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
```diff
@@ -0,0 +1,282 @@
...
+        if (res != null) {
+            resource = res;
+            ValueMap props = resource.getValueMap();
+            linkUrl = props.get(Link.PN_LINK_URL, String.class);
+
+            String linkTarget = props.get(PN_LINK_TARGET, String.class);
+            if (StringUtils.isNotEmpty(linkTarget)) {
+                setLinkTarget(linkTarget);
+            }
+            String linkAccessibilityLabel = props.get(PN_LINK_ACCESSIBILITY_LABEL, String.class);
```
stefanseifert, line 71: PN_LINK_ACCESSIBILITY_LABEL and PN_LINK_TITLE_ATTRIBUTE are marked as deprecated - should they still be used in new code? (unfortunately it's not documented why they are deprecated)

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
jckautzmann: Addressed at https://github.com/adobe/aem-core-wcm-components/pull/2224/commits/bddfc37eccfe68fc3a857082daa5dc9c1e56c5d1

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
jckautzmann: Addressed at https://github.com/adobe/aem-core-wcm-components/pull/2224/commits/bddfc37eccfe68fc3a857082daa5dc9c1e56c5d1

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/Utils.java
jckautzmann: I introduced a helper method `getOptionalLink()` at https://github.com/adobe/aem-core-wcm-components/pull/2224/commits/624c108ed14f56dd2b361b645f68368129866033

jckautzmann:
I addressed all the feedback for this pull request. If I don't get any feedback until Friday (01/july), I will merge this pull request together with the pull request about the new `LinkManager` and `LinkBuilder` APIs ( https://github.com/adobe/aem-core-wcm-components/pull/2196).
cc: @kwin @stefanseifert @ky940819

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
stefanseifert: understood, ok. than we need this here for backward compatibility.

PR #2225, opened 2022-06-23
AdaptiveImageServlet isn't respecting the contentPolicyDelegatePath

### LINKED ISSUE #2209: AdaptiveImageServlet isn't respecting the contentPolicyDelegatePath
## Bug Report

**Current Behavior**
In the past, there's support for contentPolicyDelegatePath in AdaptiveImageServlet (AIS) which helps in deleting the resolution of the image content policy to a specific resource. I think it was added for supporting targeting on Image component. Once the image component is targeted, the default rendering is part of the page, but the targeted experiences are stored under /content/campaigns, where we can't specify the content policies. Through contentPolicyDelegatePath, we can tell AIS to use the content policy of the default component inside the page.

I'm working on adding the support for Targeting inside the Experience Fragments and the image component has a defined content policy that sets allowedRenditionWidths. When the image component is targeted, Image component is making this call to request the adaptive rendition for the screen size
http://localhost:4502/content/campaigns/wknd/master/TestActivity/chrome-experience/master-image/jcr:content/par.html?contentPolicyDelegatePath=/content/experience-fragments/wknd/language-masters/en/contributors/kumar-selveraj/master/jcr:content/root/container/image/default&_=1654621623364

But this request is failing with 404 and the below error
```
07.06.2022 10:09:42.768 *ERROR* [[0:0:0:0:0:0:0:1] [1654621782763] GET /content/campaigns/wknd/master/TestActivity/chrome-experience/master-image/_jcr_content/par/image.coreimg.75.1200.jpeg/1654020856124/adobestock-241578158.jpeg HTTP/1.1] com.adobe.cq.wcm.core.components.internal.servlets.AdaptiveImageServlet Invalid image request
java.lang.IllegalArgumentException: Third selector must contain a valid width information (selector > 0)

```
Further investigation in the AIS, showed that content policy for this request isn't being set as we're not passing the `request` object to 
`policyManager.getPolicy(imageResource)` since this [commit ](https://github.com/adobe/aem-core-wcm-components/pull/754)

Is the support for `contentPolicyDelegatePath` dropped deliberate? Is there a better way to achieve the same?

**Expected behavior/code**
As the targeted experiences of the image component can't have the content policy itself, Please provide an alternate way to delegate the content policy of specific resource.

**Environment**
- AEM CS SDK aem-sdk-quickstart-2022.5.7493.20220524T082827Z-220401
- Core Components version - core.wcm.components.content-2.19.2.zip
- JRE version - openjdk version "11.0.7" 2020-04-14 LTS

**Possible Solution**
I've tried to add the `request` object like
```
policyManager.getPolicy(imageResource, request);
```
and also added the request object in all other calls where this is required to be able to pass this to [getContentPolicy](https://github.com/adobe/aem-core-wcm-components/blob/main/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L852)
This worked, as the `policyManager` reads the `contentPolicyDelegatePath` request parameter and able to resolve the content policy from the `contentPolicyDelegatePath` value

ravkiran:
| Fixed Issues?            | Fixes #2209
| Tests Added + Pass?      | TODO

Adding `request` object to policyManager.getPolicy, will be able to read the `contentPolicyDelegatePath` from the request parameter and read the content policy from the  value of`contentPolicyDelegatePath`. Updated other private methods to include the `request` object.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -247,7 +247,7 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
             }
             if (!handleIfModifiedSinceHeader(request, response, lastModifiedEpoch)) {
 
-                Map<String, Integer> transformationMap = getTransformationMap(selectorList, component);
+                Map<String, Integer> transformationMap = getTransformationMap(selectorList, component, request);
                 Integer jpegQualityInPercentage = transformationMap.get(SELECTOR_QUALITY_KEY);
                 double quality = jpegQualityInPercentage / 100.0d;
                 int resizeWidth = transformationMap.get(SELECTOR_WIDTH_KEY);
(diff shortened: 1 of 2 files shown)


vladbailescu:
Changes were picked up in  #2256

Issue #1289, opened 2020-11-30
Don't use i18n for strings visible on Publish instances


kwin:
## Feature Request

All text visible to end users of the web site should be manageable without code deployment.

As i18n dictionary translation is no longer supported at run time with AEM as a Cloud Service (https://experienceleague.adobe.com/docs/experience-manager-cloud-service/sites/sites-cloud-changes.html?lang=en#capabilities-not-available-at-runtime) the [HTL option i18n](https://experienceleague.adobe.com/docs/experience-manager-htl/using/htl/expression-language.html?lang=en#internationalization) should no longer be used with texts/labels visible in the end user's HTML (e.g. https://github.com/adobe/aem-core-wcm-components/blob/dbdd2c84ad728a3249689a8802a43a60f4b39fd6/content/src/content/jcr_root/apps/core/wcm/components/embed/v1/embed/embeddable/youtube/youtube.html#L22)

 A runtime editable alternative would be to leverage Context-Aware configurations for all i18n purposes affecting the end user / publish instance.

henrykuijpers:
Imho, this issue should be renamed to "i18n strings used are not editable on instances with immutable /apps content". I don't think the proposed alternative is a good one: We need something that is carried from Sling (or maybe AEM, though I think it's a Sling-thing). This means that new functionality needs to be developed.

The core components should only reuse that functionality. Since that new functionality will not be available in all AEM versions these core components are being used, I would propose to create some kind of bridge, so that backwards compatibility with older AEM versions can be maintained.

There are actually a few locations in AEM where this issue was probably also encountered:
* One pattern that relates to this issue is already "hacked" in AEM: com.day.cq.dam.core.impl.I18nResourceBundleProvider defines an aggregated resource bundle provider, that is able to collect multiple ResourceBundleProvider implementations and then query all of them to request ResourceBundles from them.
* The JcrResourceBundleProvider is already doing quite a lot of the work that is needed and could be updated to have multiple implementations that can find / detect ResourceBundles that should be used for that specific context. I18NFilter (that is also found in the i18n impl bundle) is also having this CombinedResourceBundleProvider implementation and it also has a RequestLocaleResolver, that can be used to detect the locale that should be used. 
[...]

PR #2242, opened 2022-07-04
Remove usage of Granite.I18n.get for JS that is used on published site

### LINKED ISSUE #2236: Remove usage of Granite.I18n.get for JS that is used on published site
https://github.com/adobe/aem-core-wcm-components/blob/0875847c6cfbe56717a7022cb7e228e04a395ccb/content/src/content/jcr_root/apps/core/wcm/components/commons/site/clientlibs/link/js/link.js#L22

`Granite.I18n.get `requires `granite.utils` clientlib which we should NOT load on public facing sites - we use it for areas that are when the pages is edited - in the editor.

When I change the line to

`var linkAccessibilityHtml = "<span class='" + linkAccessibilityClass + "'>opens in a new tab</span>";`

I can remove the `granite.utils` clientlib and the site loads ok

jckautzmann:
- removes the use of `Granite.I18n.get()` and the `granite.utils clientlibs` on publish instance
- the link accessibility script, when included to the page, adds an accessible and i18n string to all the links of the page, when:
    - accessibility is enabled through a data attribute at page level (always the case for Page v3)
    - an i18n string is available in the page HTML through a data attribute (always the case for Page v3)
- introduces data attributes in the Page v3:
    - `data-cmp-link-accessibility-enabled`: to notify that the link accessibility can be supported
    - `data-cmp-link-accessibility-text`: to hold the accessible i18n string appended to all the links of the page

| Fixed Issues?            | Fixes #2236 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ For existing projects, take example from the [AEM Project Archetype](https://git
 
 Core Components | AEM as a Cloud Service | AEM 6.5       | Java SE | Maven
 ----------------|------------------------|---------------|---------|---------
-[2.20.4](https://github.com/adobe/aem-core-wcm-components/releases/tag/core.wcm.components.reactor-2.20.4) | Continual | 6.5.13.0+ (*) | 8, 11 | 3.3.9+
+[2.20.6](https://github.com/adobe/aem-core-wcm-components/releases/tag/core.wcm.components.reactor-2.20.6) | Continual | 6.5.13.0+ (*) | 8, 11 | 3.3.9+
 
 For the requirements from previous Core Component releases, see [Historical System Requirements](VERSIONS.md).
 
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -4,6 +4,7 @@ See below for a full list of system requirements for historical versions of the 
 
 Core Components | Extension | AEM 6.5 | AEM 6.4 | AEM 6.3 | Java | Maven
 ----------------|-----------|---------|---------|---------|------|------
+[2.20.6](https://github.com/adobe/aem-core-wcm-components/releases/tag/core.wcm.components.reactor-2.20.6) | - | 6.5.13.0+ | - | - | 8, 11 | 3.3.9+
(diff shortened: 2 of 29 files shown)


content/src/content/jcr_root/apps/core/wcm/components/page/v3/page/page.html
```diff
@@ -22,6 +22,8 @@
     <head data-sly-call="${head.head @ page = page, pwa = pwa}"></head>
     <body class="${page.cssClassNames}"
           id="${page.id}"
+          data-cmp-link-accessibility-enabled
+          data-cmp-link-accessibility-text="${'opens in a new tab' @ i18n}"
```
Buuhuu, line 26: Would it be possible to expose a dictionary here with labels defined by the Model? This would allow us in CIF to reuse the approach without the need to overwrite the page.html? 

We could even introduce a new model interface "Dictionary" which is serialised to a json key-value map. Custom implementation could extend this using the delegation pattern and add any additional labels they need for the same/similar cases.

jckautzmann: the long-term solution is to find a replacement to the `Granite.i18n` clientlib so that we can internationalise strings on publish instances. Your suggestion would be a step towards this goal. IMHO this replacement should be part of the product. For now I would keep it as it is as a short term solution.

Buuhuu: Shall we have a ticket to track this long-term solution than?

Buuhuu: #1289 is about server side i18n. This PR is targeting to remove client side i18n based on the usage of Granite.I18n. For that it doesn't matter where the translations come from, but how they are exposed to a script on a site.

kwin: #1289 is about replacing Sling Dictionaries with something maintainable at run time (no matter whether leveraging on server side with HTL i18n or on client side with some JS).

bpauli (approved):

PR #2246, opened 2022-07-05
Quick Site Creation - XF styling

### LINKED ISSUE #2245: Quick Site Creation - XF styling
We recently observed a behaviour that we want to confirm regarding Quick Site Template and Experience Fragment Styling
in AEM version (6698)
We had one of the Site template where we were able to get styling up in XF page as well i.e in /content/experiencefragments path .
in AEM Verison (7575)
We get the XF styling fine when rendered on page but when we load that XF itself in path /content/experiencefragments - styling CSS which is there in template does not load.

Do you foresee anything that has been done in recent version that could result in such behaviour.We are able to replicate the same behavior on multiple instances (edited)

bpauli:
- add filter to exclude internal adobe models from LatestVersionImplementationPicker

fixes #2245

| Fixed Issues?            | Fixes #2245 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


vladbailescu (changes requested):
We need at least unit tests that cover both scenarios.

bpauli:
@vladbailescu added unit test in 2aeff22c29b9979e7a632b1e0382d2dd3956c60e

vladbailescu:
Tests were failing due to a missing dependency and improper RT in breadcrumb test content. Fixed those in af4f7c9 but I have two concerns:
- the implementation picker seems to be more restrictive now, makes we wonder if this will hit some customers that have badly configured content/RTs
- build time has almost doubled on my PC, could be related to added dependencies

bpauli:
@vladbailescu I found another approach to create a test case with an inner dummy class which implements the Page interface. So I reverted the dependency update and test content changes.
The implementation picker will now only filter models wich are starting with `com.adobe.cq`. Customers with a different namespace should therefore not be effected.

vladbailescu (approved):

PR #2252, opened 2022-07-12
[DataLayer] Add a configuration to not include the data layer client …

### LINKED ISSUE #2250: [DataLayer] Add a configuration to not include the data layer client library
It is sometimes desired to include the data layer client library through Adobe Launch. In this case a property which applies to the site is needed to disable the inclusion of the library through the Page component.

jckautzmann:
…library

- introduce a context-aware property to the existing DataLayerConfig to disable the inclusion of the data layer client library in the Page component

| Fixed Issues?            | Fixes #2250 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/DataLayerConfig.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/DataLayerConfig.java
@@ -31,4 +31,12 @@ public @interface DataLayerConfig {
     @Property(label="Data Layer enabled")
     boolean enabled() default false;
 
+    /**
+     *
+     * @return {@code true} if the data layer client library is not included through the Core Page component,
+     * {@code false} otherwise. It defaults to {@code false} (client library included).
+     */
+    @Property(label="Data Layer client library not included")
+    boolean noClientlibIncluded() default false;
+
 }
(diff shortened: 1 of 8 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/DataLayerConfig.java
```diff
@@ -31,4 +31,12 @@
     @Property(label="Data Layer enabled")
     boolean enabled() default false;
 
+    /**
+     *
+     * @return {@code true} if the data layer client library is not included through the Core Page component,
+     * {@code false} otherwise. It defaults to {@code false} (client library included).
+     */
+    @Property(label="Data Layer client library not included")
+    boolean noClientlibIncluded() default false;
```
vladbailescu, line 40: I would name this `skipClientlibInclude` as I always get confused and have to think twice about the value of `<Something>` when we have something like `no<Something>=true`

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/models/v3/PageImplTest.java
```diff
@@ -89,4 +89,10 @@ protected void testIsClientlibsAsync_undefinedPolicy() throws Exception {
         assertFalse(async);
     }
 
+    @Test
+    protected void testIsDataLayerClientlibIncluded() {
```
vladbailescu, line 93: We should also have a test for skipping the clientlib include with the new property.

vladbailescu (changes requested):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/DataLayerConfig.java
jckautzmann: addressed in https://github.com/adobe/aem-core-wcm-components/pull/2252/commits/7cbbcddaca996a70288f0f6d4e8b8075de9b9b25

vladbailescu (approved):

PR #2261, opened 2022-07-25
SITES-4621 Keep access to unused in between content

tyge68:
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

- Updated ContentFragment model API to expose the list of "par#" that are not displayed anymore when content structure changed (usually when paragraphs are removed).
- Updated the ContentFragmentImpl to compute paragraphs and remainingInBetweenContent information
- Update "templates.html" in order to output remaining in between content at the end.
- Updated all the related tests

Feel free to commit , reuse this code.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
@@ -15,43 +15,34 @@
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.wcm.core.components.internal.models.v1.contentfragment;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-
-import javax.annotation.PostConstruct;
-import javax.inject.Inject;
-
-import org.apache.commons.lang3.StringUtils;
-import org.apache.sling.api.SlingHttpServletRequest;
-import org.apache.sling.api.resource.Resource;
-import org.apache.sling.api.resource.ResourceResolver;
-import org.apache.sling.models.annotations.Exporter;
-import org.apache.sling.models.annotations.Model;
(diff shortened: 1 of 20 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/contentfragment/ContentFragmentImpl.java
```diff
@@ -15,43 +15,34 @@
...
 import com.adobe.cq.export.json.ExporterConstants;
 import com.adobe.cq.wcm.core.components.internal.ContentFragmentUtils;
-import com.adobe.cq.wcm.core.components.util.AbstractComponentImpl;
 import com.adobe.cq.wcm.core.components.internal.models.v1.datalayer.ContentFragmentDataImpl;
 import com.adobe.cq.wcm.core.components.models.contentfragment.ContentFragment;
 import com.adobe.cq.wcm.core.components.models.contentfragment.DAMContentFragment;
 import com.adobe.cq.wcm.core.components.models.datalayer.ComponentData;
 import com.adobe.cq.wcm.core.components.models.datalayer.ContentFragmentData;
 import com.adobe.cq.wcm.core.components.models.datalayer.builder.DataLayerBuilder;
+import com.adobe.cq.wcm.core.components.util.AbstractComponentImpl;
```
vladbailescu, line 30: Could you please revert back the order of imports?

content/src/content/jcr_root/apps/core/wcm/components/contentfragment/v1/contentfragment/templates.html
```diff
@@ -43,6 +43,9 @@ <h3 class="cmp-contentfragment__title">${fragment.title}</h3>
             ${item @ context="html"}
             <div data-sly-resource="${'par{0}' @ format=itemList.count, resourceType=fragment.gridResourceType}"></div>
         </div>
+        <div data-sly-list="${fragment.remainingInBetweenContent}">
```
vladbailescu, line 46: I think this addition should be behind an option so that we don't end up appending these where they would not be desired. IMO, it's best to have a content policy option to enable showing them.

vladbailescu (changes requested):

satzlinger:
@tyge68 hi, any updates on the task? I seems like there is still feedback open and we would like to see that functionally out of the box :)

PR #2267, opened 2022-08-04
[Teaser] The setting of "Title Type" disappeared from Configure Dialog of Teaser component after #2142

### LINKED ISSUE #2266: [Teaser] The setting of "Title Type" disappeared from Configure Dialog of Teaser component after #2142
After the changes for  #2142 where the title/heading type configuration was aligned for all components, the Teaser will no longer allow selection of a title type when `Show title type` is enabled but not allowed headings are configured. Furthermore, if customers had an authored teaser and they edit it without setting the allowed headings, the authored title type/heading will be reset to default.

vladbailescu:
* Added backwards compatibility fallback to old title type selection in case there are no configured allowed types/headings.
* Updated selenium base and fixed previously failing tests
* Added new IT

Fixes #2266

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -66,7 +66,7 @@
         <maven.compiler.source>${aem.java.version}</maven.compiler.source>
         <maven.compiler.target>${aem.java.version}</maven.compiler.target>
         <jackson.version>2.11.3</jackson.version>
-        <aem.selenium.base.it.version>0.2.33-SNAPSHOT</aem.selenium.base.it.version>
+        <aem.selenium.base.it.version>0.2.34</aem.selenium.base.it.version>
     </properties>
 
     <!-- ======================================================================= -->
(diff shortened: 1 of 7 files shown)


parent/pom.xml
```diff
@@ -66,6 +66,7 @@
         <maven.compiler.source>${aem.java.version}</maven.compiler.source>
         <maven.compiler.target>${aem.java.version}</maven.compiler.target>
         <jackson.version>2.11.3</jackson.version>
+        <aem.selenium.base.it.version>0.2.33-SNAPSHOT</aem.selenium.base.it.version>
```
vladbailescu, line 69: Needs https://github.com/adobe/aem-selenium-it-base/pull/25 to be merged and released

Buuhuu (approved):

bpauli (approved):

PR #2274, opened 2022-08-12
SITES-7428: mask Adobe Campaign expression before decoding URLs

### LINKED ISSUE #2271: [Regression] New LinkBuilderImpl fails for links that contain Adobe Campaign expressions
## Bug Report

**Current Behavior**
Links that contain Adobe Campaign expressions like `/path/to/page.html?recipient=<%= recpient.id %>` worked well in the past (incl. rel. 2.20.8). With #2196 the behaviour changed as the `LinkBuilderImpl` now calls `URLDecoder.decode()` on the given url/path which fails with an exception 

```
java.lang.IllegalArgumentException: URLDecoder: Illegal hex characters in escape (%) pattern - Error at index 0 in: "= "
        at java.base/java.net.URLDecoder.decode(URLDecoder.java:232)
        at java.base/java.net.URLDecoder.decode(URLDecoder.java:142)
        at com.adobe.cq.wcm.core.components.internal.link.LinkBuilderImpl.buildLink(LinkBuilderImpl.java:159)
```

**Expected behavior/code**
Links with Adobe Campaign expressions should be supported.

**Environment**
- AEM: CS quickstart (master)
- Core Components version: master
- JRE version: 11

**Possible Solution**
I see two possible solutions:

a) restore the previous behaviour (remove the url decoding again)
b) let the PathProcessor do the decoding so that a email specific path processor could handle the Adobe Campaign expressions gracefully (the `default` implementation could be the `URLDecoder.decode()`)

Add any other context about the problem here. If applicable, add screenshots to help explain.

Buuhuu:
| Fixed Issues?            | Fixes #2271

With #2196 urls get decoded before they are passed to the `PathProcessor`. This fails for links that are not well formed with an `IllegalArgumentException`, for example when they contain Adobe Campaign expressions. While it makes sense to decode urls before processing them, this may be a regression for email components.

With this PR all Adobe Campaign expressions in an url are replaced with placeholders before the url gets decoded. Afterwards the original expressions get restored. This preserves the original behaviour for links that contain Adobe Campaign expressions and still does the decoding.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
@@ -168,8 +168,13 @@ public class LinkBuilderImpl implements LinkBuilder {
                 String maskedPath = mask(path, placeholders);
                 maskedPath = URLDecoder.decode(maskedPath, StandardCharsets.UTF_8.name());
                 path = unmask(maskedPath, placeholders);
-            } catch (UnsupportedEncodingException e) {
-                LOGGER.warn(e.getMessage());
+            } catch (Exception ex) {
+                String message = "Failed to decode url '{}': {}";
+                if (LOGGER.isDebugEnabled()) {
+                    LOGGER.warn(message, path, ex.getMessage(), ex);
+                } else {
+                    LOGGER.warn(message, path, ex.getMessage());
+                }
             }
             String decodedPath = path;
             return pathProcessors.stream()
@@ -308,6 +313,9 @@ public class LinkBuilderImpl implements LinkBuilder {
     /**
      * Masks a given {@link String} by replacing all occurrences of {@link LinkBuilderImpl#PATTERNS} with a placeholder.
      * The generated placeholders are put into the given {@link Map} and can be used to unmask a {@link String} later on.
+     * <p>
+     * For example the given original {@link String} {@code /path/to/page.html?r=<%= recipient.id %>} will be transformed to
+     * {@code /path/to/page.html?r=_abcd_} and the placeholder with the expression will be put into the given {@link Map}.
      *
      * @param original     the original {@link String}
      * @param placeholders a {@link Map} the generated placeholders will be put in
@@ -330,6 +338,9 @@ public class LinkBuilderImpl implements LinkBuilder {
 
     /**
      * Unmasks the given {@link String} by replacing the given placeholders with their original value.
+     * <p>
+     * For example the given masked {@link String} {@code /path/to/page.html?r=_abcd_} will be transformed to
+     * {@code /path/to/page.html?r=<%= recipient.id %>} by replacing each of the given {@link Map}s keys with the corresponding value.
      *
      * @param masked       the masked {@link String}
      * @param placeholders the {@link Map} of placeholders to replace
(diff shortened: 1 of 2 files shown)


herzog31 (approved):

bpauli (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
```diff
@@ -156,7 +162,12 @@ public LinkBuilderImpl(String url, SlingHttpServletRequest req, List<PathProcess
...
         if (StringUtils.isNotEmpty(path)) {
             try {
-                path = URLDecoder.decode(path, StandardCharsets.UTF_8.name());
+                // The link contain character sequences that are not well formatted and cannot be decoded, for example
+                // Adobe Campaign expressions like: /content/path/to/page.html?recipient=<%= recipient.id %>
+                Map<String, String> placeholders = new LinkedHashMap<>();
+                String maskedPath = mask(path, placeholders);
+                maskedPath = URLDecoder.decode(maskedPath, StandardCharsets.UTF_8.name());
+                path = unmask(maskedPath, placeholders);
             } catch (UnsupportedEncodingException e) {
```
Buuhuu, line 171: Maybe we should still catch the IEA and continue processing without decoding. That would also preserve the behaviour in a none-breaking way other, custom cases. Customers that use other markup than the Adobe Campaign expressions would still face the exception.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
```diff
@@ -294,4 +305,63 @@ private Pair<Page, String> resolvePage(@NotNull final Page page) {
         return new ImmutablePair<>(resolved, getPageLinkURL(resolved));
     }
 
+    /**
+     * Masks a given {@link String} by replacing all occurrences of {@link LinkBuilderImpl#PATTERNS} with a placeholder.
+     * The generated placeholders are put into the given {@link Map} and can be used to unmask a {@link String} later on.
+     *
```
jckautzmann, line 311: Could you please add an example to help understand what this method does, e.g. with `/content/path/to/page.html?recipient=<%= recipient.id %>`?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
```diff
@@ -294,4 +305,63 @@ private Pair<Page, String> resolvePage(@NotNull final Page page) {
...
+                masked = masked.replaceFirst(Pattern.quote(expression), placeholder);
+                placeholders.put(placeholder, expression);
+            }
+        }
+        return masked;
+    }
+
+    /**
+     * Unmasks the given {@link String} by replacing the given placeholders with their original value.
+     *
```
jckautzmann, line 333: Could you please add an example to help understand what this method does?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkBuilderImpl.java
```diff
@@ -294,4 +305,63 @@ private Pair<Page, String> resolvePage(@NotNull final Page page) {
...
+    private static String unmask(String masked, Map<String, String> placeholders) {
+        String unmasked = masked;
+        for (Map.Entry<String, String> placeholder : placeholders.entrySet()) {
+            unmasked = unmasked.replaceFirst(placeholder.getKey(), placeholder.getValue());
+        }
+        return unmasked;
+    }
+
+    /**
+     * Generate a new random placeholder that is not conflicting with any character sequence in the the given {@link String}.
```
jckautzmann, line 347: Could you please add an example to help understand what this method does?

Issue #2276, opened 2022-08-16
[Image] Incorrect SVG handling in case of DM enabled


saurabh-khare:
## Bug Report

**Current Behavior**
When Dynamic Media is disabled the image component uses Adaptive Image Servlet to load SVG files. An example path is `/content/wknd/language-masters/en/_jcr_content/root/container/carousel/item_1571954853062.coreimg.svg/1660579584594/flag-of-europe.svg`
Here we can see the extension of the file is .svg hence browser correctly detects the Content-Type: image/svg+xml
However when DM is enabled the SVG file is rendered as JPEG (Content-Type: image/jpeg) from S7 servers. For example 
https://s7g10.scene7.com/is/image/miturdev/comunicazione
this is due to the fact that image component uses /is/image path for image delivery while it should be /is/content.
Infact apart from S7 file type image all other files types should be loaded with /is/content

**Expected behavior/code**
Only S7 files of type image should be loaded via /is/image url. All other files types should be loaded via /is/content.
This will allow files like SVG to be rendered with corrent Content-Type. 

**Environment**
- AEM as a Cloud Service (Release: 2022.6.7904.20220629T070041Z)
- Core Components version 2.20.8
- Java(TM) SE Runtime Environment 18.9 (build 11.0.15.1+2-LTS-10)

**Possible Solution**
The condition in `initModel()` method of ImageImpl must be changed to following

```
if (asset.getMetadataValue(Scene7Constants.PN_S7_TYPE).equals(Scene7AssetType.IMAGE.getValue())) {
                dmServerPath = DM_IMAGE_SERVER_PATH;
            } else {
                dmServerPath = DM_CONTENT_SERVER_PATH;
            }

```

**Additional context / Screenshots**
![image](https://user-images.githubusercontent.com/41549648/184860998-ebe56187-5556-4523-a887-47b0ebfa8faa.png)
<img width="802" alt="Screenshot 2022-08-09 at 14 49 01" src="https://user-images.githubusercontent.com/41549648/184861130-2c543957-fb30-49b3-a762-d46e348eaf3a.png">
<img width="1141" alt="Screenshot 2022-08-09 at 14 48 08" src="https://user-images.githubusercontent.com/41549648/184861155-69d0eb4c-af07-4d6f-bb5b-82207d5636d7.png">

PR #2283, opened 2022-08-25
[Image] Component should load SVGs as content and not image from dyna…

saurabh-khare:
Condition changed on when image component should generate '/is/image' url

| Fixed Issues?            | Fixes https://github.com/adobe/aem-core-wcm-components/issues/2276
| Tests Added + Pass?      | Yes (All tests pass)
| Any Dependency Changes?  |

Only S7 files of type image should be loaded via /is/image url. All other files types should be loaded via /is/content.
This will allow files like SVG to be rendered with corrent Content-Type. Hence DM url generated by image component on DM enabled flag must use '/is/image' only when DM processor explicitly sets type as image or when it cannot be inferred.


bpauli (approved):
LGTM

PR #2301, opened 2022-09-09
[ToC] Skip early if filter not applicable

### LINKED ISSUE #2300: [TOC] Improve exit condition for TableOfContentsFilter
During the analysis of the issue #2229  it was detected that the ToC Filter can be optimised. The filter should fist check if the ToC request attribute is available in the current request before doing anything else.

(SITES-6841)

ky940819:
| Fixed Issues?            | Fixes #2300
| Patch: Bug Fix?          | Patch - optimization
| Tests Added + Pass?      | None added - existing tests pass
| Documentation Provided   | N/A - no behaviour change

Prevents the buffering of the response into a `CharResponseWrapper` if the response contents are never going to be acted upon - because the response is not `text/html` content type, or because it does not include a Table of Contents component.

This is accomplished by adding a couple of guards that bail out of the filter ASAP - passing the unmolested request/response up the filter chain - after having determined that the filter will not act on this information. 

This prevents full-buffering of responses into memory, unless they are going to be modified before being returned to the user.

Side note: I also added null type annotations to methods/parameters to help identify potential NPEs. 
No identified potential NPEs were resolved (but they do exist - according to my IDE).

----
refs #2300

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
@@ -23,8 +23,10 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
+import javax.annotation.Nullable;
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
 import javax.servlet.FilterConfig;
@@ -33,6 +35,7 @@ import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletResponse;
 
+import com.drew.lang.annotations.NotNull;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.sling.servlets.annotations.SlingServletFilter;
 import org.apache.sling.servlets.annotations.SlingServletFilterScope;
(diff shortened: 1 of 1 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/TableOfContentsFilter.java
```diff
@@ -118,48 +123,53 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+            chain.doFilter(request, response);
+            return;
+        }
+
+        if (!Optional.ofNullable(response.getContentType()).filter(ct -> ct.contains("text/html")).isPresent()) {
+            LOGGER.debug("Response content type not \"text/html\", bypassing filter");
+            chain.doFilter(request, response);
+            return;
+        }
+
```
jckautzmann, line 138: I tried it out on a sample page but the TOC renders empty. I believe it's because the request attribute is set later in the chain by the TOC component. Following code would work:
```
        if (!enabled) {
            LOGGER.debug("{} not enabled, bypassing it", TableOfContentsFilter.class.getName());
            chain.doFilter(request, response);
            return;
        }

        CharResponseWrapper responseWrapper = new CharResponseWrapper((HttpServletResponse) response);
        chain.doFilter(request, responseWrapper);
        String originalContent = responseWrapper.toString();

        if (!Optional.ofNullable(response.getContentType()).filter(ct -> ct.contains("text/html")).isPresent()) {
            LOGGER.debug("Response content type not \"text/html\", bypassing filter");
            response.getWriter().write(originalContent);
            return;
        }

        Boolean containsTableOfContents = (Boolean) request.getAttribute(TableOfContentsImpl.TOC_REQUEST_ATTR_FLAG);
        if (containsTableOfContents == null || !containsTableOfContents) {
            LOGGER.debug("request attribute {} not present or set to false, bypassing the filter {}", TableOfContentsImpl.TOC_REQUEST_ATTR_FLAG, TableOfContentsFilter.class.getName());
            response.getWriter().write(originalContent);
            return;
        }

        Document document = Jsoup.parse(originalContent);
```

ky940819:
Closed because this is not a satisfactory solution.

jckautzmann:
@ky940819 I created a pull request at #2305 . Your feedback is welcome.

PR #2302, opened 2022-09-09
[PanelContainer] Make container servlet resourceType aware

bpauli:
- get forced resourceType from request parameter

-->

| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

The container servlet is also used by the segmenteditor of the core email components project. As the segmenteditor relies on a different multifield item structure the container servlet needs to get the resource Type from the request parameter to render the correct output. Related issue of the core email components is https://github.com/adobe/aem-core-email-components/issues/213

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/ContainerServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/ContainerServlet.java
@@ -18,6 +18,9 @@ package com.adobe.cq.wcm.core.components.internal.servlets;
 import com.adobe.cq.wcm.core.components.commons.editor.dialog.childreneditor.Editor;
 import com.adobe.cq.wcm.core.components.internal.models.v1.PanelContainerImpl;
 import com.day.cq.wcm.api.WCMMode;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.sling.api.SlingConstants;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.SlingHttpServletResponse;
 import org.apache.sling.api.request.RequestDispatcherOptions;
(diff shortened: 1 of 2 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/ContainerServlet.java
```diff
@@ -106,7 +109,12 @@ protected void doPost(@NotNull SlingHttpServletRequest request, @NotNull final S
     protected void doGet(@NotNull SlingHttpServletRequest request, @NotNull SlingHttpServletResponse response) throws ServletException, IOException {
         Resource container = request.getResource();
         RequestDispatcherOptions options = new RequestDispatcherOptions();
-        options.setForceResourceType(Editor.RESOURCE_TYPE);
+        String resourceType = request.getParameter(SlingConstants.PROPERTY_RESOURCE_TYPE);
```
Buuhuu, line 112: Should this parameter maybe be limited to a fixed set of allowed resource types? Not sure if this could be misused in some way.

For example we could make the servlet configurable as a factory, and configure the editor resource type together with the `sling.servlet.resourceTypes`. So one servlet instance would handle the requests of the wcm container with the wcm editor, another servlet instance handles the segments component of the email components with the segment editor.

bpauli: @Buuhuu IMHO this is not required. The `ContainerServlet` is already limited to the `panelcontainer` resource type (see https://github.com/adobe/aem-core-wcm-components/blob/main/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/ContainerServlet.java#L50), so I wonder how this introduced parameter can be misused when the `ContainerServlet` can only include a panelcontainer in the get method with various resource types?

vladbailescu (approved):

herzog31 (approved):

PR #2321, opened 2022-10-01
[List] Prevent blank page titles

### LINKED ISSUE #2316: Removing Page's Navigation Title Creates Blank Entry in Navigation Component
## Feature Request

A page's entry in the nav component can be modified by changing its Navigation Title property. That being said, the navigation title is an optional field and, if not set, the page's entry in the nav will be created using the page's Title property. What we found was that if we set a Navigation Title, and then want to unset it to revert to using the page's Title as its entry in the nav component, we cannot. What ends up happening is that once we set and unset the Navigation Title, the page's entry is actually blank. It appears as though once we set and unset the Navigation Title, an empty navigationTitle property is left on the page's jcr:content node. That is interpreted as a purposeful blank Navigation Title instead of one that's been removed with the purpose of resetting to the page Title property.

Adjust the rendering logic such that a blank navigationTitle property is interpreted as an non-existent navigationTitle that defaults back to the page title.

**Are there alternatives?**
The only alternative is to force someone to edit the page's properties manually via crx/de or vault, which is not an Authoring task.

**Documentation**
Should be mostly the same, just adjust so that an empty navigation title property is ignored altogether

ky940819:
| Fixed Issues?            |  Fixes #2316
| Patch: Bug Fix?          | Y
| Documentation Provided   | Yes (updated JavaDoc)

Updates the `PageListItemImpl` so that blank page titles are treated the same as null page titles. This prevents an empty title from being shown in lists if the title contains only whitespace.

Adds a test to verify the sorting of list items that have a whitespace-only title (sorts based on name instead of blank title).

----
Fixes #2316

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageListItemImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PageListItemImpl.java
@@ -16,7 +16,10 @@
 package com.adobe.cq.wcm.core.components.internal.models.v1;
 
 import java.util.Calendar;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
 
+import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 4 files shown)


bpauli (approved):
@ky940819 Thank you for this PR!

PR #2339, opened 2022-10-12
#2338 allow to disable the AmpModeForwardFilter

### LINKED ISSUE #2338: ability to disable the AmpModeForwardFilter
## The problem

The AmpModeForwardFilter can incur significant overhead, because if no ampMode is specified for a requested page, always the ContentPolicyManager is requested for a policy for the requested resource. And that is not a cheap operation because it needs to read a lot of resources to build-up this policy. 

This is a problem if no AMP is used at all. This is also a problem if Core Components are not used at all, because this filter is a global one, affecting every user of AEM Cloud Service, because Core Components are installed there.

The delay of it can be significant; this is an excerpt from ```/system/console/status-slingfilter```:

```
0 : class com.adobe.cq.wcm.core.extensions.amp.internal.AmpModeForwardFilter (id: 3208, property: none); called: 9117; time: 2142ms; time/call: 234µs
```
which makes it (one of) the slowest filters in the system.

## The feature Request
There should be a chance to disable the filter. I understand that changing the policy (this filter being disabled by default) would cause backward-compatibility problems, so we need to give at least users a chance to disable it if they don't need it.

joerghoh:
| Fixed Issues?            | Fixes #2338 

Allow to disable the AmpModeForwardFilter to reduce execution time.

### DIFF
--- a/extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
+++ b/extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
@@ -32,7 +32,11 @@ import org.apache.sling.api.request.RequestDispatcherOptions;
 import org.apache.sling.api.servlets.HttpConstants;
 import org.apache.sling.engine.EngineConstants;
 import org.osgi.framework.Constants;
+import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.metatype.annotations.AttributeDefinition;
+import org.osgi.service.metatype.annotations.Designate;
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -52,9 +56,12 @@ import static com.adobe.cq.wcm.core.extensions.amp.internal.AmpUtil.*;
                 Constants.SERVICE_RANKING + "Integer=1000"
         }
 )
+@Designate(ocd = AmpModeForwardFilter.Config.class)
 public class AmpModeForwardFilter implements Filter {
 
     private static final Logger LOG = LoggerFactory.getLogger(AmpModeForwardFilter.class);
+    
+    boolean isEnabled;
 
     /**
      * @see Filter#doFilter(ServletRequest, ServletResponse, FilterChain)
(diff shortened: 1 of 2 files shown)


extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
```diff
@@ -116,4 +125,20 @@ public void init(final FilterConfig config) {}
      */
     @Override
     public void destroy() {}
+    
+    
+    @Activate
+    public void activate (Config config) {
+    	this.isEnabled = config.enabled();
+    	LOG.debug("AmModeForwardFilter isEnabled={}",isEnabled);
```
kwin, line 133: AmModeForwardFilter -> AmpModeForwardFilter

extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
```diff
@@ -116,4 +125,20 @@ public void init(final FilterConfig config) {}
      */
     @Override
     public void destroy() {}
+    
+    
+    @Activate
```
kwin, line 130: why not using constructor injection with overloaded constructors (one taking Config, calling the other one taking just the boolean flag). That eases testing a lot.

bpauli (approved):
@joerghoh Thank you for the hands up!

extensions/amp/bundle/src/test/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilterTest.java
```diff
@@ -164,5 +168,32 @@ void testFilteringWithSelectorWithAmpMode() throws IOException, ServletException
         spyAmpModeForwardFilter.doFilter(context.request(), context.response(), filterChain);
         verify(spyAmpModeForwardFilter, atLeastOnce()).doFilter(eq(context.request()), eq(context.response()) , eq(filterChain));
     }
+    
+    
+    
+    @Test
+    void testWithDisabledFilter() throws IOException, ServletException {
+    	AmpModeForwardFilter filter = Mockito.spy(new AmpModeForwardFilter(new AmpModeForwardFilter.Config() {
```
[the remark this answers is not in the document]
joerghoh: I wanted to call the only public constructor (which is also used by OSGI) at least once. Otherwise if this constructor is removed (for whatever reason) no test will fail.

extensions/amp/bundle/src/main/java/com/adobe/cq/wcm/core/extensions/amp/internal/AmpModeForwardFilter.java
joerghoh: done

kwin (approved):
Much nicer now.

PR #2344, opened 2022-10-18
SITES-8730: add integration tests to cover teaser corner cases

Buuhuu:
This adds an integration test that verifies the teaser markup. In particular it tests the teaser v1 and teaser v2 markup with a CTA action referencing an Asset.

### DIFF
--- a/testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/ComponentsIT.java
+++ b/testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/ComponentsIT.java
@@ -17,13 +17,18 @@ package com.adobe.cq.wcm.core.components.it.http;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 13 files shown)


testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/ComponentsIT.java
```diff
@@ -0,0 +1,122 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2021 Adobe
```
jckautzmann, line 2: Should be 2022

jckautzmann:
LGTM, thx! I would also add tests to cover CTAs linking to assets in Teaser v1 and v2, as theses use cases also trigger the Class Cast Exception.

jckautzmann (approved):

Issue #1473, opened 2021-03-29
Dynamic Media for all components


frappierer:
## Feature Request

Currently Dynamic Media is only available for the image component:
https://github.com/oufimtse/aem-core-wcm-components/blob/master/content/src/content/jcr_root/apps/core/wcm/components/image/v2/image/README.md

I would like to see the same set of features in:

* List or Image List Component
* Teaser Component
* In theory even in the Text Component since i can link images there as well.

The goal is, to have dynamic media handling all of you website images.

Best

kpolychr:
I can only upvote this; one of the main reasons of clients purchasing DM is to have features like smart cropping in widely-used components like teasers/heros, saving tons of authoring time. Any indication on when this enhancement can become available?

PR #2350, opened 2022-10-21
Add DM support to Teaser component

bpauli:
- provide servlet to include asset tab from image component into the the teaser dialog. In this case the cqDesign request attribute needs to be reset to the imageDelegate component
- move the DM settings from the metadata tab to the asset tap in the image dialog, so it is still only required to include the asset tab within the teaser dialog

relates to #1473


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/ImageDelegatePolicyServlet.java
```diff
@@ -0,0 +1,79 @@
...
+            String imageDelegate = properties.get(AbstractImageDelegatingModel.IMAGE_DELEGATE, String.class);
+            RequestDispatcher requestDispatcher = Optional.ofNullable(request.getResource().getValueMap().get(PN_PATH
+                            , String.class))
+                    .map(p -> new WCMFilteringResourceWrapper(resourceResolver.getResource(p), new SyntheticResource(resourceResolver,
+                            requestPathInfo.getSuffix(),
+                            imageDelegate), expressionResolver, request)).map(request::getRequestDispatcher).orElse(null);
+            if (Objects.nonNull(requestDispatcher)) {
+                requestDispatcher.include(request, response);
+            }
+        }
```
Buuhuu, line 77: Should we send a 404 here when any of the conditions are not met?

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/servlets/ImageDelegatePolicyServletTest.java
```diff
@@ -0,0 +1,71 @@
...
+import com.adobe.granite.ui.components.ExpressionCustomizer;
+import com.adobe.granite.ui.components.ExpressionResolver;
+import com.day.cq.wcm.api.designer.Style;
+import io.wcm.testing.mock.aem.junit5.AemContext;
+import io.wcm.testing.mock.aem.junit5.AemContextExtension;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+@ExtendWith({AemContextExtension.class, MockitoExtension.class})
+class ImageDelegatePolicyServletTest {
```
Buuhuu, line 39: We should include a test, that the request dispatcher is called and verify the arguments.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/ImageDelegatePolicyServlet.java
bpauli: added in 8d1bee90c0c6ddcfcd9703bb820c7196a1029be5

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/servlets/ImageDelegatePolicyServletTest.java
bpauli: added in 8d1bee90c0c6ddcfcd9703bb820c7196a1029be5

content/src/content/jcr_root/apps/core/wcm/components/commons/site/clientlibs/utils/js/utils.js
```diff
@@ -77,9 +77,11 @@
             return transformedProperties;
         };
 
+        document.dispatchEvent(new CustomEvent('core.wcm.components.commons.site.utils.loaded'));
```
Buuhuu: This will immediately call the subscriber of the event, before the object returned below actually gets assigned to the global variable. 

The event should be fired outside of this function.

PR #2360, opened 2022-11-06
#2352 calculate datalayer check only once

### LINKED ISSUE #2352: Unnecessary dataLayer checks on every component
I am currently checking all areas which do a repository access during page rendering. I found that on a standard WKND page this call / these calls are responsible for 5-7% of all requests towards the Sling Repository/Oak.

```
 com.adobe.cq.wcm.core.components.util.ComponentUtils.isDataLayerEnabled
```
Such a call uses caconfig to find the result, and does quite a number of repository accesses itself, but that's a different story.

The implementation of this method checks on the page level; that means for the whole page the result is identical. But I found a number of distinct call graphs, which all call the above mentioned method, for example:

```
        at com.adobe.cq.wcm.core.components.util.ComponentUtils.isDataLayerEnabled(ComponentUtils.java:71) [com.adobe.cq.core.wcm.components.core:2.21.2]
        at com.adobe.cq.wcm.core.components.util.AbstractComponentImpl.getData(AbstractComponentImpl.java:140) [com.adobe.cq.core.wcm.components.core:2.21.2]
        at jdk.internal.reflect.GeneratedMethodAccessor2091.invoke(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.apache.sling.scripting.sightly.render.ObjectModel.invokeBeanMethod(ObjectModel.java:448) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at org.apache.sling.scripting.sightly.render.ObjectModel.resolveProperty(ObjectModel.java:131) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at org.apache.sling.scripting.sightly.render.AbstractRuntimeObjectModel.getProperty(AbstractRuntimeObjectModel.java:153) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at org.apache.sling.scripting.sightly.impl.engine.runtime.SlingRuntimeObjectModel.getProperty(SlingRuntimeObjectModel.java:103) [org.apache.sling.scripting.sightly:1.4.20.140]
        at org.apache.sling.scripting.sightly.render.AbstractRuntimeObjectModel.resolveProperty(AbstractRuntimeObjectModel.java:75) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at libs.core.wcm.components.page.v3.page.headlibs__002e__html$1.render(headlibs__002e__html.java:181) [aem-precompiled-scripts:6.6.0.SNAPSHOT]
```

or

```
        at com.adobe.cq.wcm.core.components.util.ComponentUtils.isDataLayerEnabled(ComponentUtils.java:71) [com.adobe.cq.core.wcm.components.core:2.21.2]
        at com.adobe.cq.wcm.core.components.util.AbstractComponentImpl.getData(AbstractComponentImpl.java:140) [com.adobe.cq.core.wcm.components.core:2.21.2]
        at jdk.internal.reflect.GeneratedMethodAccessor2091.invoke(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.apache.sling.scripting.sightly.render.ObjectModel.invokeBeanMethod(ObjectModel.java:448) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at org.apache.sling.scripting.sightly.render.ObjectModel.resolveProperty(ObjectModel.java:131) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at org.apache.sling.scripting.sightly.render.AbstractRuntimeObjectModel.getProperty(AbstractRuntimeObjectModel.java:153) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at org.apache.sling.scripting.sightly.impl.engine.runtime.SlingRuntimeObjectModel.getProperty(SlingRuntimeObjectModel.java:103) [org.apache.sling.scripting.sightly:1.4.20.140]
        at org.apache.sling.scripting.sightly.render.AbstractRuntimeObjectModel.resolveProperty(AbstractRuntimeObjectModel.java:75) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at libs.core.wcm.components.languagenavigation.v2.languagenavigation.languagenavigation__002e__html.render(languagenavigation__002e__html.java:49) [aem-precompiled-scripts:6.6.0.SNAPSHOT]
```

or

```
        at com.adobe.cq.wcm.core.components.util.ComponentUtils.isDataLayerEnabled(ComponentUtils.java:71) [com.adobe.cq.core.wcm.components.core:2.21.2]
        at com.adobe.cq.wcm.core.components.util.AbstractComponentImpl.getData(AbstractComponentImpl.java:142) [com.adobe.cq.core.wcm.components.core:2.21.2]
        at jdk.internal.reflect.GeneratedMethodAccessor2091.invoke(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.apache.sling.scripting.sightly.render.ObjectModel.invokeBeanMethod(ObjectModel.java:448) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at org.apache.sling.scripting.sightly.render.ObjectModel.resolveProperty(ObjectModel.java:131) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at org.apache.sling.scripting.sightly.render.AbstractRuntimeObjectModel.getProperty(AbstractRuntimeObjectModel.java:153) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at org.apache.sling.scripting.sightly.impl.engine.runtime.SlingRuntimeObjectModel.getProperty(SlingRuntimeObjectModel.java:103) [org.apache.sling.scripting.sightly:1.4.20.140]
        at org.apache.sling.scripting.sightly.render.AbstractRuntimeObjectModel.resolveProperty(AbstractRuntimeObjectModel.java:75) [org.apache.sling.scripting.sightly.runtime:1.2.6.140]
        at libs.core.wcm.components.languagenavigation.v2.languagenavigation.item__002e__html$1.render(item__002e__html.java:90) [aem-precompiled-scripts:6.6.0.SNAPSHOT]
```
and many more.

I don't think that it is necessary to compute this information multiple times; instead the result of a initial call to ```ComponentUtil.isDataLayerEnabled``` should be cached for the whole page. This would prevent a lot of repository access and speed up the page rendering time.

joerghoh:
| Fixed Issues?            | Fixes #2352
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

Execute the checks if the dataLayer is enabled just once per request (instead of once per every component). The information is stored in a request attribute.

This patch avoids a lot of repository accesses, because each of these individual checks reaches out into caconfig which does a lot of repository access.


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ComponentUtils.java
```diff
@@ -53,24 +54,50 @@ public final class ComponentUtils {
...
     }
 
+    /**
+     * Check if data layer is enabled.
+     *
+     * @param request  The request in which context this happens
+     * @param resource The resource to check.
+     * @return True if data layer is enabled for this resource, false otherwise.
+     */
+    public static boolean isDataLayerEnabled(@Nullable SlingHttpServletRequest request,
```
bpauli, line 74: I think this can lead to some misbehaviour, as the CA configuration is resolved based on the resource path and the value is stored in the current request. It would be possible to include a resource into the current request which is located into a path were the CA configuration is different as for the path the current request is made against.
Example:
Request is against `/content/page1` where data layer is enabled. page1 includes a resource from `/content/page2` where data layer is disabled. In this case the resource from page2 would not have the data layer enabled. This might be fine but can also lead to some unexpected behaviour.
I wonder if there is no caching layer in the CA config resolver. If not, it would be a nice enhancement IMHO.

bpauli: Yes, I think the datalayer on/off switch should be based on the context of the current page. But the page can have a child resource which makes a include on a different context setting. At this is IMHO a problem which can't be solved by just checking for a request attribute.

joerghoh: Sure. I am just wondering if a per-component setting would make sense. A requested page has a single datalayer object, which is dependent on the page (and to some extent on its content) itself. In my opinion it does not really make sense to allow every included object to come with its own datalayer "policy".

Because it would be extremely hard to debug for an author, why a certain piece of the page, which should be reflected in the datalayer, is not reflected but present in the page. Or the other way around. Because it would force the author to check all embedded elements.

I think that it's the easiest to explain and understand if we have a single per-page setting (can be inherited and overriden) to define if the datalayer is enabled or not. And then just follow that policy on that page, no matter where the data rendered on the page comes from or if the included object has a different policy defined.

bpauli (approved):
@joerghoh For the data layer case I think we can use this approach. But as discussed via Slack, it would be great if we can start a discussion also on the Sling mailing list and maybe get also some general optimisation for the CA Config resolver on sling level.
cc @stefanseifert

PR #2372, opened 2022-11-10
Fix project not building with Java 17 (issue #2357)

### LINKED ISSUE #2357: Review: Build does not work with Java 17
## Bug Report

**Current Behavior**
When running a build with Java 17, the tests fails with exceptions like this:

```
Caused by: java.lang.IllegalArgumentException: Unsupported class file major version 61
	at net.bytebuddy.jar.asm.ClassReader.<init>(ClassReader.java:184)
	at net.bytebuddy.jar.asm.ClassReader.<init>(ClassReader.java:166)
	at net.bytebuddy.jar.asm.ClassReader.<init>(ClassReader.java:152)
	at net.bytebuddy.utility.OpenedClassReader.of(OpenedClassReader.java:82)
	at net.bytebuddy.dynamic.scaffold.TypeWriter$Default$ForInlining.create(TypeWriter.java:3390)
	at net.bytebuddy.dynamic.scaffold.TypeWriter$Default.make(TypeWriter.java:1930)
	at net.bytebuddy.dynamic.scaffold.inline.RedefinitionDynamicTypeBuilder.make(RedefinitionDynamicTypeBuilder.java:207)
	at net.bytebuddy.dynamic.scaffold.inline.AbstractInliningDynamicTypeBuilder.make(AbstractInliningDynamicTypeBuilder.java:116)
	at net.bytebuddy.dynamic.DynamicType$Builder$AbstractBase.make(DynamicType.java:3397)
	at org.mockito.internal.creation.bytebuddy.InlineBytecodeGenerator.transform(InlineBytecodeGenerator.java:254)

```

**Expected behavior/code**
well, it should build :-)

**Environment**
- Main branch
- JDK 17

**Possible Solution**

Mockito needs to get updated to at least version 4.4.0 (right now the version is a very old 2.24.5 :-(

ky940819:
| Fixed Issues?            | Fixes #2357 
| Patch: Bug Fix?          | Y
| Any Dependency Changes?  | Yes (many)

----

The goal of this request is to allow the project to build under Java 17.

It is *not* a goal of this request to ensure that this project functions properly in an AEM runtime under Java 17, as Java 17 is not currently a supported version of Java. However, successfully building and testing is a step closer to compatibility with Java 17, making it easier to implement whenever Java 17 becomes a supported version for AEM.
This request does not change the target or source to Java 17 – it is concerned only with the ability for the build/test to complete successfully when running Java 17.

To accomplish this, the bulk of the changes are updates to maven dependencies. These dependency updates sometimes require application code or test code changes, and frequently cause a cascade of other required dependency changes. 
Since so many dependencies are required to be updates, this opportunity is used to perform opportunistic updates to all project dependencies. The project dependencies are updated such that:

1. If the dependency is `provided`
   * The dependency is updated to the latest version present in AEM 6.5.13.0 (the stated minimum required version of AEM for this project)
2. If the dependency scope is `test`
   * If the dependency is an implementation of a bundle that is provided in AEM
      * The dependency is updated to the latest version present in AEM 6.5.13.0
   * If the dependency is purely a testing utility or library
      * The dependency is updated to the latest possible version

I did not make meaningful updates to dependencies for modules found under the `it` folder, as I am not familiar enough with those modules to be confident in changing them. However, It is possible that they picked up changes from the parent pom.

# Test Dependency Updates
The following are the TEST dependencies that are updated. This includes all dependencies with the scope `test` as well as all dependencies in the "junit.core" module – because these dependencies are only ever used in the test context, never in runtime.

|Artifact|Old Version|New Version|
|--------|-----------|-----------|
|ch.qos.logback:logback-classic|N/A|1.2.9|
|com.adobe.cq:com.adobe.cq.export.json|0.0.12|0.1.10 (provided - junit bundle)|
|com.google.guava:guava|20.0.|Removed|
|com.sun.xml.bind:jaxb-core|2.2.11|2.3.0|
|com.sun.xml.bind:jaxb-impl|2.2.11|2.3.7|
|io.wcm:io.wcm.testing.aem-mock.junit4|3.1.0|Removed|
|io.wcm:io.wcm.testing.aem-mock.junit5|3.1.0|5.1.0|
|junit:junit|4.13.1|Removed|
|org.apache.commons:commons-collections4|4.1.|Removed|
|org.apache.sling:org.apache.sling.i18n|2.4.4|2.4.14|
|org.apache.sling:org.apache.sling.models.api|1.3.6|1.3.8 (provided - junit bundle)|
|org.apache.sling:org.apache.sling.models.impl|1.4.10|Removed|
|org.apache.sling:org.apache.sling.testing.caconfig-mock-plugin|1.3.2|1.4.0|
|org.apache.sling:org.apache.sling.testing.resourceresolver-mock|1.1.22|1.4.0 (provided - junit bundle)|
|org.apache.sling:org.apache.sling.testing.sling-mock.junit4|2.3.4|Removed|
|org.apache.sling:org.apache.sling.testing.sling-mock-oak|2.1.6|Removed|
|org.javassist:javassist|3.23.1-GA|3.29.2-GA|
|org.junit.jupiter:junit-jupiter|N/A|5.9.1|
|org.junit.jupiter:junit-jupiter-api|5.3.2|Removed|
|org.junit.jupiter:junit-jupiter-engine|5.3.2|Removed|
|org.junit.jupiter:junit-jupiter-params|5.3.2|Removed|
|org.junit.platform:junit-platform-commons|1.5.2|Removed|
|org.mockito:mockito-core|2.24.5|4.8.1|
|org.mockito:mockito-junit-jupiter|2.24.5|4.8.1|
|org.powermock:powermock-api-mockito2|2.0.0|Removed|
|org.powermock:powermock-core|1.6.4|Removed|
|org.powermock:powermock-module-junit4|1.6.4|Removed|
|org.reflections:reflections|0.9.9|0.10.2|
|org.slf4j:slf4j-nop|1.7.6|1.7.25|

Note that Junit4 is entirely removed, along with all associated junit4 test dependencies, since they are no longer in use. 
All powermock dependencies are removed because they are no longer in use.

The `guava` version 20.0 test dependency is removed because version 15 exists as a dependency already (provided in AEM), none of the use of `guava` explicitly requires version 20, and it can be confusing that the outdated version 15 is used in runtime and the outdated version 20 is used in testing.

The `commons-collections4` version 4.1 test dependency is removed because version 4.4 is already available as a non-test dependency (provided in AEM).

The `junit-jupiter-(api|engine|params)` dependencies are removed only because these are combined into the single `junit-jupiter` dependency in the new version of junit.

# Runtime Dependency Updates
The following PROVIDED dependencies are updated. The minimum AEM version is indicated as a sanity check to ensure that no bundle is upgraded beyond the currently supported minimum version of AEM (6.5.13). Some of these dependencies could be updated further if AEM 6.5.14.0 were to be set as the minimum supported version – and likely others could also be updated if the minimum version is set to 6.5.15.0 when it is released on November 24 2023.

|Artifact|Old Version|New Version|Min AEM|
|--------|-----------|-----------|-------|
|org.apache.sling:org.apache.sling.servlets.annotations|1.2.4|1.2.6|N/A|
|org.slf4j:slf4j-api|1.7.6|1.7.25|<6.5.11|
|com.fasterxml.jackson.core:jackson-annotations|2.11.3|2.11.4|<6.5.11|
|com.fasterxml.jackson.core:jackson-databind|2.11.3|2.11.4|<6.5.11|
|com.fasterxml.jackson.core:jackson-core|2.11.3|2.11.4|<6.5.11|
|org.apache.sling:org.apache.sling.sitemap|1.0.2|1.0.6|6.5.13|
|javax.xml.bind:jaxb-api|2.1.|2.3.0|6.5.11|
|org.apache.jackrabbit:jackrabbit-jcr-commons|2.12.1|2.20.4|6.5.13|
|org.apache.commons:commons-lang3|3.5.|3.12.0|<6.5.11|
|commons-codec:commons-codec|1.10.|1.14.|<6.5.11|
|org.apache.httpcomponents:httpclient-osgi|4.5.4|4.5.12|6.5.13|
|com.adobe.cq.wcm:com.adobe.aem.wcm.seo|1.0.4|1.0.10|6.5.13|

# Application Code Changes
## CoreResourceWrapper
The `CoreResourceWrapper` currently requires a base property map of type `Map<String, String>`. This is overly restrictive because `ValueMap` extends `Map<String, Object>`. As a consequence of this, there are several places in the application code that override properties via the `CoreResourceWrapper`, setting the value to `"true"` or `"false"` and relying on the implicit type conversion to perform the correction from type String -> Boolean.

This was likely done because the `CoreResourceWrapper` originally only needed to override String property values – and then was subsequently expanded to override Boolean values as well. 

Although not optimal, this typically doesn’t cause a problem. However, when building under Java 17 this causes a test failure due to the inability to convert the type from String -> Boolean (it is treated as if it is not set). This is likely caused by FELIX-6157 (https://issues.apache.org/jira/browse/FELIX-6157). 

Changing the underlying property map to accept an Object as a property value corrects the testing issues under Java 17, as well as being technically correct.

This is the only code change that affects the application during runtime.

# Test Code Changes
Some of the code changes made to tests are required for testing under Java 17. Largely, this is due to changes in reflection handling in Java 17. Other test changes are required due to upgraded test dependencies (such as `Mockito` and `aem-mocks`).

## Image V3 Test Data
The test content json data for the Image V3 component tests is updated to replace the use of `"true"` and `"false"` strings with the Boolean `true` and `false`. This is done for the same reason as the update to the` CoreResourceWrapper`. 

Although this change was made purely to get the tests to pass in Java 17, I feel that the change is acceptable because the purpose of the tests is to test the Image V3 model, and not to test the implied type conversion of valuemap properties. 

## OEmbedClientImplTest
Removed use of `org.mockito.internal.util.reflection.FieldSetter` - which was both deprecated and subsequently removed from Mockito.

## AmpModeForwardFilterTest
Use of deprecated `Mockito.any()` is replaced with `ArgumentMatchers.any()`.

## MockExternalizer
The `MockExternalizer` class provided in the `junit.core` bundle is marked as deprecated.

This is done because a mock `Externalizer` service is now provided in `aem-mocks` and so it should be used instead.

The `CoreComponentsMockTest` class in the `aem-mock-plugin` bundle is updated to use the `Externalizer` service from `aem-mocks` instead of the (now deprecated) `MockExternalizer`. An override to the default (aem-mocks provided) `Externalizer` publish host is made so as to preserve the behaviour of mapping paths to `https://example.org` instead of the default `http://localhost`.

## List Item Datalayer `@type` values (in tests)
Currently, in the tests of datalayer values for list items, the `@type` value is given as `cq:Component/item` or `cq:Component/cta` (in the case of teasers). This value is incorrect and was the result of a limitation in the mocks. This limitation has been rectified in more recent versions of `aem-mocks` and so the tests are updated to check the proper `<resourceType>/item` or `<resourceType>/cta`.

## Log Tests
There are currently two cases where specific log messages are tested (`AdaptiveImageServletTest` and `InheritedFieldTest`). These cases use `Utils. mockLogger(Class, String)` to create a mock `Logger` and inject it into the `private static final Logger LOGGER` field in the class being tested. 

This type of reflection breaks Java 17 builds. 

To rectify this, logging is tested by adding an in-memory log `Appender`, which can capture the logs during the normal logging process, and then that appender is checked for the presence of the desired log entry.

This pattern does not require injection at all; however, it does require a custom `logback.xml` file in the test resources, as well as the addition of the `ch.qos.logback:logback-classic:1.2.9` test dependency.

## AdaptiveImageServletMappingConfigurationConsumerTest
The order in which services are registered during the test is changed to ensure that the OSGI component that does not implement/service an interface is registered last. 

This is required due to a current known limitation in `osgi-mocks`.

See SLING-11668 (https://issues.apache.org/jira/browse/SLING-11668)

# Plugin Updates
Most plugins were not updated. The only plugin updates are the trivial update of `jacoco` from 0.8.7 -> 0.8.8, and `spotbugs` from `4.0.4` -> `4.7.2.1`. In the case of `spotbugs` this was required to enable the build to complete under Java 17. This update causes many `EI_EXPOSE_REP` and `EI_EXPOSE_REP2` bugs to be spotted. These should probably be resolved at some point; however, for the time being they were suppressed via the `findbugs-exclude.xml` configuration file.

See https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#ei-may-expose-internal-representation-by-returning-reference-to-mutable-object-ei-expose-rep.


bpauli (approved):
LGTM!

Tested against AEM 6.5.13 and latest SDK

PR #2416, opened 2022-12-22
[Containers] Separate concept of "List Items" from "Container Items" #1271

### LINKED ISSUE #1031: [Container] Models should allow access item resources
## Feature Request

The current container models - i.e. those that implement Container interface / extend AbstractContainer class - do not allow access to the container item resources. 

Rather, these containers return `List<ListItem>` from their `Container.getItems()` method.
Since the `ListItem` interface provides only the resource name/path, we are forced into re-resolving the resource via the path to make use of it.

There are two reasons why this is not optimal. The first is that we already have the resource, using the path to re-resolve the resource is just extra (needless) work. The second reason is that the resource accessed by using the path may not be the resource that was used to construct the `ResourceListItemImpl`. 

There is a common misconception that the following statement is true
```java
Resource resource = // some resource
resource.getResourceResolver().getResource(resource.getPath()).equals(resource)
``` 
The above statement may be true in the JCR world, but it is not guaranteed to be true in sling.
In the above example, `resource` may have been a wrapped or synthetic resource, while the result of `resourceResolver().getResource(resource.getPath())` would, assuming no other ResourceProviders are in play, likely be a JCR backed resource that may or may not be equivalent.

An example of where this causes serious issues is when dealing with TemplatedResources, which create a synthetic tree that is the merger of the template resources and the page resources. 

This issue can make it very difficult to use container components in template structure where it may have nested containers with a mixture of editable and non-editable children. This is certainly true for layout containers, but I can also foresee a situation where someone may want to have tabs, accordions, or carousels where some of the panels are part of the template structure and other panels are editable by the author in-page. (for transparency, i haven't actually checked that this doesn't work for tabs/accordions/carousels, but i would be surprised if it did, as it does not for the layout container)

There a couple ways this can be resolved. 

The container components item lists are already backed by `ResourceListItemImpl`, which guarantees that the resource does exist and is known. An interface could be added:
```JAVA
public interface ResourceListItem extends ListItem {
    @NotNull
    public Resource getResource();
}
```
and then the `Container.getItems()` method could be changed to return `ResourceListItem` instead of `ListItem`. This isn't a lot of work, but may introduce API changes that are breaking. Probably not API changes that impact the HTL because ResourceListItems would still be ListItems; but projects using these methods in Java might have to update their APIs because List<ListItem> is not the same as List<ResourceListItem> even if ResourceListItem isa ListItem.

Alternatively, the getChildren() method can be added to the `Container` interface and made public in `AbstractContainerImpl.getChildren()`. This would be backwards compatible, but isn't as clean if a consumer needs access to both the information in the ListItem and the resource within the same context.

**Documentation**
N/A

----
side note: 
This is related to #999 which cannot be resolved in the "simple" container without access to the templated resource, and cannot be resolved in the "responsive" layout without the issue being resolved in `wcm/foundation/components/responsivegrid`.

Other containers can be resolved along the same lines once the API makes the nessisary information available.

bpauli:
| Fixed Issues?            | Fixes #1031 
| Any Dependency Changes?  |


bundles/core/src/test/resources/tabs/test-content.json
```diff
@@ -157,6 +157,7 @@
                             "jcr:primaryType"   : "nt:unstructured",
                             "jcr:title"         : "Tab 1",
                             "jcr:description"   : "Tab 1 description",
+                            "jcr:lastModified"  : "Wed Jul 27 2016 16:17:43 GMT+0300",
```
vladbailescu, line 160: We should remove this

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/PanelContainerListItemImpl.java
```diff
@@ -0,0 +1,57 @@
...
+import com.adobe.cq.wcm.core.components.commons.link.LinkManager;
+import com.adobe.cq.wcm.core.components.models.ListItem;
+import com.day.cq.commons.jcr.JcrConstants;
+import com.day.cq.wcm.api.Page;
+import com.day.cq.wcm.api.components.Component;
+
+/**
+ * Panel container item implementation.
+ */
+@Deprecated
```
vladbailescu, line 32: Do we create this as deprecated?

bpauli: It was deprecated in the origin PR from @ky940819. In the commit he said
>Deprecate `PanelContainerListItemImpl`
>This `ListItem` implementation is only used by panel container
>components. Panel container components will now prefer to return
>a `PanelContainerItem`, which gives direct access to the resource,
>as well as any other panel metadata that is needed.

bundles/core/src/test/resources/tabs/test-content.json
bpauli: removed in last commit

vladbailescu (approved):

PR #2441, opened 2023-02-04
replace @Inject annotation with the specific ones

joerghoh:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

Remove the use of the ```@Inject``` annotation and replace it with the more specific injector-specific one.


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java
```diff
@@ -81,11 +82,11 @@ public class ClientLibrariesImpl implements ClientLibraries {
     @Named(OPTION_RESOURCE_TYPES)
     Object resourceTypes;
 
-    @Inject
+    @ValueMapValue
```
kwin, line 85: I think using the element "name" instead of annotation @Named is even more concise.

bpauli (approved):

bpauli (changes requested):
@joerghoh Can you please have a look at the failing tests? Thanks!

joerghoh:
@bpauli done.

bpauli (approved):

Issue #2508, opened 2023-05-23
[Tabs] Broken aria-controls HTML attribute for tabs


rubnig:
## Bug Report

**Current Behavior**
In the tabs component, the tabs are having an `aria-controls` attribute  to non existing tabpanel ids. 

https://www.aemcomponents.dev/content/core-components-examples/library/core-content/tabs.html#tabs-763e4cccf7-item-c2ef7f1373-tab

![image](https://github.com/adobe/aem-core-wcm-components/assets/8969453/ea5d5af0-66a7-4fe6-9c35-4368a72e16e0)

This was previously fixed in https://github.com/adobe/aem-core-wcm-components/commit/7f6bccda4e9735fcde89766d769d89f93bed5644 but undone in https://github.com/adobe/aem-core-wcm-components/commit/a4418ab5a1b40d432432b3f904bf8c391a4940f1

**Expected behavior/code**
`aria-controls` HTML attribute should point to existing tab panels

**Environment**
- AEMaaCS 2023.5.11983.20230511T173830Z-230200
- Core Components version 2.22.10
- Oracle JDK 11.0.19

**Possible Solution**
PR:

PR #2509, opened 2023-05-23
[2588] Fix aria-controls label for tabs component

rubnig:
Fixes https://github.com/adobe/aem-core-wcm-components/issues/2508


HitmanInWis:
Looking at https://github.com/adobe/aem-core-wcm-components/blob/main/content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/tabs.html it seems this fix has already been implemented.  I'm not sure why the diff for the patch still shows a diff to the `main` branch.

PR #2533, opened 2023-06-23
[2530] AdaptiveImage: control skipping/requiring transformation via OSGi config

### LINKED ISSUE #2530: AdaptiveImage: Configurability of when to apply transformation by file type
## Feature Request

Whether the `AdaptiveImage` delivers the original or applies transformations is fixed: svgs and gifs are never transformed, other images are only transformed if there is a transformation applied (e.g. rotation, cropping, flipping). As the transformation also includes compression for jpg images which is defined system-wide, we have editors cropping images by one pixel just to make sure the compression is applied - and frequently it is overlooked, causing large image downloads impacting green web policies and site performance (CoreWebVitals affected). 

It would be great to allow configuring the transformation more closely via OSGi settings: defining which image types are always transformed and which are never transformed. This would allow preventing transformation in other formats (e.g. optimized webp rendition already available, should be delivered as is). Additionally, if one sets up compression for jpegs, one could force the transformation process (incl. compression) of all jpegs even if no manual transformation is defined. 

An alternative would be to include compression in the delivery of any jpg, but that would leave less flexibility to individual projects and would impact existing solutions.

Stefan-Franck:
| Fixed Issues?            | Fixes #2530

Ensures users can configure which image types are never transformed (currently: svg and gif) and which image types are always transformed (currently: none) via OSGi. 
By this, transformation can be easily disabled for other image types, but it is also possible to require the transformation for jpg/jpeg, thus ensuring custom compression settings are used - currently, editors in AEM have to crop jpg images by a single pixel just to deliver them in a compressed way - otherwise, web performance and green web goals are not met. 
The default configurations for these settings ensure backward compatibility, as they ensure the system behaves as before.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -118,6 +118,8 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
     private static final String SELECTOR_WIDTH_KEY = "width";
     private int defaultResizeWidth;
     private int maxInputWidth;
+    private List<String> nonTransformableImageTypes;
+    private List<String> forcedTransformationImageTypes;
 
     private AdaptiveImageServletMetrics metrics;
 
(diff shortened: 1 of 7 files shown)


Stefan-Franck:
closing due to lack of CLA

PR #2534, opened 2023-06-23
[2530] AdaptiveImage: control skipping/requiring transformation via OSGi config

### LINKED ISSUE #2530: AdaptiveImage: Configurability of when to apply transformation by file type

Stefan-Franck:
| Fixed Issues?            | Fixes #2530

Ensures users can configure which image types are never transformed (currently: svg and gif) and which image types are always transformed (currently: none) via OSGi. 
By this, transformation can be easily disabled for other image types, but it is also possible to require the transformation for jpg/jpeg, thus ensuring custom compression settings are used - currently, editors in AEM have to crop jpg images by a single pixel just to deliver them in a compressed way - otherwise, web performance and green web goals are not met. 
The default configurations for these settings ensure backward compatibility, as they ensure the system behaves as before.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -118,6 +118,8 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
     private static final String SELECTOR_WIDTH_KEY = "width";
     private int defaultResizeWidth;
     private int maxInputWidth;
+    private List<String> nonTransformableImageTypes;
+    private List<String> forcedTransformationImageTypes;
 
     private AdaptiveImageServletMetrics metrics;
 
@@ -128,12 +130,15 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
     private transient AssetStore assetStore;
 
     public AdaptiveImageServlet(MimeTypeService mimeTypeService, AssetStore assetStore, AdaptiveImageServletMetrics metrics,
-            int defaultResizeWidth, int maxInputWidth) {
+            int defaultResizeWidth, int maxInputWidth, List<String> nonTransformableImageTypes,
+            List<String> forcedTransformationImageTypes) {
         this.mimeTypeService = mimeTypeService;
         this.assetStore = assetStore;
         this.metrics = metrics;
         this.defaultResizeWidth = defaultResizeWidth > 0 ? defaultResizeWidth : DEFAULT_RESIZE_WIDTH;
         this.maxInputWidth = maxInputWidth > 0 ? maxInputWidth : DEFAULT_MAX_SIZE;
+        this.nonTransformableImageTypes = nonTransformableImageTypes;
+        this.forcedTransformationImageTypes = forcedTransformationImageTypes;
     }
 
     @Override
@@ -265,8 +270,8 @@ public class AdaptiveImageServlet extends SlingSafeMethodsServlet {
     protected void transformAndStreamAsset(SlingHttpServletResponse response, ValueMap componentProperties, int resizeWidth, double quality,
                                          Asset asset, String imageType, String imageName) throws IOException {
         String extension = mimeTypeService.getExtension(imageType);
-        if ("gif".equalsIgnoreCase(extension) || "svg".equalsIgnoreCase(extension)) {
-            LOGGER.debug("GIF or SVG asset detected; will render the original rendition.");
+        if (nonTransformableImageTypes.contains(extension)) {
+            LOGGER.debug("{} asset detected; will render the original rendition.", extension);
             metrics.markOriginalRenditionUsed();
             try (InputStream is = asset.getOriginal().getStream()) {
                 if (is != null) {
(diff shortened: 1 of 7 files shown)


Stefan-Franck:
As discussed with @kwin I updated the change such that gif and svg are always exempt from transformation (as the library cannot handle them). The field is still configurable in order to extend (e.g. adding webp to the files spooled directly).

Stefan-Franck:
Hi @joerghoh and thanks for the fast reply!

I'm completely with you - a perfect match does not require an additional transformation. But in reality, more often than not, images with the highest qualities and sometimes highest resolutions are uploaded to the DAM - and that impacts the delivery. It's also not easy to identify and optimize them if you are working in large installations with thousands of images. What happens is a slow degradation of the page weight, with more and more images adding 1 MB or more to it - thus impacting performance (and CoreWebVital rankings) and also impacting GreenWeb initiatives. 

That's why the solution is implemented as an option - out of the box, it works without additional transformation. Only if you make the deliberate choice to have a different behaviour, you can do so. 

I will add a test case in the course of this week.

Stefan-Franck:
But where does that special web-optimized rendition come from? Sounds to me like a manual process for thousands of assets that could be optimized on delivery. If somebody has such a process in place, they can use it. But if not, enforcing compression for all compressable assets is a much more reliable approach. Furthermore, the compressed assets should be cached for a rather long time, thus the rendering cost of the AdaptiveImageServlet shouldn't happen in most cases.

Stefan-Franck:
@joerghoh added three test cases: 
1) disabling transformation for asset types
2) direct spooling is active for asset without forced transformation
3) with forced transformation, the same asset is delivered compressed.

Stefan-Franck:
@joerghoh - As you can see from the test cases, the full-quality Adobe logo has 250kB, the compressed one only has 130kB, already a reduction by 50 % with the default settings (compression rate around .82). We sometimes use compression factors of .5 - .6, so it really reduces the page weight considerably if the compression is applied or not. 
As you cannot guarantee uploaded images adhere to any compression standards (only image size in general, which depends on many other factors), you will always end up with non-optimized images in the repository. Being able to easily optimize on delivery would help editors a lot - they currently crop every image on every usage by 1px only to make sure the image is compressed, which is very tedious and error-prone. 
The change would ensure the system behaves like before, and only if one client wants the compression activated (or webp images delivered as-is), they can apply the corresponding configurations.

Stefan-Franck:
Also an interesting approach, and then have the image component including that compressed rendition instead of the original one - that would work well. So we could decide the time the load occurs from on-upload (how about bulk uploads when the new design agency drops another 1.000 images) to on-delivery (how about many parallel requests and generally longer TTFBs during delivery).

kwin:
Currently the algorithm in https://github.com/adobe/aem-core-wcm-components/blob/68e429d7b072b1d01545836aa629e5eeaeadffce/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java#L518 doesn't distinguish between original and post-processed renditions, so as long as you cannot skip originals the approach with processing profiles won't work reliably.

joerghoh:
@kwin In the my ideal world you would not need to use the AdaptiveImageServlet, because the assets you are delivering are available already with the required dimensions and quality.

Stefan-Franck:
@joerghoh That would require pre-generation and storage of many assets, but would work if you have a limited set of sizes and are not expecting any changes to the required sizes. For all assets without local changes like cropping or rotation applied, one could look up if there is a rendition with the correct size in the DAM and deliver that if it exists. Setting up the DAM workflow to generate those renditions should work well (not sure about compression, though). 
I could add that as a separate PR sometime if you think that's a good solution?

bobvanmanen:
I am trying to understand the issue. The problem is that the renditions we produce during asset processing are not as efficient as what the `AdaptiveImageServlet` creates assuming the same format, dimension, etc?

Stefan-Franck:
@bobvanmanen - the AdaptiveImageServlet takes - if possible - the original asset from the DAM. As it is spooling the original directly, no compression is applied. This leads to AEM delivering unnecessarily large, uncompressed images via this servlet. At my current customer, editors actually crop all images by 1 pixel just to ensure the compression is applied.
This solution allows defining image types that will always be transformed, regardless of editorial applied transformations like the cropping.
@joerghoh suggested the alternate approach to have the correct renditions in the DAM up front, so they can be delivered directly without transformations. 
This would allow an even faster delivery on request, as the processing load is shifted from the delivery time to the upload time. I'm a bit afraid about larger bulk uploads or design changes (you need a new image width for all assets). It would also take a lot of storage space for AEM. In order to implement this approach, we would need to change the DAM Update Asset Workflow to generate compressed and correctly resized (proportionally, not stretched or padded) images according to the design. Those assets could be spooled directly, if the AdaptiveImageServlet resolves them properly (which does not work well according to @kwin ).

Stefan-Franck:
@joerghoh @bobvanmanen - any thoughts?

PR #2564, opened 2023-08-03
[SITES-15031] [Regression] Encoded URL is decoded unintentionally and…

cristic83:
… the URL doesn't work

| Fixed Issues?            | Fixes SITES-15031
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
@@ -21,13 +21,15 @@ import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.security.SecureRandom;
-import java.util.Collections;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


vladbailescu (commented):
We should also add relevant test cases in https://github.com/adobe/aem-core-wcm-components/blob/main/bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/link/DefaultPathProcessorTest.java

edoardo-goracci (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
```diff
@@ -41,7 +43,21 @@ public class LinkUtil {
...
+    private static final String[] RESERVED_CHARACTERS_ENCODED = {
+        "%20", "%21", "%22", "%23", "%24", "%25", "%26", "%27", "%28","%29",
+        "%2A", "%2B", "%2C", "%2F", "%3A", "%3B", "%3D", "%3F", "%40", "%5B", "%5D",
+        "%2a", "%2b", "%2c", "%2f", "%3a", "%3b", "%3d", "%3f", "%5b", "%5d"
+    };
+    private final static List<Pattern> PATTERNS = new ArrayList<>();
+
+    static {
+        PATTERNS.add(Pattern.compile("(<%[=@].*?%>)"));
+        PATTERNS.addAll(Arrays.stream(RESERVED_CHARACTERS_ENCODED)
```
LSantha, line 56: @cioriia , why do we need a distinct regex pattern for each character?

PR #2593, opened 2023-10-02
Updated org.json dependency

vladbailescu:


bundles/core/pom.xml
```diff
@@ -529,6 +529,11 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.adobe.granite.bundles</groupId>
+            <artifactId>json</artifactId>
```
kwin, line 534: This is really deprecated, rather use JSON-P Impl with implementation Jackson.

vladbailescu: @kwin , unfortunately we cannot get rid completely of `org.json` as it's part of a public API: https://github.com/adobe/aem-core-wcm-components/blob/main/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/services/form/FormHandler.java#L39

alexandru-stancioiu (approved):

Myronenko (approved):

bundles/core/pom.xml
LSantha: How about deprecating this method or interface in favor of anew API?

PR #2604, opened 2023-10-23
SITES-16805: init rewriter processor at the right time with toc enabled

Buuhuu:
| Fixed Issues?            | 

With the TOC Filter enabled the resourceType condition evaluation for the Sling Rewriter Pipeline configuration is not applied on the usual resource anymore: https://sling.apache.org/documentation/bundles/output-rewriting-pipelines-org-apache-sling-rewriter.html#configuring-a-processor

Usually (without the TOC Filter) the following happens:

```
/content/path/to/page (cq:Page)
 - jcr:content (sling:resourceType=myapp/components/page)
```

The `/libs/cq/Page/Page.jsp` includes the `jcr:content` child transparently without writing anything to the response, hence not calling `HttpServletResponse#getWriter()`. The `myapp/components/pagE` or any of its super types contains usually the first script that writes to the response, hence calls `HttpServletResponse#getWriter()` which initialises the Sling Rewriter processing as a side-effect with the `jcr:content` resource and it in particular its `resourceType` as context: https://github.com/apache/sling-org-apache-sling-rewriter/blob/master/src/main/java/org/apache/sling/rewriter/impl/RewriterResponse.java#L89-L119

With the TOC Filter enabled, the response is wrapped and its writer replaced with some in-memory buffering writer. The content of the buffer is than processed and written to the original writer after calling `HttpServletResponse#getWriter()`. That works functionally, however, as `HttpServletResponse#getWriter()` is called after the page was rendered the response's upstream `getWriter()` is only called now, initialising the Sling Rewriter processor in the context of the cq:Page instead of the `jcr:content`.

To see this in action, check the /system/console/request for the same page with the filter enabled/disabled and look for `LOG Found processor for post processing ProcessorConfiguration`, it should appear right after the include of the jcr:content resource's main rendering script. With the TOC Filter enabled, it appears at the end of the whole rendering.

With this PR the upstream `getWriter()` is called at the time the `CharResponseWrapper#getWriter()` is invoked. This triggers the side-effect to initialise the Sling Rewriter processor at the right time.


vladbailescu (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/CharResponseWrapper.java
```diff
@@ -37,14 +39,21 @@ public class CharResponseWrapper extends HttpServletResponseWrapper {
...
      * Returns the newly created {@link PrintWriter} object created by this wrapper
      * @return - newly created {@link PrintWriter}
      */
     @Override
-    public PrintWriter getWriter() {
+    public PrintWriter getWriter() throws IOException {
+        // call getWriter() on the original response, as this is used to initialise the Sling Rewriter Pipeline
+        // if not called this way, the TableOfContentsFilter would call getWriter() on the upstream response first in the context of a
+        // cq:Page resource, and hence the resourceType condition to select the processor configuration does not work
+        wrapped.getWriter();
```
vladbailescu, line 55: Just a note here, something we also discussed direclty: getters should not produce this kind of side-effects. But it is what it is.

Buuhuu: yes unfortunately.

PR #2607, opened 2023-10-31
Fix: [Image + Teaser] Don't require ALT text if no asset selected #2523

### LINKED ISSUE #2523: [Image + Teaser] Don't require ALT text if no asset selected
## Bug Report

**Current Behavior**
When editing an image component, if no image is provided, or the image component inherits the page image and no such image is configured, the dialog still requires an alt text to be specified. This encourages users to either add a blank/garbage alt text to allow the dialog to save, or to set the image as decorative - which may not be appropriate. 

This may seem trivial in image components, because there is no good reason to add an image component without an image; however, the image dialog is also used in other components - such as the Teaser component - where an image is often not provided, which is an entirely valid use case for a teaser. 

In this situation, the dialog is difficult to work with because the user should not have to play with alt text settings in a teaser with no image.

**Expected behavior/code**
The alt text should only be a required field if an image asset is configured.

**Environment**
- AEM 6.5.17.0
- Core Components version 2.23.0
- JRE version 11

ConnorJamesLow:
| Fixed Issues?            | Fixes #2523 
| Patch: Bug Fix?          | ✅ 
| Minor: New Feature?      | 
| Major: Breaking Change?  | 
| Documentation Provided   | ✅ Yes (code comments)

Fix the image dialog validation logic per #2523. 
- Checking "Don't provide an alternative text" should prevent an error message from showing under alt when done editing.
- Checking "Inherit from description of asset" should also prevent an error message.

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/clientlibs/editor/js/image.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/clientlibs/editor/js/image.js
@@ -13,7 +13,7 @@
  ~ See the License for the specific language governing permissions and
  ~ limitations under the License.
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
-(function($, Granite) {
+ (function($, Granite) {
     "use strict";
 
     var dialogContentSelector = ".cmp-image__editor";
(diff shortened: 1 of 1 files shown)


content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/clientlibs/editor/js/image.js
```diff
@@ -170,8 +171,15 @@
...
+            var isDecorative = $(altIsDecorativeSelector).attr('checked');
             var assetWithoutDescriptionErrorMessage = "Error: Please provide an asset which has a description that can be used as alt text.";
-            if (isAltCheckboxChecked && !seededValue) {
+
+            // Validate the following:
+            // - the image is not decorative
+            // - alt is not inherited from the DAM
+            // - the alt value is empty.
+            // If all of these are true, return a validation error
+            if (!isDecorative && !isAltCheckboxChecked && !seededValue) {
```
ConnorJamesLow, line 182: Another check that could be beneficial:
```js
// Do we have an asset at all?
var isImageSelected = !!$('[name="./fileReference"]').val(); 

if (isImageSelected && !isDecorative && !seededValue)
```

PR #2613, opened 2023-11-06
Add support for Google data layer

brobatr:
| Fixed Issues?            | None
| Patch: Bug Fix?          | None
| Any Dependency Changes?  | Yes, data layer read me

The PR adds code to:

Make the name of the data layer configurable
Option to exclude ACDL but keep the other logic
Option to push to a datalayer using the gtag() function instead of push()
Modified unit tests

Please see the updated documentation in DATA_LAYER_INTEGRATION.md

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/DataLayerConfig.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/DataLayerConfig.java
@@ -24,6 +24,8 @@ import org.apache.sling.caconfig.annotation.Property;
 @Configuration(label="Data Layer", description="Configure support for Adobe Client Data Layer")
 public @interface DataLayerConfig {
 
+    String DATALAYER_OBJECT_NAME_ADOBE = "adobeDataLayer";
+
     /**
      *
      * @return {@code true} if the data layer is enabled, {@code false} otherwise. It defaults to {@code false}.
@@ -38,5 +40,30 @@ public @interface DataLayerConfig {
      */
     @Property(label="Data Layer client library not included")
     boolean skipClientlibInclude() default false;
+
+    /**
+     *
+     * @return the name of the datalayer
+     * Defaults to adobeDataLayer.
+     */
+    @Property(label = "Data Layer object name")
+    String name() default DATALAYER_OBJECT_NAME_ADOBE;
+
+    /**
+     *
+     * @return the type of push function (push() or gtag()) of the datalayer
+     * Defaults to push.
+     */
+    @Property(label = "Use gtag function to push data")
+    boolean pushFunctionUseGtag() default false;
+
+    /**
+     *
+     * @return ACDL library not included
+     * Defaults to false.
+     */
+    @Property(label = "ACDL library not included")
+    boolean skipAcdlInclude() default false;
+
 
 }
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
@@ -38,6 +38,8 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 @ConsumerType
 public interface Page extends ContainerExporter, Component {
 
+    String DATALAYER_OBJECT_NAME_ADOBE = "adobeDataLayer";
+
     /**
      * Key used for the regular favicon file.
      *
@@ -467,4 +469,27 @@ public interface Page extends ContainerExporter, Component {
      */
     default boolean isDataLayerClientlibIncluded() {return true;}
 
+    /**
+     * Returns the name of the data layer.
+     *
+     * {@code true} if the data layer client library should be included.
+     * @since om.adobe.cq.wcm.core.components.models 12.24.0
+     */
+    default String getDataLayerName() { return DATALAYER_OBJECT_NAME_ADOBE; }
+
+    /**
+     * Checks if gtag() should be used to push to the data layer.
+     *
+     * {@code true} if gtag() should be used.
+     * @since om.adobe.cq.wcm.core.components.models 12.24.0
+     */
+    default boolean isDataLayerUseGtag() { return false; };
+
+    /**
+     * Checks if the ACDL library should be included.
+     *
+     * {@code true} if ACDL library should be used.
+     * @since om.adobe.cq.wcm.core.components.models 12.24.0
+     */
+    default boolean isDataLayerSkipAcdlInclude() { return false; };
 }
--- a/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  ******************************************************************************/
+/* global gtag */
 (function() {
     "use strict";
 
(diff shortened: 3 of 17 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -467,4 +469,27 @@ default List<String> getRobotsTags() {
      */
     default boolean isDataLayerClientlibIncluded() {return true;}
 
+    /**
+     * Returns the name of the data layer.
+     *
+     * {@code true} if the data layer client library should be included.
+     * @since om.adobe.cq.wcm.core.components.models 12.24.0
+     */
+    default String getDataLayerName() { return DATALAYER_OBJECT_NAME_ADOBE; }
```
vladbailescu, line 478: This will require a package minor version bump

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
```diff
@@ -467,4 +469,27 @@ default List<String> getRobotsTags() {
...
+     */
+    default boolean isDataLayerUseGtag() { return false; };
+
+    /**
+     * Checks if the ACDL library should be included.
+     *
+     * {@code true} if ACDL library should be used.
+     * @since om.adobe.cq.wcm.core.components.models 12.24.0
+     */
+    default boolean isDataLayerSkipAcdlInclude() { return false; };
```
vladbailescu, line 494: We already have `isDataLayerClientlibIncluded` that serves the same purpose (to control if the ACDL scripts should be included).

content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/site/js/accordion.js
```diff
@@ -458,12 +478,12 @@
         }
 
         /**
-         * Annotates the item and its internals with
-         * the necessary style and accessibility attributes to indicate it is expanded.
-         *
-         * @private
-         * @param {HTMLElement} item The item to annotate as expanded
-         */
+             * Annotates the item and its internals with
```
vladbailescu, line 481: I see no reason to reformat these

content/src/content/jcr_root/apps/core/wcm/components/page/v3/page/headlibs.html
```diff
@@ -26,5 +26,12 @@
     <sly data-sly-test="${clientLibCategoriesJsHead}" data-sly-call="${clientlib.js @ categories=clientLibCategoriesJsHead, async=page.isClientlibsAsync}"></sly>
     <sly data-sly-test="${clientLibCategories}" data-sly-call="${clientlib.css @ categories=clientLibCategories}"></sly>
     <link data-sly-test="${staticDesignPath}" href="${staticDesignPath}" rel="stylesheet" type="text/css" />
-    <sly data-sly-test="${page.data && page.dataLayerClientlibIncluded}" data-sly-call="${clientlib.js @ categories='core.wcm.components.commons.datalayer.v1', async=true}"></sly>
+    <sly data-sly-test="${!page.dataLayerSkipAcdlInclude}">
```
vladbailescu, line 29: We already have `isDataLayerClientlibIncluded` that serves the same purpose (to control if the ACDL scripts should be included).

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/DataLayerConfig.java
```diff
@@ -39,4 +41,29 @@
...
+     */
+    @Property(label = "Data Layer object name")
+    String name() default DATALAYER_OBJECT_NAME_ADOBE;
+
+    /**
+     *
+     * @return the type of push function (push() or gtag()) of the datalayer
+     * Defaults to push.
+     */
+    @Property(label = "Use gtag function to push data")
```
vladbailescu, line 57: Wouldn't it be enough to know that the name of the datalayer is gtag?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/DataLayerConfig.java
```diff
@@ -39,4 +41,29 @@
...
+    boolean pushFunctionUseGtag() default false;
+
+    /**
+     *
+     * @return ACDL library not included
+     * Defaults to false.
+     */
+    @Property(label = "ACDL library not included")
+    boolean skipAcdlInclude() default false;
```
vladbailescu, line 66: We already have `skipClientlibInclude` that serves the same purpose (to control if the ACDL scripts should be included).

vladbailescu (changes requested):
Build is failing due to baseline check.
Also, to keep it simple, I would only add one extra property in the config/model to differentiate between ACDL and GTAG.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/models/Page.java
brobatr: hi @vladbailescu  

unfortunately isDataLayerClienlibIncluded removes the datalayer.js file not just the ACDL.  This may be a bug: when the flag is set to true there is an odd state - some component datalayer interaction happens as before (accordian, text etc), but the components that are caught in a loop in datalayer.js are no longer pushed to the datalayer.  In short we get a semi-enabled state.

This is the reason: 

assets: {
                js: [
                    "src/scripts/datalayer/v1/polyfill.js",
                    "node_modules/@adobe/adobe-client-data-layer/dist/adobe-client-data-layer.min.js",
                    "src/scripts/datalayer/v1/datalayer.js"
                ]
            }

 For the GDL I need to keep datalayer.js.  This has led to the ugly approach of "take DL away and put part of it back again"

Perhaps it would be better to change the existing functionality so that only the ACDL code is removed by isDataLayerClientlibIncluded?  If so I can include that here and it would allow me to make the whole thing cleaner.

content/src/content/jcr_root/apps/core/wcm/components/page/v3/page/headlibs.html
brobatr: see Page.java explanation above

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/DataLayerConfig.java
brobatr: no, the datalayer can be called anything or normally dataLayer, regardless of whether GA (gtag) or GTM is used.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/DataLayerConfig.java
brobatr: see reply to Paga.java above

brobatr:
> Build is failing due to baseline check. Also, to keep it simple, I would only add one extra property in the config/model to differentiate between ACDL and GTAG.

This is rooted in the same issue as described in Page.java

there are four states now, assuming data layer is enabled:

ACDL + datalayer.js included - skipClientlibInclude = true (isDataLayerClientlibIncluded = false)
(probably a bug) ACDL + **datalayer.js** not included - skipClientlibInclude = false  (isDataLayerClientlibIncluded = true)
datalayer.js included, push using push() (GTM) - skipClientlibInclude = false, skipAcdlInclude() = true, usegtag=false
datalayer.js  included, push using gtag() (GA) - skipClientlibInclude = false, skipAcdlInclude() = true, usegtag=true

what would be good, if the second state is actually a bug is:

enabled: datalayer.js included

ACDL included - skipClientlibInclude = false
ACDL not included - skipClientlibInclude = true
Use gtag - usegtag = true
Use push - usegtag = false

brobatr:
blocked by https://github.com/adobe/aem-core-wcm-components/issues/2620

brobatr:
will create a new PR after #2620 is fixed

Issue #2606, opened 2023-10-30
[Client Library] Non-deterministic behavior of `HashSet` might fail the test `ClientLibrariesImplTest.testGetCategoriesWithInjectedResourceTypesAndInheritance` 


SaaiVenkat:
## Bug Report

**Current Behavior**
- `ClientLibrariesImpl` loads the `categoriesArray` after instantiation at line no: 157 https://github.com/adobe/aem-core-wcm-components/blob/6a102d1d5879d142123a547c00e2d226e475a636/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java#L149-L157
- This value is set from `Set<String>` returned by `getCategoriesFromComponent()`, if the result from `Utils.getStrings(categories)` and `Utils.getStrings(raCategories)` are empty.
- However, looking at the implementation of `getCategoriesFromComponent()`, `HashSet` is used to store the filtered categories https://github.com/adobe/aem-core-wcm-components/blob/6a102d1d5879d142123a547c00e2d226e475a636/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java#L298-L313
- The [javadoc of HashSet](https://docs.oracle.com/javase/8/docs/api/java/util/HashSet.html) mentions that order of a HashSet will not be maintained over time. But the above implementation of `categoriesArray` assumes that the `categories` elements return from `getCategoriesFromComponents()` to be returned in a specific order. This assumption might result cause issues with Client Libraries usage, when the environment or java version is changed.

#### Steps to Reproduce 
- The above issue could be observed with the failure of `com.adobe.cq.wcm.core.components.internal.models.v1.ClientLibrariesImplTest.testGetCategoriesWithInjectedResourceTypesAndInheritance`, when run through [NonDex](https://github.com/TestingResearchIllinois/NonDex), which shuffles the order of elements in a `HashSet`.
```
# Clone the Repo
https://github.com/adobe/aem-core-wcm-components.git

# Compile the module byte-buddy-dep
mvn clean install -pl bundles/core -am -DskipTests

# (Optional) Run the unit test
mvn -pl bundles/core test -Dtest=com.adobe.cq.wcm.core.components.internal.models.v1.ClientLibrariesImplTest#testGetCategoriesWithInjectedResourceTypesAndInheritance

# Run the unit test using NonDex
[...]

PR #2621, opened 2023-11-17
[2606] Fixed Non-deterministic behavior of HashSet that might fail the test ClientLibrariesImplTest

SaaiVenkat:
| Fixed Issues?            | Fixes [`#2606`](https://github.com/adobe/aem-core-wcm-components/issues/2606)
| Tests Added + Pass?      | Tests Pass

- Since the non-deterministic order of `HashSet` causes the issue, we need to make it deterministic.
- This is done by changing the implementation of `categories` from `HashSet` to `LinkedHashSet`. 
- This ensures that the order of elements is maintained in their order of insertion similar to
https://github.com/adobe/aem-core-wcm-components/blob/0efca91f4b7b5f56a02fefabbdec273c8b77ffd8/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java#L324

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ClientLibrariesImpl.java
@@ -22,7 +22,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


SaaiVenkat:
@vladbailescu Since you approved these changes [here](https://github.com/SaaiVenkat/aem-core-wcm-components/pull/2), I am raising this PR.

vladbailescu (approved):

PR #2622, opened 2023-11-17
[2620] fix datalayer skipClientlibInclude logic

### LINKED ISSUE #2620: [DataLayer] Removing ACDL also removes datalayer content
## Bug Report

**Current Behavior**
If the option to remove the ACDL library is used by setting skipClientlibInclude to true, the functionalily in datalayer.js is also removed.  This means that information about the components on a page is no longer pushed to the datalayer.

**Expected behavior/code**
skipClientlibInclude should only remove the ACDL library and not affect the datalayer content.

**Environment**
- AEMaaCS 2023.11.14227.20231108T162349Z-23100
- JRE version 11.0.14+8-LTS-263

**Possible Solution**
Split out the ACDL from the clientlib containing datalayer.js so that only the ACDL is removed when skipClientlibInclude is set to true.  datalayer.js should then only be removed if the datalayer is disabled

datalayer_bug

brobatr:
| Fixed Issues?            |  Fixes #2620
| Patch: Bug Fix?          |
| Documentation Provided   | N

The fix separates the ACDL library from the file datalayer.js and thereby allows the exclusion of the ACDL while retaining the datalayer.js logic (assuming datalayer is enabled).  

These were previously handled as a single clientlib, core.wcm.components.commons.datalayer.v1 and this could be removed by setting the property skipClientlibInclude in the DataLayerConfig CAC to true.  

With this fix the skipClientlibInclude property now only removes the ACDL, leaving a plain array datalayer with the content provided by datalayer.js.

Please see screenshots in issue #2620 to see the issue

### DIFF
--- a/content/clientlib.config.js
+++ b/content/clientlib.config.js
@@ -28,8 +28,18 @@ module.exports = {
             assets: {
                 js: [
                     "src/scripts/datalayer/v1/polyfill.js",
-                    "node_modules/@adobe/adobe-client-data-layer/dist/adobe-client-data-layer.min.js",
                     "src/scripts/datalayer/v1/datalayer.js"
+                ]
+            }
+        },
+        {
+            name: "core.wcm.components.commons.datalayer.acdl",
+            serializationFormat: "xml",
+            allowProxy: true,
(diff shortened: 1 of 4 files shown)


brobatr:
mistake in code - will fix and recreate

PR #2624, opened 2023-11-19
[2620] fix datalayer ACDL inclusion logic issue

### LINKED ISSUE #2620: [DataLayer] Removing ACDL also removes datalayer content

brobatr:
Contextualised components.

| Fixed Issues?            |  Fixes #2620
| Patch: Bug Fix?          |
| Documentation Provided   | N

The fix separates the ACDL library from the file datalayer.js and thereby allows the exclusion of the ACDL while retaining the datalayer.js logic (assuming datalayer is enabled).  

These were previously handled as a single clientlib, core.wcm.components.commons.datalayer.v1 and this could be removed by setting the property skipClientlibInclude in the DataLayerConfig CAC to true.  

With this fix the skipClientlibInclude property now only removes the ACDL, leaving a plain array datalayer with the content provided by datalayer.js.

Please see screenshots in issue #2620 to see the issue

### DIFF
--- a/content/clientlib.config.js
+++ b/content/clientlib.config.js
@@ -18,10 +18,24 @@
  */
 module.exports = {
     context: __dirname,
-    clientLibRoot: "src/content/jcr_root/apps/core/wcm/components/commons/datalayer/v1/clientlibs",
     libs: [
         {
             name: "core.wcm.components.commons.datalayer.v1",
+            path: "src/content/jcr_root/apps/core/wcm/components/commons/datalayer/v1/clientlibs",
+            serializationFormat: "xml",
+            allowProxy: true,
+            jsProcessor: ["default:none", "min:gcc;compilationLevel=whitespace"],
+            assets: {
+                js: [
+                    "src/scripts/datalayer/v1/polyfill.js",
+                    "node_modules/@adobe/adobe-client-data-layer/dist/adobe-client-data-layer.min.js",
+                    "src/scripts/datalayer/v1/datalayer.js"
+                ]
+            }
+        },
+        {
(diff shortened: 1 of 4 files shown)


content/clientlib.config.js
```diff
@@ -28,10 +28,20 @@ module.exports = {
             assets: {
                 js: [
                     "src/scripts/datalayer/v1/polyfill.js",
-                    "node_modules/@adobe/adobe-client-data-layer/dist/adobe-client-data-layer.min.js",
```
vladbailescu, line 31: This would be a breaking change for any customer that relies on `core.wcm.components.commons.datalayer.v1` to bring in the ACDL module.

I think we should instead define a `core.wcm.components.commons.datalayer.v1` clientlib which is lightweight and an `core.wcm.components.commons.datalayer.acdl` clientlib to bring the ACDL module in.

cbogdan0707 (approved):

vladbailescu (approved):

LSantha (approved):

PR #2698, opened 2024-03-17
ASSETS-33747 named smartcrop in image core v3 component powered by fastly

mohiaror:
* ASSETS-33747 named smartcrop in image core v3 component powered by fastly

Instead of fastly smartcrops, expect named smartcrops in delivery URLs.

* ASSETS-33747 named smartcrop in image core v3 component powered by fastly

Add editor code to render image through delivery URL and support smartcrop.

* @trivial fix linting errors.

* Show smartcrop renditions group even if DM is disabled but polaris is enabled.

* Update code to also show dynamicmedia group when dialog is loaded with remote asset.

* Make dependency on clientbuilderfactory optional and check for null. This is to avoid rewriting/fixing all Image test cases.

* ASSETS-33747 named smartcrop in image core v3 component powered by fastly

Bug fixes.

* Minor bug fixes.

* Fix test cases.

* @trivial Fix existing test cases.

* ASSETS-33747 named smartcrop in image core v3 component

Review comments. Add a separate test case for smartcrop. Add a failing testcase for auto smartcrop.

* ASSETS-33747 named smartcrop in image core v3 component

Incorporate review feedback.

* @trivial Hide the smartcrop option if remote asset does not have smartcrops.

* ASSETS-33747 named smartcrop in image core v3 component

Add test cases for srcset and responsehandler code.

* @trivial Move the code to show smartcrop dropdown only after fetching the metadata.

* @trivial Fix eslint error.

---------

| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/_cq_design_dialog/.content.xml
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/_cq_design_dialog/.content.xml
@@ -51,9 +51,6 @@
                                         text="Enable DM features"
                                         uncheckedValue="false"
                                         value="{Boolean}true">
-                                        <granite:rendercondition jcr:primaryType="nt:unstructured"
-                                                      sling:resourceType="granite/ui/components/coral/foundation/renderconditions/feature"
-                                                      feature="com.adobe.dam.asset.scene7.feature.flag"/>
                                     </enableDmFeatures>
                                     <enableAssetDelivery
                                         jcr:primaryType="nt:unstructured"
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
@@ -16,13 +16,25 @@
 package com.adobe.cq.wcm.core.components.internal.models.v3;
 
 import java.awt.*;
+import java.io.IOException;
+import java.io.StringReader;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Scanner;
 
 import javax.annotation.PostConstruct;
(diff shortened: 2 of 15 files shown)


content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/_cq_design_dialog/.content.xml
```diff
@@ -51,9 +51,6 @@
                                         text="Enable DM features"
                                         uncheckedValue="false"
                                         value="{Boolean}true">
-                                        <granite:rendercondition jcr:primaryType="nt:unstructured"
```
LSantha, line 54: I'm not sure if this is the way to go. This means that DM features would be always available even if no scene7 or NGDM is available. Is that what you want?

mohiaror: @LSantha Is there a rendercondition to check for OSGI configuration's value in core components? Or any other way apart from renderconditions?

FWIW, removal of this rendercondition will only make the enable DM features checkbox visible in the design dialog of image component. It will not enable it by default. An author will have to update the template to enable the feature manually.

LSantha: @mohiaror , I don't think there is such render condition but you can create one.

I know this thing is about the design dialog but as you can see it was still provided with a render condition to avoid any confusion for customer not using this feature.

mohiaror: @LSantha I have added an OR condition for both DM and NGDM. Please see if it looks Ok to you.

LSantha (approved):

PR #2705, opened 2024-03-28
[Carousel] Remove attribut duplication on "previous" button

renow-luxembourg:
Bug Report

**Current Behavior :** 
In the carousel component, we can define a label and an aria-label attribute on carousel control buttons.
It work well on the "Next" button.
On the "Previous" button, we can change only the label, its aria-label attribut isn't customizable. 
This causes restitution differences for people using assistive tool (like the NVDA screen reader).

**Expected behavior/code :** 
Keep the same behavior for both "Next" and "Previous" control button.

**Environment :**
- AEM 6.5.19

**Possible Solution :**
Removing the aria-label attribut duplication solve the issue.


LSantha (approved):

renow-luxembourg:
Adobe CLA filled and signed.

PR #2717, opened 2024-04-08
[Core Components] OEmbedClientImpl leaks network connections

vladbailescu:
* Refactored code to move HTTP client creation to a dedicated method and wrap the whole method that does the reading in try-with-resources


Myronenko (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/embed/OEmbedClientImpl.java
```diff
@@ -156,24 +152,18 @@ protected OEmbedClientImplConfigurationFactory.Config getConfiguration(String ur
...
-                    .setDefaultRequestConfig(rc)
-                    .build()){
-                response = httpClient.execute(new HttpGet(url));
-            }
+            return HttpClients.custom().setDefaultRequestConfig(rc).build();
         }
+    }
+
+    protected InputStream getData(String url, HttpClient httpClient) throws IOException, IllegalArgumentException {
+        HttpResponse response = httpClient.execute(new HttpGet(url));
```
kwin, line 166: I really recommend implementing https://www.javadoc.io/doc/org.apache.httpcomponents/httpclient/4.3.2/org/apache/http/client/ResponseHandler.html instead and pass that to execute. Otherwise you need to manually make sure to consume the inputstream fully (e.g. with https://hc.apache.org/httpcomponents-core-4.4.x/current/httpcore/apidocs/org/apache/http/util/EntityUtils.html#consumeQuietly(org.apache.http.HttpEntity)).

vladbailescu: The JSON/XML un-marshalling should take care of consuming the input stream. Seems a bit ineffective to consume it to a string/buffer (could be large) and then feed it through the un-marshalers

kwin: The idea is rather to call the unmarshaller inside your `ResponseHandler`!

vladbailescu: @kwin , the `ResponseHandler`-based implementation would look something like: #2718

I can't say I particularly like it or find it better, the `HttpResponse` is not closable and we are not reusing the client any way between calls.

PR #2718, opened 2024-04-08
[Core Components] OEmbedClientImpl leaks network connections

vladbailescu:
* Refactored code to add a ResponseHandler


vladbailescu:
TBH, I think this approach over-complicates things. Will try to stick to #2717

PR #2723, opened 2024-04-12
SITES-19780 - Embed component should have more aggressive timeouts

andreeadracea:
Embed component should have more aggressive timeouts

| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

Add the possibility to set the SocketTimeout and ConnectionTimeout values in the OSGI configuration available for each provider. 
Changed the usage of maven-antrun-plugin maven plugin in order to use <target> instead of <tasks>


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/embed/OEmbedClientImplConfigurationFactory.java
```diff
@@ -77,6 +77,18 @@ public class OEmbedClientImplConfigurationFactory {
             description = "Describes whether the provider response HTML is allowed to be displayed in an unsafe context."
         )
         boolean unsafeContext() default false;
+
+        @AttributeDefinition(
+            name = "Socket Timeout",
+            description = "The time waiting for data – after establishing the connection; maximum time of inactivity between two data packets."
+        )
+        int socketTimeout() default 1000;
```
LSantha, line 85: Should we use these aggressive timeouts probably proposed by the customer as default?

For me the safer way would be to preserve the old defaults for both timeouts and let this customer configure the more aggressive timeouts for their solution. Such dramatic change of the defaults (60x and 5x reduction) may cause surprises for other customers who weren't complaining about this.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/embed/OEmbedClientImplConfigurationFactory.java
```diff
@@ -77,6 +77,18 @@ public class OEmbedClientImplConfigurationFactory {
...
+
+        @AttributeDefinition(
+            name = "Connection Timeout",
+            description = "The time to establish the connection with the remote host."
+        )
+        int connectionTimeout() default 1000;
```
LSantha, line 91: Please define constant for the timeouts instead magic values across the code.

content/pom.xml
```diff
@@ -272,7 +272,7 @@
                             <execution>
                                 <phase>prepare-package</phase>
                                 <configuration>
-                                    <tasks>
+                                    <target>
```
LSantha, line 275: Is this some accidental modification?

andreeadracea: When trying to install core-components using the cloud profile for sdk instance, I got the error that the maven-antrun-plugin changed and now is using target instead of tasks. This error should appear for everyone when installing the core-components with this command: mvn clean install -PautoInstallSinglePackage,cloud
https://maven.apache.org/plugins/maven-antrun-plugin/

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/embed/OEmbedClientImplConfigurationFactory.java
andreeadracea: In my opinion we should use an aggressive timeout by default that can be changed by the customers based on the providers they use. In this case the customers that are using our default providers don't have to wait unnecessary time for connection. We have defined config for providers that are offering a fast connection. 
As a middle ground, could we set a less aggressive timeout such 3000ms that should cover both opinions?

content/pom.xml
Myronenko: Fixed/reverted with https://github.com/adobe/aem-core-wcm-components/pull/2723/commits/b152ea1e435b32b97f75d3d3279cfe6e683a5b76

Explicitly specifying the plugin version avoids the issue in the newer/latest versions

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/embed/OEmbedClientImplConfigurationFactory.java
LSantha: @andreeadracea , I've looked into the consequences of using too aggressive timeouts. The way the apache HTTP connection is used implies if a network error occurs (no route to hots, connection timeout, socket timeout etc) then the edit dialog of embed component cannot validate the url and during page rendering for endusers a page with embed component cannot render the embedded player. There is not any kind of retry in place, the only option is for the user to reload the page or the author try to enter the url again in the edit dialog. Without retry a very aggressive timeout is not recommended but the connection between the AEM instance and the embedded service must be fast so, I also think we can reduce the default timeouts. My suggestion is to use 2000 ms for connection timeout and 5000 ms for socket timeout.

@vladbailescu , @joerghoh what do you think?

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/services/embed/OEmbedClientImplConfigurationFactory.java
```diff
@@ -39,6 +39,9 @@ public class OEmbedClientImplConfigurationFactory {
             description = "Configuration for defining oEmbed endpoints."
     )
     public @interface Config {
+
+        int defaultConnectionTimeout = 2000, defaultSocketTimeout=5000;
```
LSantha, line 43: These are constants by default, so the NAMING_CONVENTION_FOR_CONSTANTS ( :-) ) still applies.

LSantha (approved):

RaduADumitru (approved):

PR #2657, opened 2024-01-15
Support Google Tag Manager Data Layer

brobatr:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | N
| Patch: Bug Fix?          | N
| Minor: New Feature?      | Y
| Major: Breaking Change?  | N
| Tests Added + Pass?      | Yes
| Documentation Provided   | Yes (code comments and or markdown)
| Any Dependency Changes?  |
| License                  | Apache License, Version 2.0

See discussion #2565 

The PR allows renaming of the Data Layer.  Together with the bugfix #2620 (previously merged) this allows a generic data layer to be configured, i.e. remove ACDL, rename, without making a breaking change.  If GTM code is on the page and the Data Layer name is changed to the Google default name of dataLayer, the array will be consumed and extended by GTM thus creating a Google Event-Driven Data Layer.  The GDL push() method is available and very similar to that of the ACDL, so all related logic remains the same as before.

Note that support for GTag, the Google global tag, has not been included despite some earlier discussion. This would have been an experimental feature and a significantly more complex implementation.  This could be a future enhancement.

The screenshot below shows the Core Component Data Layer used be Google Tag Manager, with a Google event ID and normal OOTB GTM gtm.start timing event added as normal by GTM.

core-components-google

PR #2724, opened 2024-04-12
SITES-21233 - Improve datalayer initialization

vladbailescu:
* Accounts for cases where the proxy components override HTML and do not carry over the existing initialization logic.


vladbailescu:
#2657 replaced the fallback of initialization to an empty array

PR #2798, opened 2024-06-26
Update accordion.js initizialization

pat-lego:
adobe datalayer can an error when getting data from DOM if not properly set.


RaduADumitru:
Closing because, as per discussion, problem was already fixed by previous change.

PR #2871, opened 2024-09-29
#2853 The AdaptiveImageServlet should send redirects to blobs 

### LINKED ISSUE #2853: AdaptiveImageServlet should not stream binaries
## Feature Request

When possible the AdaptiveImageServlet should not stream binaries through the JVM; there are a number of cases where the asset binary is read from the blobstore and then streamed as-is to the requester. IF the Blobstore supports the download of the binary directly (see [Direct Binary Access](https://jackrabbit.apache.org/oak/docs/features/direct-binary-access.html)) this feature should be used instead.

As in AEM CS the streaming of binaries through the JVM is an antipattern, we should implement this feature here as well. It will be transparent for endusers, as this feature requests adapts the behaviour of the AdaptiveImageServlet to that behavior used by the AEM CS ootb asset download.

Implement the approach described in the above Oak doc.

**Are there alternatives?**
nope

**Documentation**

joerghoh:
| Fixed Issues?            | Fixes #2853 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

The AdaptiveImageServlet always streams its result through the JVM, even if its an already existing rendition with a matching blob in the blobstore. In AEM CS can delivered more efficiently with sending a redirect to the blobstore, and then the Fastly CDN follows this redirect and pulls the data directly from the blobstore. Then AEM does not need to stream that binary through the JVM.

This patch improves the handling not to stream the binary when possible. The feature is configurable and off by default.

(A word to the tests: In JUnit5 it's not possible to parametrize entire test classes anymore, which would have been my preferred way for the tests. Instead you would have to parametrize the tests on their own. To avoid that I duplicated the tests which were failing with the new  feature turned and tagged the duplicates accordingly. Now the ``setup()`` can initialize the AdaptiveImageServlet accordingly. Happy to learn about a way, which allows to run each tests with the new feature turned off and on, without annotating each test.

Existing tests were not changed, to make sure that all the changes did not affect the default behavior.

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/servlets/AdaptiveImageServlet.java
@@ -32,7 +32,10 @@ import java.util.SortedSet;
 import java.util.TreeSet;
 
 import javax.jcr.Binary;
+import javax.jcr.Node;
+import javax.jcr.Property;
 import javax.jcr.RepositoryException;
+import javax.jcr.ValueFormatException;
 import javax.servlet.http.HttpServletResponse;
 
 import org.apache.commons.io.FilenameUtils;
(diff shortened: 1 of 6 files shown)


kwin:
Why not leveraging https://sling.apache.org/documentation/bundles/rendering-content-default-get-servlets.html#streamrendererservlet

joerghoh:
@kwin I want to get rid of that stream approach where possible, not encourage it :-)

kwin:
@joerghoh Please read both documentation and code more thoroughly. That streamrendererservlet does the redirect exactly as you implemented (despite its name): https://github.com/apache/sling-org-apache-sling-servlets-get/blob/507a84c5cc2ede84710dd719f97bb483629b969e/src/main/java/org/apache/sling/servlets/get/impl/helpers/StreamRenderer.java#L159

joerghoh:
@kwin Unfortunately just sending the link to the raw URL of the blob is not enough in this case.

kwin:
To me the URI being generated in https://github.com/apache/sling-org-apache-sling-jcr-resource/blob/15bfe9a6281618e5f14419c3979194cb5d0148f0/src/main/java/org/apache/sling/jcr/resource/internal/helper/jcr/BinaryDownloadUriProvider.java#L111 and exposed via `ExternalizedInputStream` look pretty much the same!

joerghoh:
Yes, that's correct, I will adapt the PR accordingly. Thanks @kwin for the pointer, I was not aware that we have in Sling that code already.

Need to wait for #2875 to be merged, before I can continue here.

PR #2857, opened 2024-09-13
#2656 - Accordion : Content editor dialog is adding "None" twice in Expanded …

Ranjani-JSB:
…items field #2656 - Update accordion.js

https://github.com/adobe/aem-core-wcm-components/issues/2656

| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes # 2656
| Patch: Bug Fix?          | Yes
| Minor: New Feature?      | No
| Major: Breaking Change?  | No
| Tests Added + Pass?      | N/A
| Documentation Provided   | Yes (code comments and or markdown)
| Any Dependency Changes?  | N/A
| License                  | Apache License, Version 2.0

PR #2890, opened 2024-11-07
SITES-26750 Accordion : Content editor dialog is adding "None" twice in Expanded Items field

RaduADumitru:

Steps to reproduce bug:
1. Add an accordion component and include three items, such as text components.
2. Navigate to the properties tab and select the "Single item expansion" checkbox.
3. Open the dropdown menu, which displays all three items along with an extra "None" option.
4. Go to the Items tab and delete one item, for example, item 2.
5. Return to the properties tab and open the dropdown menu again, observing that there are now two "None" options.

This is caused by the way expanded select is updated: it removes all the current items where value is not "", then adds "None", then readds items from children editor. Issue is fixed by also cleaning up "None" value, whereas this didn't happen before. Adapts fix from https://github.com/adobe/aem-core-wcm-components/pull/2857

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/editor/js/accordion.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/accordion/v1/accordion/clientlibs/editor/js/accordion.js
@@ -187,8 +187,8 @@
         var cmpChildrenEditor = $(childrenEditor).adaptTo("cmp-childreneditor");
         if (cmpChildrenEditor) {
             if (singleExpansion) {
-                // Check "None" option exists or not, if not then add textcontent
-                if(!expandedSelect.items._container.textContent.includes("None")){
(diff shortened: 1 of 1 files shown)


RaduADumitru:
2 more possible issues I noticed with the original fix, upon further inspection:

- The check for removing "None" option doesn't account for localization: if the localized value of "None" is different than the original, then the "None" option won't be removed. Could be fixed by also checking for the localized value in the check
- The check, since it uses includes() instead of strict equality, can also accept values that simply contain "None" in the text but are not this exact value. Could be fixed by using strict equality instead of includes

Will rectify

RaduADumitru:
Changed approach to do full cleanup, including of "None" element. Couldn't see any elements of value "" which we do want to keep, other than "None" element. Would be IMO the safest way to avoid other unforeseen problems, such as if other elements with value "" are added.

cbogdan0707 (approved):

PR #2912, opened 2025-01-27
[SITES-28496] CF date rendering fix

alexandru-stancioiu:
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


RaduADumitru (approved):
I searched for the difference between the two and I found the issue to be pretty interesting - leaving it here for reference:

https://support.pega.com/discussion/do-not-use-yyyy-uppercase-java-date-formatting#:~:text=In%20summary%2C%20%22yyyy%22%20represents,however%20they%20may%20be%20different.

> In summary, "yyyy" represents the calendar year whereas "YYYY" represents the year of the week, used in the ISO year-week calendar. In most cases, "yyyy" and "YYYY" yield the same number, however they may be different. That’s a subtle difference that only causes problems around a year change so your code could have been running perfectly fine all year around only to cause a problem in the new year.

spopescu (approved):

RaduADumitru (commented):
Can you also replace other instances of YYYY to yyyy, assuming it would be safe? I searched and found some in test content for instance:

https://github.com/adobe/aem-core-wcm-components/blob/main/bundles/core/src/test/resources/com/adobe/cq/wcm/core/components/internal/servlets/contentfragment/test-content.json

myronenko (approved):

myronenko:
> Can you also replace other instances of YYYY to yyyy, assuming it would be safe? I searched and found some in test content for instance:
> 
> https://github.com/adobe/aem-core-wcm-components/blob/main/bundles/core/src/test/resources/com/adobe/cq/wcm/core/components/internal/servlets/contentfragment/test-content.json

Agree, most probably there is no point to use `YYYY`

bundles/core/src/test/resources/com/adobe/cq/wcm/core/components/internal/servlets/contentfragment/test-content.json
```diff
@@ -84,11 +84,11 @@
...
                         "1538637399370": {
                             "jcr:primaryType": "nt:unstructured",
-                            "valueFormat": "YYYY-MM-DD[T]HH:mm:ss.000Z",
-                            "emptyText": "YYYY-MM-DD HH:mm",
+                            "valueFormat": "yyyy-MM-DD[T]HH:mm:ss.000Z",
+                            "emptyText": "yyyy-MM-DD HH:mm",
                             "listOrder": "5",
                             "valueType": "calendar",
-                            "displayedFormat": "YYYY-MM-DD HH:mm",
+                            "displayedFormat": "yyyy-MM-DD HH:mm",
```
RaduADumitru, line 91: Final comment: dug further and apparently dd is day of month while DD is day of year (example February 1 is dd 01 and DD 32: https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html

Can you also change DD to dd to avoid future confusion here?

RaduADumitru: MM is month while mm is minutes apparently - here it's fine

alexandru-stancioiu: sure, fixed DD.

PR #2918, opened 2025-02-13
ASSETS-46423 Support authoring preview versions of assets in sites core component

mohiaror:
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/ImageImpl.java
```diff
@@ -365,6 +378,13 @@ private String prepareNgdmSrcUriTemplate() {
         return ret;
     }
 
+    private String getRepoId(String remoteRepository) {
```
kwin, line 381: This deserves a bit of documentation, also the names are quite off here, the parameter called `remoteRepository` is populated by `NextGenDynamicMediaConfig.getRepositoryId()` so there seem to be two different kind of repository ids. Would be good to have a clearer distinction in method names and variable names.

mohiaror: @kwin This is still WIP. I forgot to move it to draft. Just did.

I will add the documentation once the approach is finalized.

mohiaror: BTW, there is only one repository. If `NextGenDynamicMediaConfig.getRepositoryId()` is configured then it will most probably be a remote repository. In some cases it can also be the repository on which page authoring is happening, but the way assets are picked will be different. This is already checked in a previous check in this class `isNgdmSupportAvailable` and only in that case is this code called.

mohiaror:
Closing this in favor of - https://github.com/adobe/aem-core-wcm-components/pull/2931

PR #2921, opened 2025-02-14
SITES-28693 - Teaser Component Renders Broken HTML When Title is Empty

### LINKED ISSUE #2660: [Teaser V2] Invalid HTML if teaser has link + image but no Title
## Bug Report

**Current Behavior**
When a teaser has:

1. The Link field set
2. An image set
3. Blank title

then the resulting HTML is invalid because of nested anchor tags - which is prohibited.

**Expected behavior/code**
The component should not create HTML that is invalid.

**Environment**
- AEM 6.5.19.0
- Core Components 2.23.4
- JRE version 11

**Possible Solution**
The issue stems from the `v2/TeaserImpl.java` model `initImage()` method:
```JAVA
    protected void initImage() {
        ...
        if (StringUtils.isNotEmpty(getTitle()) || getTeaserActions().size() > 0) {
            overriddenImageResourceProperties.put(Teaser.PN_IMAGE_LINK_HIDDEN, Boolean.TRUE.toString());
        }
        ...
    }
```

We see here that the link on the image is suppressed if there is a non-empty title or there are teaser actions.
However, if we look at the component html `teaser/v2/teaser/teaser.html` we see the following:
```HTML
    <a class="cmp-teaser__link"
       data-sly-attribute="${teaser.link.htmlAttributes}"
       data-sly-unwrap="${!teaser.link.valid || !teaser.actions.empty}"
       data-cmp-clickable="${teaser.data ? true : false}">
        <div class="cmp-teaser__content">
            <sly data-sly-call="${pretitleTemplate.pretitle @ teaser=teaser}"></sly>
            <sly data-sly-call="${titleTemplate.title @ teaser=teaser}"></sly>
            <sly data-sly-call="${descriptionTemplate.description @ teaser=teaser}"></sly>
            <sly data-sly-call="${actionsTemplate.actions @ teaser=teaser}"></sly>
        </div>
        <sly data-sly-call="${imageTemplate.image @ teaser=teaser}"></sly>
    </a>
```

Here we see that the outer anchor is unwrapped (removed) if the link is invalid or there are CTA links.

The problem is that the logic for if the image should be linked, and the logic for if the entire Teaser should be linked are not in agreement. This can lead to situations where both the Teaser and the Image are linked - which causes broken HTML because links are not allowed inside of links. 

The solution to this is to make sure that the case for when an image is linked and when the teaser is linked are mutually exclusive. 
I'm not entirely sure what the blankness of the title has to do with if the image should be linked or not, removing that condition from the V2 model implementation would probably resolve the issue. 

A more sustainable approach, that would require a change to the model interface, would be moving the logic for the `data-sly-unwrap` into a method called something like `boolean getLinkTeaser()` and then the HTL becomes `data-sly-unwrap="${!teaser.linkTeaser}"` and then this method can be re-used inside models `initImage()` method as such:
```JAVA
        if (this.getLinkTeaser()) {
            overriddenImageResourceProperties.put(Teaser.PN_IMAGE_LINK_HIDDEN, Boolean.TRUE.toString());
        }
```

or, better yet
```JAVA
        overriddenImageResourceProperties.put(Teaser.PN_IMAGE_LINK_HIDDEN, Boolean.valueOf(this.getLinkTeaser()).toString());
```

vladbailescu:
* Updated teaser to only allow image to render links if all other elements are empty (pre-title, title, description, actions), in which case it will unwrap the general link on the teaser content.

Fixes #2660

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/TeaserImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/TeaserImpl.java
@@ -82,7 +82,7 @@ public class TeaserImpl extends com.adobe.cq.wcm.core.components.internal.models
         overriddenImageResourceProperties.put(PN_ID, String.join(ID_SEPARATOR, this.getId(), IMAGE_ID_PREFIX));
 
         // Only allow image to render links if all other elements are empty
-        if (!(StringUtils.isAllEmpty(getTitle(), getPretitle(), getDescription()) && getTeaserActions().isEmpty())) {
+        if (!(StringUtils.isAllBlank(getTitle(), getPretitle(), getDescription()) && getTeaserActions().isEmpty())) {
             overriddenImageResourceProperties.put(Teaser.PN_IMAGE_LINK_HIDDEN, Boolean.TRUE.toString());
         }
         super.initImage();
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/TeaserImpl.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/TeaserImpl.java
@@ -81,7 +81,8 @@ public class TeaserImpl extends com.adobe.cq.wcm.core.components.internal.models
         overriddenImageResourceProperties.put(Teaser.PN_ACTIONS_ENABLED, Boolean.valueOf(actionsEnabled).toString());
         overriddenImageResourceProperties.put(PN_ID, String.join(ID_SEPARATOR, this.getId(), IMAGE_ID_PREFIX));
 
-        if (StringUtils.isNotEmpty(getTitle()) || getTeaserActions().size() > 0) {
+        // Only allow image to render links if all other elements are empty
+        if (!(StringUtils.isAllEmpty(getTitle(), getPretitle(), getDescription()) && getTeaserActions().isEmpty())) {
             overriddenImageResourceProperties.put(Teaser.PN_IMAGE_LINK_HIDDEN, Boolean.TRUE.toString());
         }
         super.initImage();
--- a/testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/TeaserIT.java
+++ b/testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/TeaserIT.java
@@ -1,9 +1,23 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2025 Adobe
+ ~
+ ~ Licensed under the Apache License, Version 2.0 (the "License");
+ ~ you may not use this file except in compliance with the License.
+ ~ You may obtain a copy of the License at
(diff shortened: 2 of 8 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/TeaserImpl.java
```diff
@@ -81,7 +81,8 @@ protected void initImage() {
         overriddenImageResourceProperties.put(Teaser.PN_ACTIONS_ENABLED, Boolean.valueOf(actionsEnabled).toString());
         overriddenImageResourceProperties.put(PN_ID, String.join(ID_SEPARATOR, this.getId(), IMAGE_ID_PREFIX));
 
-        if (StringUtils.isNotEmpty(getTitle()) || getTeaserActions().size() > 0) {
+        // Only allow image to render links if all other elements are empty
+        if (!(StringUtils.isAllEmpty(getTitle(), getPretitle(), getDescription()) && getTeaserActions().isEmpty())) {
```
spopescu, line 85: nitpick: I generally prefer isBlank()/isAllBlank() - it's a bit safer where whitespace is just as unwanted as null/empty strings.

vladbailescu: @spopescu , done, thanks!

LSantha:
@vladbailescu , it looks like, when image + valid link is set and pre-title, title, description, actions are blank then we still have nested anchors.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v2/TeaserImpl.java
```diff
@@ -81,7 +81,8 @@ protected void initImage() {
         overriddenImageResourceProperties.put(Teaser.PN_ACTIONS_ENABLED, Boolean.valueOf(actionsEnabled).toString());
         overriddenImageResourceProperties.put(PN_ID, String.join(ID_SEPARATOR, this.getId(), IMAGE_ID_PREFIX));
 
-        if (StringUtils.isNotEmpty(getTitle()) || getTeaserActions().size() > 0) {
+        // Only allow image to render links if all other elements are empty
+        if (!(StringUtils.isAllBlank(getTitle(), getPretitle(), getDescription()) && getTeaserActions().isEmpty())) {
```
LSantha, line 85: We may need '||' here instead of '&&" as before.

vladbailescu: I double-checked and I think this is correct:
image

vladbailescu: I will prepare some test content and an IT nevertheless

testing/it/http/src/test/java/com/adobe/cq/wcm/core/components/it/http/TeaserIT.java
```diff
@@ -0,0 +1,51 @@
+package com.adobe.cq.wcm.core.components.it.http;
```
LSantha, line 1: The build is failing because the file header is missing, reproducible locally by building the it subproject.

LSantha (approved):
LGTM.

PR #2928, opened 2025-03-21
SITES-30116 AEM Teaser/Image Core Component Alternative Text Field Issue

RaduADumitru:
| Fixed Issues?            | `SITES-30116` 
| Tests Added + Pass?      | Yes (existing)

Jira [SITES-30116](https://jira.corp.adobe.com/browse/SITES-30116)

Fixes alt text from page image not being shown inside image / teaser components on initial load.

Steps to reproduce:

0. Add a page image with alt text to a page inside WKND content, such as /content/wknd/es/es
1. Drag and drop an image or teaser component.
2. Select the "Inherit alternative text from page" checkbox in the Asset tab.
3. Observe the alternative text field, which should be populated but is instead disabled and blank.
4. Uncheck and then recheck the "Inherit alternative text from page" checkbox to see if the field populates correctly.

In the usual case it does not populate directly, likely due to a race condition because in certain cases, such as when console / debugger is open, the alt text is in fact populated correctly.

Issue is fixed by explicitly adding the alt text from page to the text field if it exists and the inherit alt text from page image checkbox is checked. The fix applies to both Image and Teaser Components since they use the same code here.

This scenario is covered by an existing IT, which was already passing though - [see here](https://github.com/adobe/aem-core-wcm-components/blob/main/testing/it/e2e-selenium/src/test/java/com/adobe/cq/wcm/core/components/it/seljup/tests/image/ImageTests.java#L417), likely due to the unpredictable nature of the race condition + added slowness by the Selenium IT environment

Additionally slightly cleaned up the code to use preexisting vars for selectors in a few places

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/clientlibs/editor/js/image.js
+++ b/content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/clientlibs/editor/js/image.js
@@ -173,6 +173,14 @@
                         captionTuple.hideCheckbox(true);
                     }
                 }
+
+                // Ensure alt text from page is visible on initial load if imageFromPageImage is checked
+                if (imageFromPageImage && imageFromPageImage.checked) {
+                    var isAltFromPageImageChecked = document.querySelector('coral-checkbox[name="./altValueFromPageImage"]').checked;
+                    if (isAltFromPageImageChecked && altTextFromPage) {
+                        $altTextField.adaptTo("foundation-field").setValue(altTextFromPage);
+                    }
+                }
             });
         }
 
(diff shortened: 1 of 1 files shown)


cbogdan0707 (approved):

spopescu (approved):

RaduADumitru:
Jenkins build Job on this branch succeeded without any test failures

content/src/content/jcr_root/apps/core/wcm/components/image/v3/image/clientlibs/editor/js/image.js
```diff
@@ -97,7 +98,7 @@
 
             imageFromPageImage = dialogContent.querySelector("coral-checkbox[name='./imageFromPageImage']");
 
-            altFromPageTuple = new CheckboxTextfieldTuple(dialogContent, "coral-checkbox[name='./altValueFromPageImage']", "input[name='./alt']");
+            altFromPageTuple = new CheckboxTextfieldTuple(dialogContent, altFromPageCheckboxSelector, altInputSelector);
```
RaduADumitru, line 101: "input[name='./alt']" was already defined as altInputSelector, so used the var instead

spopescu (approved):

PR #2930, opened 2025-03-31
[Carousel] Accessibility improvments

renow-luxembourg:
### Feature Request
In a recent accessibility audit, several issues were found in the carousel component :
- For people using vocalization tool (NVDA, JAWS, VoiceOver, etc.), each slide content is vocalized twice due to the 2 aria-live attributes.
- Controls and Indicators should come before the slides content in the tabulation order
- Indicator are not initially focusable, we have to add the tabindex attribute before starting the carousel (mostly in carousel without autoplay)
- On each slide there are 2 attributes that give the accessible name : aria-label and aria-labelledby, some vocalization tool read both and it's a bit annoying
- When the carousel controls buttons only contain an icon, it's a nice-to-have to allow the user to display to know what this button is about using a title attribute
- Carousel indicator item should be actionable using the spacebar. The easy way to do it is to use the native button element.

**Environment**
AEM 6.5
Core Components version 2.28.0

**Documentation**
We have to implement the properties specified in the design pattern of carousel and tab :
- https://www.w3.org/WAI/ARIA/apg/patterns/carousel/
- https://www.w3.org/WAI/ARIA/apg/patterns/tabs/


vladbailescu:
@renow-luxembourg , thank you for submitting this PR! Since it changes the indicators element and position in the DOM structure, it's unfortunately not backwards compatible and would break existing implementations (for Cloud Service Core Components get automated upgrades). We might be able to consider this change for a future version of the Carousel component, meanwhile we recommend customers implement such modifications by overlaying in their own proxy components.

PR #2931, opened 2025-04-02
ASSETS-46423 Support authoring preview versions of assets in sites core component

mohiaror:
Add a preview token builder which, on getting a preview key in environment variables, will always generate and append the preview token to DM with OpenAPIs images. This preview token will ensure that preview version of an asset is visible in image component if it exists. For customers who do not have the preview secret key set in their environment variables, the existing delivery URL will be generated.

| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/PreviewTokenBuilderUtils.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/PreviewTokenBuilderUtils.java
@@ -1,0 +1,80 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2025 Adobe
+ ~
+ ~ Licensed under the Apache License, Version 2.0 (the "License");
+ ~ you may not use this file except in compliance with the License.
+ ~ You may obtain a copy of the License at
+ ~
+ ~     http://www.apache.org/licenses/LICENSE-2.0
+ ~
+ ~ Unless required by applicable law or agreed to in writing, software
+ ~ distributed under the License is distributed on an "AS IS" BASIS,
+ ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ ~ See the License for the specific language governing permissions and
+ ~ limitations under the License.
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+package com.adobe.cq.wcm.core.components.internal;
+
+import java.nio.charset.StandardCharsets;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.text.SimpleDateFormat;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.AbstractMap;
+import java.util.Date;
+import java.util.Map;
+import java.util.TimeZone;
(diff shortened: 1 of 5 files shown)


mohiaror:
@LSantha , @vladbailescu  Can you please help review and merge this PR? I have added sufficient test cases to cover all the newly added functionality.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/PreviewTokenBuilderUtils.java
```diff
@@ -0,0 +1,82 @@
...
+        SimpleDateFormat dateFormat = new SimpleDateFormat(DATE_PATTERN);
+        dateFormat.setTimeZone(TimeZone.getTimeZone(UTC));
+        return dateFormat;
+    });
+
+    public static Map.Entry<String, String> buildPreviewToken(String assetId) {
+        try {
+            String secretKey = readKeyFromEnvVar(PREVIEW_KEY);
+            if (assetId == null || StringUtils.isBlank(secretKey)) {
+                throw new Exception("Invalid input parameters");
```
LSantha, line 54: I would recommend simply to return null here, without throwing an exception only to cache it later in the same method (flow control with exceptions anti-pattern) especially since this feature may not be enabled for most customers. You may still do debug logging like feature not configured.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/PreviewTokenBuilderUtils.java
```diff
@@ -0,0 +1,82 @@
...
+            }
+            LocalDateTime nowPlusThirty = LocalDateTime.now().plusMinutes(30);
+            Date expirationTime = Date.from(nowPlusThirty.atZone(ZoneId.systemDefault()).toInstant());
+            final String expiryTimeStr = dateFormat.get().format(expirationTime);
+            return new AbstractMap.SimpleEntry<>(computeSignature(assetId, expiryTimeStr, secretKey), expiryTimeStr);
+        } catch (Exception e) {
+            // we do not want to put warn or error here because for most of the customers preview key will not be set
+            // in that case normal flow should work for them without any warnings.
+            LOG.debug("Could not generate preview token for asset {}. Exception : {}", assetId, e.getMessage());
```
LSantha, line 63: With the comment above you may consider logging the real errors here not only on debug level.

mohiaror: @LSantha  makes sense. I have incorporated your review comments and updated the code and tests with commit - https://github.com/adobe/aem-core-wcm-components/pull/2931/commits/4ebe107492d58e3683e37cced0984da9a4ac0d89

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/util/ngdm/PreviewTokenBuilderUtils.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Date;
+import java.util.Map;
+import java.util.TimeZone;
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
```
LSantha, line 34: Please add javadoc to public class and public methods.

indra2gurjar: Done

LSantha (approved):

PR #2942, opened 2025-05-13
[Tabs] Accessibility improvments

renow-luxembourg:
### Feature Request

In a recent accessibility audit, it was observed that the Tabs component does not follow the structure of the W3C design pattern : Tabs indicator should use the "button" tag with the role="tab" (element with"button" tag already have an implicit "tabindex" and natively support the enter/space keyboard activation)

To take into account the case of multiple tabs (or tabs with a long text content), we organized some user test. It seems that it's better to have all the tabs in one row in desktop and mobile with an horizontal scroll. To fix-it, we also add a tabindex="0" attribute on the tablist container and we do a small improvment to the CSS file. 

We are also thinking about adding an "Orientation" parameter to describe if the tablist will be displayed horizontally or vertically ( [branch also available on github](https://github.com/renow-luxembourg/aem-core-wcm-components/tree/renow_fix_tabs_orientation))

**Environment**
AEM 6.5
Core Components version 2.28.0

**Documentation**
We have to implement the properties specified in the design pattern of tabs :
https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

### DIFF
--- a/content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/clientlibs/site/css/tabs.less
+++ b/content/src/content/jcr_root/apps/core/wcm/components/tabs/v1/tabs/clientlibs/site/css/tabs.less
@@ -16,7 +16,8 @@
 
 .cmp-tabs__tablist {
     display: flex;
-    flex-wrap: wrap;
+    overflow: auto hidden;
+    flex-wrap: nowrap;
     padding-left: 0;
     list-style: none;
 }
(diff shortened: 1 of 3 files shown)


vladbailescu:
@renow-luxembourg , thank you for submitting this PR! Since this changes the HTML elements for the tabs, it is a breaking change for existing implementations and can only be considered for a future version of the Tabs component.

PR #2951, opened 2025-06-13
Issue #2950 [LanguageNavigation v2] Accessibility and usability improvments 

### LINKED ISSUE #2950: [LanguageNavigation v2] Remove nav element + improvement
## Feature Request

In a recent accessibility audit, it was observed that the LanguageNavigation component doesn't fully respect the RGAA/RAWeb accessibility guidelines for French-speaking countries.
Main relevant problem :
- LanguageNavigation should not use the nav tag as it's not considered as a navigation following the Glossary definition (see below) 

Other improvement nice-to-have requested by some of our Client :
- Automatically hide the component if the website is in only 1 language
- Display the ISO code instead of the full language name (e.g EN instead of English, FR instead of FRENCH)
- Add a parameter to display the language as a dropdown list using the disclosure design pattern (we have some website with more than 15 languages)

Replace the nav element by a div. The aria-label attribute need to be transferred to the ul element to be taken into account by accessibility tools.

**Are there alternatives?**
An alternative can be to use the role="presentation" instead of role="navigation" to remove the element semantic. But it isn't super clean ...

**Documentation**
Glossary : https://accessibilite.public.lu/en/raweb1/glossaire.html#menu-and-navigation-bar

renow-luxembourg:
| Fixed Issues?            | Fixes #2950 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Any Dependency Changes?  |

Changes description :
- Replace the nav container by a div and move the aria-label attribute to the ul element
- Automatically hide the component if the website is in only 1 language
- Display both the ISO code and the full language name. One of them can be visually hidden using some CSS similar to the "sr-only" class.
- Add a parameter in the cq_dialog to display the language as a dropdown list using the disclosure design pattern

If it's easier, I can split this pull request into several.

### DIFF
--- a/.github/workflows/maven-deploy.yml
+++ b/.github/workflows/maven-deploy.yml
@@ -14,7 +14,7 @@
 # limitations under the License.
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-name: Deploy to OSSRH
+name: Deploy to Central Repository
 
 # Run workflow on commits to default branch
 on: workflow_dispatch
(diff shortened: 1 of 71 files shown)


vladbailescu:
@renow-luxembourg , thank you for submitting this PR! Since the changes affect the HTML structure of the component output, they are not backwards compatible and will break existing implementations on automatic upgrades, therefore they can only be considered for a future version of the LanguageNavigation component.

PR #2961, opened 2025-07-22
SITES-33687 - URL with Query + Fragment Reordered by Teaser Core Component

vladbailescu:
* Updated fragment processing logic
* Added test for edge case

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
@@ -115,7 +115,7 @@ public class LinkUtil {
      */
     public static String escape(final String path, final String queryString, final String fragment) {
         boolean pathContainsFragment = false;
-        if (StringUtils.contains(path, fragment)) {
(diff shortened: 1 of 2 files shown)


spopescu (approved):

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
```diff
@@ -115,7 +115,7 @@ public static String decode(final String url) throws UnsupportedEncodingExceptio
      */
     public static String escape(final String path, final String queryString, final String fragment) {
         boolean pathContainsFragment = false;
-        if (StringUtils.contains(path, fragment)) {
+        if (StringUtils.contains(path, "#" + fragment)) {
```
RaduADumitru, line 118: Initially thought the issue could still happen in theory if the # character were to be used in other ways in the url, but I checked and it looks like it is a reserved character based on RFC Specification for URLs, so this shouldn't happen - [see here](https://stackoverflow.com/questions/1856785/characters-allowed-in-a-url)

RaduADumitru (approved):

PR #2998, opened 2025-10-17
issues/2996

pat-lego:
Skip experience fragments


vladbailescu:
@pat-lego , I added a null check to your PR and now it passes tests. Is this still something you would like added? If yes, please update the PR description to reflect why you would like this change so team can review and include in a future release!

PR #3018, opened 2026-03-11
[Carousel] Accessibility improvments - v2

renow-luxembourg:
Here is a simplify version of the following PR: https://github.com/adobe/aem-core-wcm-components/pull/2930

This PR doesn't modify the DOM structure to avoid any backwards compatibility issues.

Suggestions that are regression-free to improve accessibility :
- Add the title attribute to the action buttons (previous, next, play and pause)
- Remove the double speech by deleting one of the aria-live attribute
- Correct accessible name of the the tabpanel by removing the aria-labelledby attribute (an aria-label is already present)
- Add the tabindex attribute to the tabpanel to improve the keyboard navigation
- Initially add the tabindex and aria-selected attributes to the tab to improve the keyboard navigation
- Add a title attribute to the indicator tabs buttons


content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
```diff
@@ -19,9 +19,8 @@
      data-panelcontainer="${wcmmode.edit && 'carousel'}"
      id="${carousel.id}"
      class="cmp-carousel"
-     role="group"
+     role="region"
```
santhanbattina, line 22: Switching from role="group" to role="region" makes sense here since the carousel is a distinct section. Just wondering if we should ensure carousel.accessibilityLabel is always present so the region doesn’t end up unnamed for screen readers.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
```diff
@@ -19,9 +19,8 @@
...
      aria-label="${carousel.accessibilityLabel}"
-     aria-live="polite"
```
santhanbattina, line 24: Removing aria-live="polite" seems like a good change. Auto-rotating carousels can otherwise keep announcing updates and become noisy for screen reader users.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
```diff
@@ -40,11 +39,11 @@
              id="${item.id}-tabpanel"
              class="cmp-carousel__item${item.name == carousel.activeItem ? ' cmp-carousel__item--active' : ''}"
              role="tabpanel"
-             aria-labelledby="${item.id}-tab"
```
santhanbattina, line 43: The new "Slide X of Y" label is clearer than relying on aria-labelledby. Quick question: does itemList.count start at 1 here? Just checking to avoid the first slide being announced as “Slide 0”.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
```diff
@@ -40,11 +39,11 @@
...
              aria-roledescription="slide"
              aria-label="${(carousel.accessibilityAutoItemTitles && item.title) || 'Slide {0} of {1}' @ format=[itemList.count, carousel.items.size], i18n}"
              data-cmp-data-layer="${item.data.json}"
-             data-cmp-hook-carousel="item"></div>
+             data-cmp-hook-carousel="item"
+             tabindex="-1"></div>
```
santhanbattina, line 46: I noticed tabindex="-1" was added to the slide container. Is this used for programmatic focus when slides change? Just wanted to confirm the intent.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/controls.html
```diff
@@ -19,13 +19,15 @@
         <button class="cmp-carousel__action cmp-carousel__action--previous"
                 type="button"
                 aria-label="${carousel.accessibilityPrevious || 'Previous' @ i18n}"
+				title="${carousel.accessibilityPrevious || 'Previous' @ i18n}"
```
santhanbattina, line 22: Adding the title attribute on the navigation buttons looks helpful for hover tooltips. Since we already have aria-label, this shouldn’t affect accessibility behavior, right?

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
LSantha: @renow-luxembourg , do we really need this role change? This component does not guarantee an accessible name because accessibilityLabel is optional, so switching from group to region can turn existing unlabeled carousels into unnamed component. It may be safer to keep it as is.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
LSantha: This may break the tab to panel labeling relationship. Could we avoid this?

LSantha:
@renow-luxembourg , you may need to adjust ComponentsIT.testCarousel() since the markup is different now. Please try to execute the ITs locally and see.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
renow-luxembourg: The role=region attribute value is recommended by the [W3C design pattern](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-2-tablist/).
I think that it's only a nice to have.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
renow-luxembourg: The aria-labelledby attribute has a higher priority than the aria-label attribute in calculating the accessibility name.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/carousel.html
renow-luxembourg: The tabindex attribute of the tabpanel allows the focus to be visible when this element receives focus programmatically.

content/src/content/jcr_root/apps/core/wcm/components/carousel/v1/carousel/controls.html
renow-luxembourg: The aria-label attribute takes precedence over the title attribute for calculating the accessibility name. Therefore, there will be no difference for the speaker tool reading the text. However, mouse users can obtain additional information by hovering their cursor over the text.

renow-luxembourg:
@LSantha, I've adapted the tests manually but I can't get to run ComponentsIT.testCarousel() locally with JUnit (LoginToken unauthorized). Could you take a look at my latest commit if it's ok for you ?

renow-luxembourg:
I consulted with our accessibility experts about using role="group" instead of role="region". Even though the design pattern proposes region, MDN clarifies that “The region role must have an accessible name” and that it “should be reserved for content sections significant enough that users may want to navigate to them directly.”  
Given this, we believe role="group" may be more appropriate in this context.
Source: [MDN — section element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section#technical_summary)
I reverted my code changes to keep role="group".
During our screen reader tests, we also observed that the control buttons (Previous, Next, and the navigation dots) should not remain inside the aria-live region, as this leads to excessive and repetitive announcements.
I have created [another branch in my repo](https://github.com/renow-luxembourg/aem-core-wcm-components/tree/renow_fix_carousel3) with this modification, though I am worried it might cause some backward‑compatibility problems.
What are your thoughts on this?

PR #3021, opened 2026-04-08
SITES-42596 Core Image Component floods logs with “LinkUtil abs_path requested” when Dynamic Media hotspot images are drag-dropped - avoid logging error for standard quickview urls since they are supported product functionality

RaduADumitru:
| Fixed Issues?           | Jira [SITES-42596](https://jira.corp.adobe.com/browse/SITES-42596)

Jira [SITES-42596](https://jira.corp.adobe.com/browse/SITES-42596)

## Summary

After dragging an image onto an Image V2 or V3 component that was previously configured for Dynamic Media with hotspots, rendering the image produced unnecessary **error** logs. The `quickview:` URL is treated as invalid by `LinkUtil.escape`, which logged an error even though fallback URL construction still succeeds.

Quickview is supported OOTB for Dynamic Media interactive images, so this scenario should not be logged as an error. The change logs a **debug** message when the quickview scheme is detected in the escape fallback path (helpful for tracing hotspot usage) and keeps **error** logging for other failures.

**Code changes**

- `LinkUtil.escape`: on exception, if the parsed scheme is `quickview`, log at DEBUG with a short message; otherwise keep existing ERROR logging.
- `LinkUtilTest`: Logback `ListAppender` tests assert quickview paths emit DEBUG and no ERROR; null inputs and a custom scheme still emit ERROR as before. Minor cleanup: remove unnecessary `throws` on mailto/tel tests.

## Reproduction (WKND)

- Configure a Dynamic Media image with a hotspot per [interactive images](https://experienceleague.adobe.com/en/docs/experience-manager-65/content/assets/dynamic/interactive-images).
- Add Image V2 or V3 to a page and drag-drop that asset onto the component.

Previously, logs showed `LinkUtil` / `URIException: abs_path requested` at ERROR. With this PR, quickview cases log at DEBUG instead.

## Alternatives considered

- Filter quickview URLs in Image V2 `ImageImpl.buildAreas()` — would drop areas from the model and HTL output.
- Special-case quickview before URI construction — would encode the URL and change behavior for customer code relying on the raw quickview string ([interactive images – Cloud Service](https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/assets/dynamicmedia/interactive-images)).

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
@@ -155,6 +155,8 @@ public class LinkUtil {
 
         } catch (Exception e) {
             if (parsed != null && isQuickviewLink(parsed.getScheme())) {
+                // SITES-42596 Since quickview URLs are supported OOTB by the product,
+                // do not log error for them
                 LOG.info("Quickview URL detected, using fallback construction: {}", path);
             } else {
                 LOG.error(e.getMessage(), e);
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
@@ -154,10 +154,10 @@ public class LinkUtil {
             }
 
         } catch (Exception e) {
-            if (parsed != null && isQuickviewLink(parsed.getScheme())) {
+            if (parsed != null && isQuickviewScheme(parsed.getScheme())) {
                 // SITES-42596 Since quickview URLs are supported OOTB by the product,
                 // do not log error for them
-                LOG.info("Quickview URL detected, using fallback construction: {}", path);
+                LOG.debug("Quickview URL detected, using fallback construction: {}", path);
             } else {
                 LOG.error(e.getMessage(), e);
             }
@@ -296,11 +296,7 @@ public class LinkUtil {
         }
     }
 
-    private static boolean isQuickviewLink(String link) {
-        if (link != null) {
-            return link.startsWith(QUICKVIEW_PATTERN);
-        } else {
-            return false;
-        }
+    private static boolean isQuickviewScheme(String scheme) {
+        return QUICKVIEW_PATTERN.equals(scheme);
     }
 }
--- a/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
+++ b/bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
@@ -65,6 +65,8 @@ public class LinkUtil {
 
     private static final String TEL_PATTERN = "tel";
 
+    private static final String QUICKVIEW_PATTERN = "quickview";
+
     //SITES-18137: imitate the exact behavior of com.google.common.net.URL_FRAGMENT_ESCAPER
     private static final BitSet URL_FRAGMENT_SAFE_CHARS = new BitSet(256);
 
@@ -152,7 +154,11 @@ public class LinkUtil {
             }
 
         } catch (Exception e) {
-            LOG.error(e.getMessage(), e);
+            if (parsed != null && isQuickviewLink(parsed.getScheme())) {
+                LOG.info("Quickview URL detected, using fallback construction: {}", path);
+            } else {
+                LOG.error(e.getMessage(), e);
+            }
             StringBuilder sb = new StringBuilder(path);
             if (queryString != null) {
                 sb.append("?").append(maskedQueryString);
@@ -283,6 +289,14 @@ public class LinkUtil {
     private static boolean isTelLink(String link) {
         if (link != null) {
             return link.startsWith(TEL_PATTERN);
+        } else {
+            return false;
+        }
+    }
+
+    private static boolean isQuickviewLink(String link) {
+        if (link != null) {
+            return link.startsWith(QUICKVIEW_PATTERN);
         } else {
             return false;
         }
--- a/bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtilTest.java
+++ b/bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtilTest.java
@@ -94,7 +94,7 @@ class LinkUtilTest {
     @Test
     void escape_nullInputs() {
         String escaped = LinkUtil.escape(null, null, null);
-        assertNull(null, escaped);
+        assertEquals("", escaped);
     }
 
     @Test
--- a/bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtilTest.java
+++ b/bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtilTest.java
@@ -90,4 +90,17 @@ class LinkUtilTest {
         String escapedQuickview = LinkUtil.escape(path, null, null);
         assertEquals(path, escapedQuickview);
     }
+
+    @Test
+    void escape_nullInputs() {
+        String escaped = LinkUtil.escape(null, null, null);
+        assertNull(null, escaped);
+    }
+
+    @Test
+    void escape_customLinkResultEqualToInput() {
+        String path = "custom:abc";
+        String escaped = LinkUtil.escape(path, null, null);
+        assertEquals(path, escaped);
+    }
 }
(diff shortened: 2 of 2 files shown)


bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
```diff
@@ -152,7 +154,13 @@
             }
 
         } catch (Exception e) {
-            LOG.error(e.getMessage(), e);
+            if (parsed != null && isQuickviewLink(parsed.getScheme())) {
+                // SITES-42596 Since quickview URLs are supported OOTB by the product,
+                // do not log error for them
+                LOG.info("Quickview URL detected, using fallback construction: {}", path);
```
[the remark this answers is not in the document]
RaduADumitru: Don't see path information as sensitive since it is stored in the content structure and intended to be exposed

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
```diff
@@ -152,7 +154,13 @@ public static String escape(final String path, final String queryString, final S
...
```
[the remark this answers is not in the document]
RaduADumitru: Removed the placeholders

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
```diff
@@ -152,7 +154,13 @@ public static String escape(final String path, final String queryString, final S
...
+            } else {
+                LOG.error(e.getMessage(), e);
+            }
```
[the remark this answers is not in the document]
RaduADumitru: The general quickview URL case which is added by the product leads to the second case; errors in the first may actually point to invalid url configuration, even if it is technically a quickview url.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtilTest.java
```diff
@@ -71,16 +71,36 @@ void escape_whenEscapingPathWithFragment_thenFragmentForwardSlashIsNotEncoded()
...
+        String path = "quickview:fragment=Camping in Western Australia&size=400,300&reservedVal_productPath=/content/experience-fragments/wknd/ca/en/featured/camping-western-australia";
+        String escapedQuickview = LinkUtil.escape(path, null, null);
+        assertEquals(path, escapedQuickview);
+    }
+
+    @Test
+    void escape_nullInputs() {
+        String escaped = LinkUtil.escape(null, null, null);
+        assertNull(null, escaped);
+    }
```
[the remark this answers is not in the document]
RaduADumitru: Corrected the test and made it more specific, checks for "" now which is the string returned in this case. Looks like assertNull considers "" as null

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
```diff
@@ -287,4 +295,12 @@ private static boolean isTelLink(String link) {
...
         }
     }
+
+    private static boolean isQuickviewLink(String link) {
+        if (link != null) {
+            return link.startsWith(QUICKVIEW_PATTERN);
+        } else {
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
RaduADumitru: Changed to exact comparison of the scheme

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
```diff
@@ -152,7 +154,13 @@ public static String escape(final String path, final String queryString, final S
             }
 
         } catch (Exception e) {
-            LOG.error(e.getMessage(), e);
+            if (parsed != null && isQuickviewLink(parsed.getScheme())) {
+                // SITES-42596 Since quickview URLs are supported OOTB by the product,
+                // do not log error for them
+                LOG.info("Quickview URL detected, using fallback construction: {}", path);
+            } else {
```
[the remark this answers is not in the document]
RaduADumitru: Logging at debug now. Path / url info is not sensitive

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtilTest.java
```diff
@@ -71,16 +71,36 @@ void escape_whenEscapingPathWithFragment_thenFragmentForwardSlashIsNotEncoded()
...
         String escapedTel = LinkUtil.escape(path, null, null);
         assertEquals(path, escapedTel);
     }
+
+    @Test
+    void escape_quickviewLinkResultEqualToInput() {
+        String path = "quickview:fragment=Camping in Western Australia&size=400,300&reservedVal_productPath=/content/experience-fragments/wknd/ca/en/featured/camping-western-australia";
+        String escapedQuickview = LinkUtil.escape(path, null, null);
+        assertEquals(path, escapedQuickview);
+    }
```
myronenko: The entire purpose of this PR is to change `ERROR` → `DEBUG` for quickview URLs, but `escape_quickviewLinkResultEqualToInput` only asserts the return value. It doesn't verify that `LOG.debug()` was called instead of `LOG.error()`. A Mockito mockStatic on LoggerFactory or a log appender assertion would make this test actually prove the bug is fixed. Without it, someone could revert the log change and all tests would still pass.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
```diff
@@ -287,4 +295,8 @@ private static boolean isTelLink(String link) {
             return false;
         }
     }
+
+    private static boolean isQuickviewScheme(String scheme) {
+        return QUICKVIEW_PATTERN.equals(scheme);
```
myronenko, line 300: [RFC 3986 §3.1](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1) says URI schemes are case-insensitive.
`QUICKVIEW_PATTERN.equals(scheme)` won't match `Quickview:` or `QUICKVIEW:`.
Copilot's original suggestion of `equalsIgnoreCase` was more correct. In practice DM likely always produces lowercase, so this is low risk.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
myronenko, line 134: The quickview detection only applies in the second `catch (Exception e)` block. The first catch `(URIException e)` still unconditionally logs at `ERROR`.
If a malformed quickview URL triggers the first parse failure `(new URI(maskedPath, false))`, `parsed` will be `null`, and the second `catch` will also log at `ERROR` since the `parsed != null` guard fails.

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtilTest.java
RaduADumitru: Very good point, I wasn't aware that there were ways to test this, but I've checked and there is a pattern already used in the codebase for testing logs.

I've applied this pattern to the newly added tests as well. Test for quickview will check that debug is logged without any error logs, while other cases are shown to log error.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
RaduADumitru: DM produces lowercase URLs and I wouldn't see reasons for this to change; any other interventions would be made by external factors, the intent here is to bypass error logging for the standard product cases only. For this reason I'd still leave it as it is now. The exact case matching also matches other scheme comparisons made in this class like for `tel` and `mailto`.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtil.java
RaduADumitru: Copilot also brought up a similar point, but in this code it is intended that only the second parse failure handles the quickview url case, since this is the one that will be reached in the general product case. If the first parse were to fail, this would indicate an actual malformation of the URL which should be logged.

Also, the first parse failure throwing an error does not automatically make the second parse throw as well; this scenario is tested in the newly added null input test case, where I have added an assertion that only one error will be logged (the first parse will fail because the URI constructor used there fails on null inputs).

Of course, there could be cases where both the first and second parse fail, but in this case I'd consider it justified to log both parse attempts to better indicate the issues that led to both parses failing - this was also the preexisting pattern, unrelated to the scope of the PR itself

bundles/core/src/test/java/com/adobe/cq/wcm/core/components/internal/link/LinkUtilTest.java
myronenko: Fixed in https://github.com/adobe/aem-core-wcm-components/pull/3021/changes/b575551714fdafc2164043ea473b194a2d84ad81

myronenko (approved):

PR #3032, opened 2026-05-11
SITES-36662 - [CoreCmp][IT] Update integration tests for PDF Viewer component to not use Granite clientlibs

vladbailescu:
* Fixes https://jira.corp.adobe.com/browse/SITES-36662
* Created dedicated clientlibs, templates and policies that do not include Forms clientlibs (which in turn include Granite clientlibs)
* Update PDF Viewer examples and IT content to use the new templates


YahorC (approved):

examples/ui.apps/src/content/jcr_root/apps/core-components-examples/clientlibs/clientlib-base/.content.xml
```diff
@@ -3,4 +3,4 @@
     jcr:primaryType="cq:ClientLibraryFolder"
     allowProxy="{Boolean}true"
     categories="[cmp-examples.base]"
-    embed="[cmp-examples.fonts,core.wcm.components.accordion.v1,core.wcm.components.tabs.v1,core.wcm.components.carousel.v1,cmp-examples.fonts,core.wcm.components.pdfviewer.v1,core.wcm.components.image.v3,core.wcm.components.image.v2,core.wcm.components.breadcrumb.v2,core.wcm.components.search.v2,core.wcm.components.form.text.v2,cmp-theme.core-components-clean,core.wcm.components.form.container.v2,core.wcm.components.text.v2,core.wcm.components.embed.v1,cmp-examples.site,core.wcm.components.commons.site.link]"/>
+    embed="[cmp-examples.fonts,core.wcm.components.accordion.v1,core.wcm.components.tabs.v1,core.wcm.components.carousel.v1,cmp-examples.fonts,core.wcm.components.pdfviewer.v1,core.wcm.components.image.v3,core.wcm.components.image.v2,core.wcm.components.breadcrumb.v2,core.wcm.components.search.v2,cmp-theme.core-components-clean,core.wcm.components.text.v2,core.wcm.components.embed.v1,cmp-examples.site,core.wcm.components.commons.site.link]"/>
```
myronenko: `cmp-examples.fonts` appears twice in the `embed` list (positions 1 and 5). The duplicate causes the fonts stylesheet to be loaded twice — remove the second occurrence.

```suggestion
    embed="[cmp-examples.fonts,core.wcm.components.accordion.v1,core.wcm.components.tabs.v1,core.wcm.components.carousel.v1,core.wcm.components.pdfviewer.v1,core.wcm.components.image.v3,core.wcm.components.image.v2,core.wcm.components.breadcrumb.v2,core.wcm.components.search.v2,cmp-theme.core-components-clean,core.wcm.components.text.v2,core.wcm.components.embed.v1,cmp-examples.site,core.wcm.components.commons.site.link]"/>
```

testing/it/it.ui.content/src/content/jcr_root/conf/core-components/settings/wcm/policies/.content.xml
```diff
@@ -186,7 +186,7 @@
                                 jcr:primaryType="nt:unstructured"
                                 jcr:title="Default page policy"
                                 sling:resourceType="wcm/core/components/policy/policy"
-                                clientlibs="cmp-test.base"
+                                clientlibs="cmp-test.base.forms"
```
myronenko, line 189: The "Default page policy" for all IT pages is switched from `cmp-test.base` to `cmp-test.base.forms`, which makes every IT page load form clientlibs (and transitively Granite clientlibs). This is the opposite direction from the PR's stated goal of isolating form dependencies. If the intent is to keep existing form-dependent tests passing, this should be explained in the PR description. If the intent is to restrict form clientlibs to PDF Viewer pages only, a separate template + policy for those IT pages (parallel to the `base-page` approach in the examples module) would be the consistent approach.

LSantha:
A few observations to consider from my augment based review:

"The examples-side restructuring looks clean — splitting cmp-examples.base / cmp-examples.base.forms, the new base-page template, and switching the PDF Viewer example page to it is exactly the shape we want for SITES-36662.

A few points worth addressing before merging:

**1. IT side doesn’t actually validate the no-Granite scenario**

The PR’s stated goal is to ensure ITs exercise PDF Viewer without Forms/Granite clientlibs, but:

- The deployed PDF Viewer IT page (/content/core-components/pdfviewer) uses simple-template-v3 (page v3), not the v2 page template whose policy this PR edits. So the IT policy change has no effect on that page.
- PdfViewerIT.java creates its page at runtime via Commons.createPagePolicy(...) with only CLIENTLIBS_PDFVIEWER_V1, so it also doesn’t exercise the regression path from SITES-22059.

As it stands, the IT module wouldn’t have caught the original regression.

**2. The IT default-policy switch goes the opposite direction**

Switching the page-v2 “Default page policy” (policy_1606475224445) from cmp-test.base to cmp-test.base.forms means every IT page resolving to that policy (simple-template, responsive-template — used by teaser, etc.) now loads Forms+Granite clientlibs. That’s the opposite of the PR goal and broadens the runtime footprint of many unrelated IT pages.

I’d suggest mirroring the examples-side approach in IT:

- Keep the default policy on cmp-test.base (no forms).
- Add a dedicated cmp-test.base.forms-bound policy + template for the pages/tests that actually need forms (the form IT tests use defaultPageTemplate via AdminBaseUITest, so we’d need to either give them their own template or update the test to use the forms-enabled policy explicitly).
- Add a small no-forms PDF Viewer IT page (or switch the existing one) so there is at least one test path that runs with cmp-test.base only."

examples/ui.apps/src/content/jcr_root/apps/core-components-examples/clientlibs/clientlib-base/.content.xml
vladbailescu: It was duplicated before this PR, I wanted to keep the changes to a minimum. We can include this cleanup too, thanks!

> The duplicate causes the fonts stylesheet to be loaded twice

That is incorrect, embeded clientlibs are all appended (and deduplicated) in one file.

https://experienceleague.adobe.com/en/docs/experience-manager-65/content/implementing/developing/introduction/clientlibs#embedding-code-from-other-libraries

testing/it/it.ui.content/src/content/jcr_root/conf/core-components/settings/wcm/policies/.content.xml
vladbailescu: The new `cmp-test.base.forms` effectively replaces previous `cmp-test.base`, with the exact same contents. For now, I only switched PDF Viewer pages to use the new/slimmer `cmp-test.base`.

vladbailescu:
@LSantha , thanks for your review!

> 1. IT side doesn’t actually validate the no-Granite scenario

I only switched the HTTP IT content to use the new policy. As you noted, the previous Selenium IT only includes the component clientlib, however the test did not catch the Granite dependency as it does not exercise the viewer part, only the dialog editing.

> 2. The IT default-policy switch goes the opposite direction

I explained above, this is a false positive from AI review: the new `cmp-test.base.forms` clientlib is equivalent to the previous `cmp-test.base` one, only the PDF Viewer pages were change to use the new/slimmer `cmp-test.base` clientlib.

examples/ui.apps/src/content/jcr_root/apps/core-components-examples/clientlibs/clientlib-base/.content.xml
myronenko: removed in https://github.com/adobe/aem-core-wcm-components/pull/3032/changes/8a34d2f4aa041992674207714f899377a477f27d

myronenko (approved):

LSantha:
@vladbailescu , I see, makes sense.  Thank you for the update.

PR #3067, opened 2026-07-30
GRANITE-71247: ContentAI Search / Quick Search v3 constant decoupling, null-safety, and clear-button UX fixes

Apoorv-R98:
| Fixed Issues?            | Related to GRANITE-70028
| Minor: New Feature?      |
| Major: Breaking Change?  |
| Tests Added + Pass?      | Existing tests updated where applicable; no new tests for the CSS/UX-only changes
| Documentation Provided   | Code comments where non-obvious; component README updated
| Any Dependency Changes?  |

## Summary

Fixes a set of UX/race-condition bugs in the ContentAI Supported Search (`contentaisearch`) component and a couple of related issues in Quick Search v3 (`search`), found while building a real demo site against these components.

### ContentAI Supported Search (`contentaisearch`)
- **Removed search-as-you-type debounce; search now only fires on explicit submit** (Enter / mobile "Go", or the AI-answer toggle changing) — previously a fresh, often-incomplete query fired on every ~300ms typing pause, causing visible flicker and repeated network calls.
- **Fixed a stale-response race** across results search, gensearch, and load-more — a slower, older response arriving after a newer one could overwrite it with stale data.
- **Clear button now invalidates in-flight requests** instead of letting a late response silently repopulate results after the user cleared the field.
- **Fixed result/source links** — falls back to `metadata.source`/`data.source` when `metadata.url` is absent, matching Content AI's acquisition-indexing convention (previously such items linked to nothing).
- **Render the AI-generated summary's Markdown** (bold, links, bullet lists) instead of showing raw `**`/`[text](url)` syntax as literal text, with a word-by-word reveal.
- **Results now respect the configured `resultsSize`** as a true cap on the final merged results across all configured content sources (previously could return up to N × resultsSize for N sources) — also fixed `totalResults` to sum across sources instead of taking the max.
- **Deduped in-flight requests, fixed a Load More error state, and reuse a single HTTP client** instead of creating one per request.
- Made `modelFactory` `protected` for correct OSGi field injection; documented the Content AI API key OSGi configuration.

### Quick Search v3 (`search`)
- **Renamed the "AI Search" toggle label to "Semantic Search"** for clearer, more accurate wording (toggle label + design dialog "hide toggle" text/description). Internal identifiers (`hideAiSearchToggle` property, `aiToggle` JS hook, `cmp-search__ai-toggle` CSS classes) are left unchanged to avoid a breaking rename of authored policy config.
- **Decoupled `ContentAISupportedSearch`'s default content source type from `ContentAIClient`** — moved `DEFAULT_CONTENT_SOURCE_TYPE` onto the `ContentAISupportedSearch` model interface itself, so the component's model layer no longer depends on a service-layer class it otherwise has no relationship to. `ContentAIClient` keeps its own copy for its own independent default `search()`/`genSearch()` overloads.
- **Null-guarded `search.js`'s status-message updates** — `_hideSearchResultsStatusMessage`/`updateSearchResultsStatusMessageElement` previously threw a null-pointer error whenever this compiled JS initialized against an older v1 markup instance of the same component on the same page (a real scenario for any project running two component versions side by side), since v1 never renders the `.cmp_search__info` element v3 introduced.
- **Reworked the clear (×) button styling** for a real hit area — moved position/sizing from the bare icon onto the button itself (`top: 50%` + `translateY(-50%)`, robust to field-height changes vs. the old fixed offset), added a circular padded hit area with hover/`:focus-visible` feedback (previously an unstyled icon with no affordance).


Apoorv-R98:
Splitting this into two smaller, focused PRs for easier review:

- #3068 - ContentAI Supported Search (`contentaisearch`) component fixes
- #3069 - Quick Search v3 (`search`) component fixes

Closing this one in favor of those two.

PR #3081, opened 2026-08-31
GRANITE-72581: Enable Content AI Search v1 for AMS / classic AEM 6.5 LTS

Apoorv-R98:
## Summary
Resolves [GRANITE-72581](https://jira.corp.adobe.com/browse/GRANITE-72581) - enables Content AI Search (v1) support for Adobe Managed Services (AMS), while keeping the visitor-facing generative-search toggle restricted to properly supported environments.

## Behavior
| Environment | Gen-search toggle |
|---|---|
| AEM as a Cloud Service | follows author's dialog setting |
| Classic AEM 6.5, branded `LTS` version qualifier (e.g. `6.5.2.LTS`) - incl. AMS | follows author's dialog setting |
| Classic AEM 6.5, bare/unbranded version (e.g. `6.5.0`) | always hidden (unsupported) |

Search results and generative-search backend calls (`ContentAIClientImpl`) are unaffected by this gate - they already work on 6.5/AMS via the `baseUrlOverride` OSGi config, since those environments don't expose the `AEM_PROGRAM_ID`/`AEM_ENV_ID`/`AEM_SERVICE` env vars AEM as a Cloud Service uses to derive its own Content AI host.

## Changes
- Renamed `AemCloudPlatformDetector` -> `AemVersionDetector` (now covers more than cloud detection) and added `isUnsupportedClassic65ServicePack()`, which checks for the classic AEM 6.5 `LTS` version qualifier.
- Wired the new check into `ContentAISupportedSearchImpl#resolveGenSearchToggleVisible()`.
- Updated `ContentAIConfig`/`ContentAIClientImpl` docs to describe `baseUrlOverride` as the supported way to configure Content AI on 6.5 LTS/AMS, not just local dev.
- Updated the `ContentAISupportedSearch` model javadoc and component `README.md`.
- Updated/added unit tests in `AemVersionDetectorTest` and `ContentAISupportedSearchImplTest` covering Cloud Service, branded classic 6.5 LTS, bare classic 6.5, and future service packs.

## Testing
Full `bundles/core` unit test suite: 1187 tests, 0 failures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vladbailescu (commented):
## Review summary — GRANITE-72581

Clean, well-tested, and well-documented change. Verified the version-gating logic (`6.5.0`→hidden, `6.5.2.LTS`/`6.5.21.lts`→visible, `6.6.0`/`2026.x`→visible, `6.5.5.NOTLTSBUILD`→hidden via exact-qualifier match), the `genSearch` `contentSource.type` removal (endpoint 400s on it — real fix, now asserted), and the doc/README updates. Test coverage is strong and nothing is lost from the deleted `AemCloudPlatformDetectorTest` (all cases carried into `AemVersionDetectorTest`). CI is green. The two GitHub Advanced Security comments are already resolved — the final code uses `Strings.CI.equals` (commons-lang3), not the deprecated `StringUtils.*`.

**0 critical / 0 blocking.** Two non-blocking points, posted inline:

1. **Fail-open default when the platform is indeterminable.** The gate moved from an allow-list (`!isCloudPlatform` → hidden on everything non-cloud) to a deny-list (only non-LTS 6.5 hidden). When the version can't be resolved (null `ProductInfoProvider`/`ProductInfo`/version), or on other non-cloud classic (e.g. 6.4), the toggle now follows the author setting (default visible) instead of being hidden. Inert on real Cloud/LTS, but worth confirming the widening is intentional rather than defaulting to hidden when the version is unknown.

2. **v1/v3 gating asymmetry.** Quick Search v3 (`SearchImpl`) still uses `isCloudPlatform` (hidden even on branded 6.5 LTS), while Content AI Search v1 now allows 6.5 LTS. Flagging so the divergence is a deliberate scoping choice.

Nit (no change needed): `contentSourceType` is now unused in `genSearch` — the doc comment already explains the signature-symmetry rationale.

Nice work overall — happy to see it merge once the above are acknowledged.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ContentAISupportedSearchImpl.java
```diff
@@ -119,7 +119,7 @@ private void initModel() {
     }
 
     private boolean resolveGenSearchToggleVisible() {
-        if (!AemCloudPlatformDetector.isCloudPlatform(productInfoProvider)) {
```
vladbailescu, line 122: Non-blocking — behavior change worth a conscious confirm: this flips the toggle gate from an allow-list (previously `!isCloudPlatform(...)` → hidden on all non-cloud) to a deny-list (only non-LTS 6.5 hidden). As a result, when the version can't be determined (null `ProductInfoProvider`/`ProductInfo`/version — `isNonLtsClassic65` returns `false` there, per `isNonLtsClassic65_falseWhenProviderMissing`) or on any other non-cloud classic (e.g. 6.4), the toggle now follows the author setting (default visible) rather than being hidden. On real Cloud/LTS this is inert since ProductInfo is present, but consider defaulting to hidden when the platform is indeterminable to preserve the prior safe default.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v3/SearchImpl.java
```diff
@@ -61,7 +61,7 @@ public class SearchImpl extends com.adobe.cq.wcm.core.components.internal.models
 
     @PostConstruct
     private void initV3Model() {
-        boolean hideByDefault = !AemCloudPlatformDetector.isCloudPlatform(productInfoProvider);
+        boolean hideByDefault = !AemVersionDetector.isCloudPlatform(productInfoProvider);
```
vladbailescu, line 64: Question (non-blocking): Quick Search v3 keeps `isCloudPlatform` gating (hidden even on branded 6.5 LTS), while Content AI Search v1 now allows 6.5 LTS via `isNonLtsClassic65`. If the divergence is intentional (this PR is scoped to Content AI Search v1), no change needed — just confirming it's deliberate rather than an overlooked spot.

Apoorv-R98: this is intentional as Quick Search v3 only supports CLoud Platform and Content AI Search v1 now supports - Cloud as well as AEM 6.5 LTS.

bundles/core/src/main/java/com/adobe/cq/wcm/core/components/internal/models/v1/ContentAISupportedSearchImpl.java
Apoorv-R98: Fixed in b9d850f, replaced the deny-list isNonLtsClassic65 with an allow-list i`sGenSearchSupportedPlatform()` (true only for Cloud Service or branded 6.5 LTS, false otherwise — including indeterminate platform). `resolveGenSearchToggleVisible()` now fails closed again on an unresolvable platform, matching pre-PR behavior. 
Added tests for the indeterminate cases, including one confirming the toggle stays hidden even when the author property would otherwise enable it.

YahorC (approved):

PR #3082, opened 2026-09-01
feat: expose LinkHandler interface (#1818)

### LINKED ISSUE #1818: [Link Hander] expose interface of Link Handler to be able to use it in other Models

DavidSMonedero:
| Q                     | A 
| --------------------- | ---
| Fixed Issues?         | Fixes #1818
| Patch: Bug Fix?       |
| Minor: New Feature?   | Yes
| Major: Breaking Change? |
| Tests Added + Pass?   | Yes
| Documentation Provided | Yes (code comments and or markdown)
| Any Dependency Changes? |
| License               | Apache License, Version 2.0

## Description
Exposes the `LinkHandler` interface to the public API package (`com.adobe.cq.wcm.core.components.commons.link`).

This change allows custom Sling Models to inject and leverage the internal `LinkHandler` service directly for resolving component links in a standardized way.


DavidSMonedero:
@adobe-cla-bot check

